PR #225, opened 2023-10-13
feat(transaction-selection): add block size transaction selector

Gabriel-Trintinalia:
Add block size limit validation.

This is an initial implementation of the validation and is still using AVERAGE_TX_RLP_SIZE. The next step is to replace it with the actual transaction size.


arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
```diff
@@ -0,0 +1,113 @@
...
+  public TransactionSelectionResult evaluateTransactionPostProcessing(
+      final PendingTransaction pendingTransaction,
+      final TransactionProcessingResult processingResult) {
+
+    int L1TransactionSize =
+        calculateL1TransactionSize(pendingTransaction, processingResult.getLogs());
+    if (isTransactionExceedingL1BlockSizeLimit(L1TransactionSize)) {
+      log.atTrace()
+          .setMessage(
+              "Cumulative block size in bytes including tx {} is {} greater than the max allowed {}, skipping tx")
```
macfarla, line 60: suggestion -
              "tx {hash} size {} would cause block (cumulative size {}) to exceed max {} bytes, skipping tx"
eg
tx 0x1234 (size 52) would cause block (cumulative size 499) to exceed max 500 bytes, skipping tx

acceptance-tests/src/test/java/linea/plugin/acc/test/BlockBytesLimitTest.java
```diff
@@ -0,0 +1,103 @@
...
+   * count is set to ten, which exceeds the maximum block bytes size set in the getTestCliOptions
+   * method. The maximum block bytes size is calculated to accommodate one transaction and two logs.
+   * Therefore, a transaction with ten logs exceeds this limit and cannot be mined into a block.
+   */
+  @Test
+  public void shouldNotMineTransactionWhenBlockBytesExceeded() throws Exception {
+    verifyTransactionReceipt(BigInteger.TEN, eth::expectNoTransactionReceipt);
+  }
+
+  private void verifyTransactionReceipt(BigInteger logCount, Function<String, Condition> condition)
```
macfarla, line 74:
```suggestion
  private void verifyTransactionReceipt(BigInteger txSize, Function<String, Condition> condition)
```

macfarla: txSizeInBytes ?

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
```diff
@@ -0,0 +1,113 @@
...
+      PendingTransaction pendingTransaction) {
+    return TransactionSelectionResult.SELECTED;
+  }
+
+  @Override
+  public TransactionSelectionResult evaluateTransactionPostProcessing(
+      final PendingTransaction pendingTransaction,
+      final TransactionProcessingResult processingResult) {
+
+    int L1TransactionSize =
```
macfarla, line 55: I don't love the uppercase L but a lowercase l doesn't look good either

macfarla: this is better transactionL1Size

Gabriel-Trintinalia: done

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
Gabriel-Trintinalia: done

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
```diff
@@ -0,0 +1,114 @@
...
+import org.hyperledger.besu.evm.log.Log;
+import org.hyperledger.besu.plugin.data.TransactionProcessingResult;
+import org.hyperledger.besu.plugin.data.TransactionSelectionResult;
+import org.hyperledger.besu.plugin.services.txselection.PluginTransactionSelector;
+
+@Getter
+@Slf4j
+public class MaxBlockSizeTransactionSelector implements PluginTransactionSelector {
+  private int cumulativeBlockBytesSize = EMPTY_L1_BLOCK_SIZE;
+  private final int maxBytesPerBlock;
```
pinges, line 34: We usually have the constants first and then the other fields

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Checks if the total size of all transactions in a block would exceed the maximum allowed size
+   * if the given transaction were added.
+   *
+   * @param L1TransactionSize the L1 Transaction Size in Bytes.
+   * @return true if the total call data size would be too big, false otherwise.
+   */
+  private boolean isTransactionExceedingL1BlockSizeLimit(int L1TransactionSize) {
```
pinges, line 78: final is missing and the variable should start with a lower case letter

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
```diff
@@ -0,0 +1,114 @@
...
+    try {
+      return Math.addExact(cumulativeBlockBytesSize, L1TransactionSize) > maxBytesPerBlock;
+    } catch (final ArithmeticException e) {
+      // Overflow won't occur as totalBlockSize won't exceed Integer.MAX_VALUE
+      return true;
```
pinges, line 83: should we not return false, just to be safe?

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Updates the total size when a transaction is selected.
+   *
+   * @param pendingTransaction The selected transaction.
+   */
+  @Override
+  public void onTransactionSelected(
+      final PendingTransaction pendingTransaction, final TransactionProcessingResult result) {
+    final int transactionL1Size = calculateL1TransactionSize(pendingTransaction, result.getLogs());
```
pinges, line 95: we should not have to calculate the size here again, we have done that in the post processing already

arithmetization/src/test/java/net/consensys/linea/sequencer/txselection/MaxBlockBytesSizeTransactionSelectorTest.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  private void verifyTransactionSelection(
+      final PluginTransactionSelector selector,
+      final PendingTransaction transaction,
+      final TransactionProcessingResult processingResult,
+      final TransactionSelectionResult expectedSelectionResult) {
+    var selectionResult = selector.evaluateTransactionPostProcessing(transaction, processingResult);
+    assertThat(selectionResult).isEqualTo(expectedSelectionResult);
+    notifySelector(txSelector, transaction, processingResult, selectionResult);
```
pinges, line 84: The notify here does not really do anything because you are only calling the verifyTransactionSelection once in each of the tests.
You could add one, where you call the verifyTransactionSelection twice, where the second transaction exceeds the limit (but the second tx on it's own would have been OK).

pinges (changes requested):

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
Gabriel-Trintinalia: No, the method returns true if the transaction will overflow the maxBytesPerBlock. In that case, we don't want the transaction to be selected

arithmetization/src/test/java/net/consensys/linea/sequencer/txselection/MaxBlockBytesSizeTransactionSelectorTest.java
Gabriel-Trintinalia: I need to notify the selector otherwise I cannot assert the cumulativeBlockSize value in the selector.

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
Gabriel-Trintinalia: do you suggest saving that size so we can reuse it?

Gabriel-Trintinalia: @pinges

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockSizeTransactionSelector.java
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2: Copyright Consensys Software Inc

delehef:
I am not sure what is the goal of this PR, given that we already seem to ensure L1 size limits [through the standard limit mechanism]https://github.com/Consensys/besu-sequencer-plugins/blob/main/arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java)?

Gabriel-Trintinalia:
Closing as implemented by https://github.com/Consensys/besu-sequencer-plugins/pull/445

PR #226, opened 2023-10-13
Add a PluginTransactionSelector to check the limit for lines of trace

pinges:


acceptance-tests/src/test/java/linea/plugin/acc/test/PluginTest.java
```diff
@@ -266,6 +294,50 @@ public void deniedTransactionNotAddedToTxPool() throws Exception {
         .isEqualTo("destination address is a precompile address and cannot receive transactions");
   }
 
+  @Test
+  public void transactionIsNotMinedWhenTooManyTraceLines() throws Exception {
```
macfarla, line 298: this test name doesn't seem to match the code

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2:
```suggestion
 * Copyright Consensys Software Inc.
```

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
```diff
@@ -0,0 +1,100 @@
...
+    for (var e : lineCounts.entrySet()) {
+      final String module = e.getKey();
+      if (!moduleLimits.containsKey(module)) {
+        final String errorMsg =
+            "Module " + module + " does not exist in the limits file: " + limitFilePath;
+        log.error(errorMsg);
+        throw new RuntimeException(errorMsg);
+      }
+      if (lineCounts.get(module) > moduleLimits.get(module)) {
+        return TransactionSelectionResult.BLOCK_FULL;
```
macfarla, line 90: do we need to give any more details? like which module limit was exceeded?

macfarla: does it matter which order the module limits are checked?

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -126,7 +138,16 @@ public void traceContextExit(MessageFrame frame) {
   }
 
   /** When called, erase all tracing related to the last included transaction. */
-  public void popTransaction() {
-    hub.popTransaction();
+  public void popTransaction(final PendingTransaction pendingTransaction) {
+    if (hashOfLastTransactionTraced.equals(pendingTransaction.getTransaction().getHash())) {
```
macfarla, line 142: why do we need to check this? when would it not match?

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
pinges: As far as I understand any of the modules over the limit means that the block is full. We could log something if that was helpful ...

pinges: Order should not matter, AFAIU

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
pinges: We have to only call the hub.pop() here if the startTransaction() for this transaction has been called. The popTransaction() is called from the plugin when the onTransactionNotSelected() in the BlockTransactionSelector is called and that can happen before the startTransaction() in the MainnetTransactionProcessor is called, e.g. because of one of the evaluateTransactionPreProcessing() calls failed.

acceptance-tests/src/test/java/linea/plugin/acc/test/PluginTest.java
pinges: The test sends 5 transactions, each of these transactions keeps the trace lines for the add module under the limit, but a second of these transactions pushes the block over the limit.
The test makes sure that all transactions are mined, but only one per block.

macfarla: I think that's a helpful comment to add...

macfarla:
```suggestion
  public void secondTransactionIsNotIncludedInBlock_WhenTooManyTraceLines() throws Exception {
```

macfarla: or maybe `onlyOneTransactionFitsInABlock`

macfarla (approved):
LGTM

PR #243, opened 2023-10-20
feat: ensure that the predicted line count is correct

delehef:
Failing in predicting the correct number of lines may lead to catastrophic failure in the proving process.

Therefore, we ensure that the predicted number of lines actually matches the number of generated lines.


OlivierBBB:
For documentation purposes exerpts from the relevant slack discussion:

image

image

PR #317, opened 2023-11-06
Validate tx calldata size in the tx validator plugin

fab-10:
Move the validation of the calldata size of a transaction, from the Linea Besu fork, into the TransactionValidator plugin, plus some minor optimization and code reorgs


macfarla:
There is an actual test failure I think
```
PluginTest > deniedTransactionNotAddedToTxPool() FAILED
    org.opentest4j.AssertionFailedError at PluginTest.java:238
```

PR #335, opened 2023-11-08
RPC endpoint to get traces counters from ZkTracer

### LINKED ISSUE #179: Implement rollup_getTracesCounters
This method is intended to be used for recovery purposes to replay the traces counting for a block previously created and already in the canonical chain, e.g coordinator crash, bug in zkBESU traces counting.

rollup_getTracesCountersByBlockNumberV1(blockNumber, expectedTracesEngineVerision)

Parameters:
blockNumber: String - decimal encoded integer of the start block number
expectedTracesEngineVersion: String - version of expected traces version system. If the runtime version is not the expected version, an error shall be returned. In case of traces being cached, if cached version is not the same as expected one, new traces shall be generated. 

Response: Object
result: Object - with
- tracesCounters: Object with moduleName: number of lines per module
- tracesEngineVersion: String - version of traces module

error: Object with error info
- type: an enum of possible errors.
- message: Detail error message

macfarla:
get traces counters

built on top of #345

fixes #179


arithmetization/src/main/java/net/consensys/linea/tracegeneration/rpc/TracesCounters.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright ConsenSys AG.
```
Gabriel-Trintinalia, line 2:
```suggestion
* Copyright Consensys Software Inc.
```

macfarla: fixed

macfarla:
still needs ATs

macfarla:
// see com.google.common.cache.Cache for possible cache implementation

arithmetization/src/main/java/net/consensys/linea/tracegeneration/rpc/RollupGetTracesCountersByBlockNumberV0.java
```diff
@@ -0,0 +1,79 @@
...
+      return getTracesCounters(params);
+    } catch (Exception ex) {
+      throw new PluginRpcEndpointException(ex.getMessage());
+    }
+  }
+
+  @VisibleForTesting
+  TracesCounters getTracesCounters(final TracesCountersRequestParams params) {
+    if (traceService == null) {
+      traceService = initTraceService();
```
fab-10, line 64: can this done once, at object creation time?

arithmetization/src/main/java/net/consensys/linea/tracegeneration/LineCountsByBlockCache.java
```diff
@@ -0,0 +1,48 @@
...
+      final TraceService traceService, final long blockNumber) {
+    if (!lineCountsCache.asMap().containsKey(blockNumber)) {
+      // trace the block, which will add the counters to the cache
+      traceService.traceBlock(blockNumber, new ZkTracer());
+    }
+    if (lineCountsCache.asMap().containsKey(blockNumber)) {
+      return lineCountsCache.getIfPresent(blockNumber);
+
+    } else {
+      //      throw new RuntimeException("traceBlock() failed. Line counts do not exist in cache.");
```
fab-10, line 44: remove?

arithmetization/src/main/java/net/consensys/linea/tracegeneration/rpc/RollupGetTracesCountersByBlockNumberV0.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+    return new TracesCounters(
+        params.runtimeVersion(),
+        LineCountsByBlockCache.getBlockTraces(traceService, params.fromBlock()));
```
fab-10, line 68: In case the block is not present we return a null map, isn't better to return an error response?

non-fungible-nelson:
can we have a status on this?

ahamlat:
@macfarla Even with my last suggestion, I couldn't make this PR work on a Forest Archive node on Linea Goerli on an old block. I don't think the cache should be the one that has to return the result all the time. We should get the result from the cache if the block is already traced, or trace the block, get the line counting, and **then** update the cache.
Ideally, we should have a light mode where the tracing is done only to generate the number of lines per module, not the traces file.

PR #346, opened 2023-11-09
Refactor : Clean acc tests

Gabriel-Trintinalia:
- This PR splits the PluginTest class in more meaningful classes.  
- Introduces a common default CLI object to be used
- Splits some methods in smaller pieces and reuse them when possible


macfarla (approved):
LGTM just spdx headers

Issue #245, opened 2023-10-21
Unify the MXPX detection between HUB / MXP modules


OlivierBBB:
# Main issue

The `hub` and `mxp` modules both implement a variant of `ixMxpx()` which detects $\texttt{memoryExpansionException}$ 's. We should only have one at some point.

PR #359, opened 2023-11-10
feat: implement PHC

delehef:
- Addresses #245 
- Fixes reference tests

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Exceptions.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Exceptions.java
@@ -305,10 +305,13 @@ public final class Exceptions {
           MLOAD,
           MSTORE,
           MSTORE8 -> {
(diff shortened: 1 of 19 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Exceptions.java
```diff
@@ -300,49 +300,49 @@ public void prepare(final MessageFrame frame, GasProjector gp) {
...
       default -> {
-        if (this.outOfGas = isOutOfGas(frame, opCode, gp)) {
+        if (this.outOfGas == isOutOfGas(frame, opCode, gp)) {
           return;
         }
       }
     }
 
-    if (this.invalidCodePrefix = isInvalidCodePrefix(frame)) {
+    if (this.invalidCodePrefix == isInvalidCodePrefix(frame)) {
```
powerslider, line 345: This return is unnecessary. Either move this condition before the switch statement or change to logic in some way.

powerslider (approved):

Issue #358, opened 2023-11-10
Skip tracing of codeless frames & Geth vs Besu execution engines


delehef:
# Main point

Although Besu will execute a single ‶virtual `STOP`″ in frames without bytecode, this opcode *shall not* be traced by zkBesu.

## Details

The issue arises for 2 different reasons: line counting in Java vs. Go and associated tracing + Besu vs LineaTo wit:
- zkGeth vs. zkBesu line counting / instruction:
  - [ ] #361 
    - instruction_decoder.go
    - opCode.yaml
- Geth vs. Besu instruction sequence:
  - [x] #362 
    - the Geth and Besu execution engines produce inconsistent instruction sequences
- Besu vs. Linea zkEvm spec:
  - the Linea spec enforces certain operations to be done efficiently (e.g. the `STOP`'s from #362 aren't traced)

PR #360, opened 2023-11-10
fix: do not trace code-less frames

delehef:
Addresses #358


arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
OlivierBBB, line 1: LGTM

OlivierBBB (approved):
Looks good

PR #428, opened 2023-11-23
More acceptance tests and use of Clique consensus

fab-10:


Gabriel-Trintinalia (approved):
LGTM - Nice work with the tests readability

PR #429, opened 2023-11-23
Fix cumulative user gas used in block calculation

fab-10:


ahamlat (approved):

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/MaxBlockCallDataTransactionSelector.java
```diff
@@ -53,11 +53,13 @@ public TransactionSelectionResult evaluateTransactionPreProcessing(
...
     if (isTransactionExceedingBlockCallDataSizeLimit(transactionCallDataSize)) {
-      log.trace(
-          "BlockCallData {} greater than {}, completing operation",
-          transactionCallDataSize,
-          maxBlockCallDataSize);
-      return BLOCK_CALLDATA_FULL;
+      log.atTrace()
+          .setMessage(
+              "BlockCallData including this tx is {} greater than the max allowed {}, skipping tx")
+          .addArgument(() -> blockCallDataSize + transactionCallDataSize)
```
ahamlat, line 59: Is here blockCallDataSize or  blockCallDataSize + transactionCallDataSize ?

fab-10: the latter, since `blockCallDataSize` still does not contain the current tx calldata size

ahamlat: I believe it is already added with onTransactionSelected, right ?

fab-10: no, the sum is done at last step when the tx is selected in the `onTransactionSelected`

PR #434, opened 2023-11-27
Fix/issue 433/shadow node

### LINKED ISSUE #433: Fix shadow node issues
Errors found in blocks:
 - [X] 245982: erroneous `DataCopy` gas computation
 - [x] 247011: erroneous `Exp` gas computation
 - [X] 248205: erroneous `DataCopy` gas computation
 - [X] 254058: erroneous `DataCopy` gas computation

delehef:
Fixes #433

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Exceptions.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Exceptions.java
@@ -82,6 +82,44 @@ public final class Exceptions {
     this.codeSizeOverflow = codeSizeOverflow;
   }
 
+  @Override
+  public String toString() {
+    if (this.invalidOpcode) {
+      return "Invalid opcode";
+    }
+    if (this.stackUnderflow) {
+      return "Stack underflow";
+    }
+    if (this.stackOverflow) {
+      return "Stack overflow";
+    }
+    if (this.outOfMemoryExpansion) {
+      return "Out of MXP";
+    }
+    if (this.outOfGas) {
+      return "Out of gas";
+    }
+    if (this.returnDataCopyFault) {
+      return "RDC fault";
+    }
+    if (this.jumpFault) {
+      return "JMP fault";
+    }
+    if (this.staticFault) {
+      return "Static fault";
+    }
+    if (this.outOfSStore) {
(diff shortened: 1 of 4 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/DataCopy.java
```diff
@@ -18,23 +18,19 @@
...
 import org.hyperledger.besu.evm.internal.Words;
 
 public final class DataCopy implements GasProjection {
   private final MessageFrame frame;
   private long offset = 0;
-  private long size = 0;
-  private int bitSize = 0;
+  private long len = 0;
 
   public DataCopy(MessageFrame frame) {
```
OlivierBBB, line 29: Where is this method used ? Asking because the offset/size arguments won't be at the same position for `CODECOPY` as they will be for `EXTCODECOPY`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Exceptions.java
```diff
@@ -82,6 +82,44 @@ public Exceptions(
...
+    if (this.stackUnderflow) {
+      return "Stack underflow";
+    }
+    if (this.stackOverflow) {
+      return "Stack overflow";
+    }
+    if (this.outOfMemoryExpansion) {
+      return "Out of MXP";
+    }
+    if (this.outOfGas) {
```
OlivierBBB, line 99: It would be nice to harmonize notations between the spec and the implementation. E.g. the spec calls these things
- `memoryExpansionException`
- `returnDataCopyException`
- `invalidJumpException`
- `staticContextException` 
- `sstoreException`
- `maxCodeSizeException`

Although confusingly we have both a

OlivierBBB (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Exceptions.java
delehef: I think it would be nice to have a more general name harmonization PR in the whole plugin when the dust will have settled down.

PR #443, opened 2023-11-28
perf: more optimizations

delehef:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
letypequividelespoubelles, line 1: Currently, `RomLex` must be called BEFORE `Rom`, `RlpTxn`, `TxnData`, `RAM` (all modules recquiring `CODE_FRAGMENT_INDEX`)

letypequividelespoubelles (changes requested):

letypequividelespoubelles (approved):

PR #445, opened 2023-11-29
feat: implement L1 block size & Keccak limits

### LINKED ISSUE #444: Implement the computation of KECCAK count counter

delehef:
Fix #444


arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1Block.java
```diff
@@ -0,0 +1,169 @@
...
+
+  @Override
+  public void traceEndTx(
+      WorldView worldView,
+      Transaction tx,
+      boolean status,
+      Bytes output,
+      List<Log> logs,
+      long gasUsed) {
+    this.rlpSizes.push(this.rlpSizes.pop() + tx.encoded().size());
```
letypequividelespoubelles, line 167: Don't you want the RLP(log) here ?

letypequividelespoubelles (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1Block.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @Override
+  public void popTransaction() {
+    this.rlpSizes.pop();
+    this.l2l1Logs.pop();
+  }
+
+  @Override
+  public void tracePreOpcode(MessageFrame frame) {
```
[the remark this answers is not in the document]
delehef: Indeed, thx

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1Block.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.datatypes.Transaction;
+import org.hyperledger.besu.evm.log.Log;
+import org.hyperledger.besu.evm.log.LogTopic;
+import org.hyperledger.besu.evm.worldstate.WorldView;
+
+@RequiredArgsConstructor
+@Accessors(fluent = true)
+public class L1Block implements Module {
```
Filter94, line 40: Do you have a spec to compare the implementation to? I remember we had series of tickets, that I could find, but I wonder if there's a newer document where all of that compiled?

delehef: [This PR](https://github.com/Consensys/zk-geth/pull/35) from Alex, which he didn't merge in the end.

AlexandreBelling: Also, this would deserve some equivalence testing. To make sure both implementations returns the same values. Seems easy enough and worthwhile

delehef: Testing on “same values” is probably not a good idea as the Geth PR has neither been tested nor deployed.

If the prover team has some pairs of correct input/output, we would do our best to test them.

letypequividelespoubelles (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1Block.java
FlorianHuc: can we add few unit test for the line counts using date you build.
Then we can add some tests using existing blocs from testnet based on our validation of this coming we.

AlexandreBelling (approved):

PR #453, opened 2023-12-01
feat: consider `TRACES_DIR` in the traces API

delehef:


arithmetization/src/main/java/net/consensys/linea/continoustracing/SlackNotificationService.java
```diff
@@ -65,12 +65,11 @@ public void sendCorsetFailureNotification(
...
-                                    "Block hash: *"
-                                        + blockHash
-                                        + "*\n\n"
-                                        + "Trace verification failed with the following error:\n\n"
+                                    "Trace verification failed with the following error:\n\n"
                                         + "```"
-                                        + validationResult.corsetOutput()
+                                        + validationResult
+                                            .corsetOutput()
+                                            .replaceAll("\u001B\\[[;\\d]*m", "")
```
daniellehrner, line 72: What does this do exactly? It might make sense to add a comment explaining this one line, because at least for me it is not clear without spending some time looking into it.

delehef: It removes all the color/bold ANSI escape codes from Corset, because Slack refuses them

lorenzogentile404 (commented):
Ok

PR #455, opened 2023-12-04
Use a simple "for" loop instead of a stream to improve performance

ahamlat:
We noticed from CPU profiling during block production, i.e transaction evaluation in order to include them in a block that State.lineCount consumes a significant amount of CPU, related to the usage of a stream.

image

JMH Microbenchmarking confirmed that a simple loop is much better in this use case on a collection with 100k elements

```
Benchmark                            Mode  Cnt      Score     Error  Units
TestStreamsVsLoops.testMethodLoop    avgt   50   9976.233 ± 138.659  ns/op
TestStreamsVsLoops.testMethodStream  avgt   50  64048.445 ± 386.111  ns/op
```


delehef (approved):

ahamlat:
#457

PR #459, opened 2023-12-05
temp: geth counting compatibility

delehef:
This is a temporary PR, **never to be merged**, that should only live as long as we need to mix a Geth tracer with a Besu sequencer.

Tests are expected to fail.


letypequividelespoubelles:
We finally don't use geth anymore :)

PR #476, opened 2023-12-07
changed default limits file name to toml

macfarla:
fixes default file name extension. File is processed as TOML since #448


Gabriel-Trintinalia (approved):
lgtm

PR #506, opened 2023-12-19
style: update name of prec limits to avoid confusion with old geth name

letypequividelespoubelles:


delehef (approved):

letypequividelespoubelles:
so @FlorianHuc the idea is :
- we use new names for the new limits on Besu/main (tipically number of relevant (ie without exceptions, etc) call, weighted by the size of the call (nb of blocks for hash precompile)
- on the temp branch, I'll create/rename module which counts the number of call to the precompile (even if not succesfull because of exceptions) and give them the same name as in Geth. 
Is it ok for you ?

FlorianHuc (approved):

PR #512, opened 2023-12-21
feat: introduce a trivial assembler in `BytecodeCompiler`

delehef:
Introduce a way to generate test bytecode in mnemonic form. For instance:

```
BytecodeCompiler.newProgram()
                .assemble("""
      ; Perform some addition
      PUSH8  02
      PUSH32 0x1234
      ADD
      """)
               .compile()
```

### DIFF
--- a/arithmetization/src/test/java/net/consensys/linea/zktracer/testing/BytecodeCompiler.java
+++ b/arithmetization/src/test/java/net/consensys/linea/zktracer/testing/BytecodeCompiler.java
@@ -17,6 +17,9 @@ package net.consensys.linea.zktracer.testing;
 
 import static net.consensys.linea.zktracer.types.Conversions.bigIntegerToBytes;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


arithmetization/src/test/java/net/consensys/linea/zktracer/testing/BytecodeCompiler.java
```diff
@@ -46,6 +49,41 @@ private static Bytes toBytes(final int x) {
...
+        if (line.isBlank()) continue;
+        if (line.startsWith(";")) continue;
+
+        String[] ls = line.split("\\s+");
+        final OpCode opCode = OpCode.fromMnemonic(ls[0]);
+        this.op(opCode);
+        final int pushSize = opCode.byteValue() - (int) OpCode.PUSH1.byteValue() + 1;
+        final boolean isPush = pushSize >= 1 && pushSize <= 32;
+        if (isPush) {
+          this.immediate(Bytes.fromHexString(ls[1], pushSize));
```
OlivierBBB, line 74: Does this mean that we can write arbitrary size hexstrings after a PUSHX and it will only fail if there are fewer bytes than `pushSize` ?

delehef: The opposite; if there are fewer bytes that `pushSize`, then `fromHexString` will left-pad them with 0's; if there are too many, it will throw an exception.

OlivierBBB (approved):
LGTM

arithmetization/src/test/java/net/consensys/linea/zktracer/testing/BytecodeCompiler.java
```diff
@@ -46,6 +49,41 @@ private static Bytes toBytes(final int x) {
...
+      try {
+        if ((line = bufReader.readLine()) == null) break;
+        if (line.isBlank()) continue;
+        if (line.startsWith(";")) continue;
+
+        String[] ls = line.split("\\s+");
+        final OpCode opCode = OpCode.fromMnemonic(ls[0]);
+        this.op(opCode);
+        final int pushSize = opCode.byteValue() - (int) OpCode.PUSH1.byteValue() + 1;
+        final boolean isPush = pushSize >= 1 && pushSize <= 32;
```
powerslider, line 72: Is it possible to use `opCode.isPush()` method instead?

delehef: We could; but we need the `pushSize` in any case.

PR #541, opened 2024-01-14
refactor: add @EqualsAndHashCode annotations and remove corresponding methods

powerslider:
… methods

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/bin/BinOperation.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/bin/BinOperation.java
@@ -21,7 +21,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Stream;
 
-import com.google.common.base.Objects;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.experimental.Accessors;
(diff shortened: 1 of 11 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/bin/BinOperation.java
```diff
@@ -34,37 +34,24 @@
...
-
-  @Override
-  public boolean equals(Object o) {
-    if (this == o) return true;
-    if (o == null || getClass() != o.getClass()) return false;
-    final BinOperation that = (BinOperation) o;
-    return java.util.Objects.equals(opCode, that.opCode)
-        && java.util.Objects.equals(arg1, that.arg1)
-        && java.util.Objects.equals(arg2, that.arg2);
+  public BinOperation(OpCode opCode, BaseBytes arg1, BaseBytes arg2) {
```
delehef, line 51: Could that be a `@RequiredArgsConstructor`?

powerslider: Yes, I was just worried not to break something, if I do it, but actually there is no issue, so I will fix it.

delehef (approved):

PR #544, opened 2024-01-17
fix: semantics of LinkedList

delehef:


powerslider (approved):

powerslider (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/list/StackedList.java
```diff
@@ -51,19 +51,23 @@ public String toString() {
...
+    this.lists.addLast(new CountedList<>());
   }
 
   public void enter(int initialCapacity) {
-    this.lists.push(new CountedList<>(initialCapacity));
+    this.lists.addLast(new CountedList<>(initialCapacity));
+  }
+
+  public E getLast() {
+    return this.lists.getLast().getLast();
```
OlivierBBB, line 62: What is the explanation for the double `getLast` ?

OlivierBBB: BTW would `getLatest` be more descriptive and accurate than `getLast` ?

OlivierBBB (commented):
I don't fully get it but if Tsvetan agrees I'll take it

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/list/StackedList.java
delehef: `getLast` is the default Java method name for such accessors.

PR #561, opened 2024-01-22
feat: mechanics to capture conflations & replay them as test cases

delehef:
This PR adds two feature:

 - an RPC endpoint available under `rollup_captureConflation`, accepting a first and last block as `params`, and returning a JSON-encoded representation of the minimal set of information required to replay the given block range without requiring a fully-synchronized blockchain state, e.g.:
 ```
curl -X POST 'http://localhost:8545'
 --data '{
    "jsonrpc":"2.0",
    "method":"rollup_captureConflation",
    "params":["296519", "296521"], "id":"1"
  }'
| jq '.result.capture' -r 
| gzip > arithmetization/src/test/resources/replays/my-test-case.json.gz
```
 - an extension of the test framework to load a minimal state as defined above, and replay it as a unit test without requiring access to an archive of the blockchain state, through the use of the `ReplaysTests.replay(String fileName)` static method.

### CODE REVIEW

File: arithmetization/src/test/java/net/consensys/linea/zktracer/ReplaysTests.java [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+public class ReplaysTests {
+  public static void replay(String filename) {
+    ToyExecutionEnvironment environment = ToyExecutionEnvironment.builder().build();
+    environment.replay(filename);
+  }
+
+  @Test
+  void replayOneFile() {
+    replay("/home/franklin/asdf.json");
```
letypequividelespoubelles, line 29: put smtg more generic ;)

letypequividelespoubelles (approved):
LGTM

arithmetization/src/test/java/net/consensys/linea/zktracer/ReplaysTests.java
delehef: As soon as we have a fix for the tx tracing issue.

powerslider (approved):

PR #565, opened 2024-01-23
Check upfront profitability + Unprofitable txs cache and retry limit

### LINKED ISSUE #564: Reduce the overhead of unprofitable transactions during block creation
to improve/fix the transaction selection that is a bit overloaded by unprofitable transactions by:
* adding a cache to store the gas used values to avoid reprocessing the transaction every block
* additionally rejecting transactions before executing them if the profitability formula isn't matched using the gas LIMIT value

fab-10:
With this PR we want to mitigate the impact of unprofitable txs on the block creation process.
Fixes #564 

1. do a profitability check before processing the tx, using the `gasLimit`, this way we can filter txs that are upfront unprofitable without never processing them. If this check is not passed then the new `TX_UNPROFITABLE_UPFRONT` code is returned.
2. remember about unprofitable txs across block creations, this way we can put a limit to the number of these txs that are reprocessed every block. If an cached unprofitable tx is not retried due to max number of retries reached, then the new `TX_UNPROFITABLE_RETRY_LIMIT` code is returned.
3. only retry unprofitable txs if the current block minGasPrice is decreased, if not the new `TX_UNPROFITABLE_MIN_GAS_PRICE_INCREASED` code is returned

Two new options:

1. `plugin-linea-unprofitable-cache-size` max number of unprofitable txs we keep track of. Default 100_000.
2. `plugin-linea-unprofitable-retry-limit` how many cached unprofitable txs could be retried every block, Default 10.


julien-marchand:
LGTM so far

arithmetization/src/test/java/net/consensys/linea/sequencer/txselection/selectors/ProfitableTransactionSelectorTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+    // second try of the first tx
+    verifyTransactionSelection(
+        transactionSelector,
+        mockEvaluationContext1.setMinGasPrice(minGasPriceBlock2),
+        mockTransactionProcessingResult1,
+        TX_UNPROFITABLE_MIN_GAS_PRICE_NOT_DECREASED,
+        null);
+  }
+
```
jonesho, line 256: Can we have a test case, that would see in pre-processing result as `TX_UNPROFITABLE_UPFRONT`? This new result is quite critical to this next feature but I don't see any of those in the tests.

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/ProfitableTransactionSelector.java
```diff
@@ -43,57 +122,91 @@ public TransactionSelectionResult evaluateTransactionPostProcessing(
...
     }
-    return SELECTED;
   }
 
-  @Override
-  public TransactionSelectionResult evaluateTransactionPreProcessing(
-      final TransactionEvaluationContext<? extends PendingTransaction> evaluationContext) {
-    return SELECTED;
+  private void registerAsUnProfitable(final Transaction transaction) {
+    if (unprofitableCache.size() > unprofitableCacheSize) {
```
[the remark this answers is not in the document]
fab-10: Will change to >=.

arithmetization/src/test/java/net/consensys/linea/sequencer/txselection/selectors/ProfitableTransactionSelectorTest.java
fab-10: Yes will do

fab-10: test added

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/ProfitableTransactionSelector.java
```diff
@@ -30,11 +40,80 @@
...
+          transaction,
+          minGasPrice.getAsBigInteger().doubleValue(),
+          effectiveGasPrice,
+          gasLimit)) {
+        return TX_UNPROFITABLE_UPFRONT;
+      }
+
+      if (unprofitableCache.contains(transaction.getHash())) {
+        // only retry unprofitable txs if the min gas price went down
+        if (minGasPriceDecreased.isTrue()) {
```
Filter94, line 86: This is useful to tell when we definitely don't want to reprocess any unprofitable transactions (during gas price increase), but it's not very useful for cases when gas price fluctuates. I think we can improve this by storing gas price at which given tx was considered unprofitable and take this into account.

Note that `PriorityBlockingQueue` sorted by checked gas price won't work well here because there is no monotonous dependency between profitability and gas price. Heap sorted by profitability may work though

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/ProfitableTransactionSelector.java
```diff
@@ -30,11 +40,80 @@
...
+  @VisibleForTesting protected static Wei prevMinGasPrice = Wei.MAX_WEI;
+
   private final int verificationGasCost;
   private final int verificationCapacity;
   private final int gasPriceRatio;
   private final double minMargin;
   private final int adjustTxSize;
+  private final int unprofitableCacheSize;
+  private final int unprofitableRetryLimit;
+  private int unprofitableRetries;
```
Filter94, line 53: If we maintain a heap sorted by profitability, we might not need this. Because:
* If we're still selecting profitable transactions from this heap, we're doing useful job and should continue
* If we got an unprofitable transaction from the heap, we should just stop, without any retry limits because following transactions will be less profitable

Filter94: And one caveat with profitability heap cache. When gas price decreses, and we're selecting any profitable transaction from profitability cache, we must either select or re-evaluate all of the newly profitable transactions. Otherwise some transactions in this cache that weren't re-evaluated might get stuck with last known bad profitability

Filter94: Another caveat that state in theory may change transaction's profitability (gas consumed)

fab-10: Remember that here we are not in the txpool, and we cannot alter the order in which txs are evaluated, this PR is a quick way to mitigate the impact of unprofitable txs, until we have better solution with txpool that takes in consideration the profitability of a tx, and process txs in that order.
So here what we can do is try to understand if it is worth to retry the execution of a previously unprofitable tx, in case it is again a candidate for inclusion. 
Anyway your suggestion is good to implement a better filter, remembering the _unprofitable_ price for the tx, and re-processing it only if the current min price is lower, we should be able to do better guesses.
If you agree we can deploy this simple mitigation as is, and try to build a better filter in a next PR.

Filter94: Makes sense to me. I think this PR adds some value, so we can deploy it

Filter94 (approved):

macfarla (approved):

jonesho (approved):
LGTM

PR #566, opened 2024-01-23
feat: implement EC_DATA tracing

delehef:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/ec_data/EcDataOperation.java
```diff
@@ -104,12 +107,24 @@ private static int ecTypeToNRows(int ecType, final Bytes input) {
...
-    return false; // TODO:
+  private static boolean isOnG2(Bytes xBytes) {
+    final BigInteger pXIm = extractParameter(xBytes.slice(0, 32));
+    final BigInteger pXRe = extractParameter(xBytes.slice(32, 32));
+    final BigInteger pYIm = extractParameter(xBytes.slice(64, 32));
+    final BigInteger pYRe = extractParameter(xBytes.slice(96, 32));
+    final Fq2 pX = Fq2.create(pXRe, pXIm);
+    final Fq2 pY = Fq2.create(pYRe, pYIm);
+    final AltBn128Fq2Point p2 = new AltBn128Fq2Point(pX, pY);
+    final AltBn128Fq2Point pPowQ = p2.multiply(Q);
```
OlivierBBB, line 125: Maybe we can request @yelhousni 's review of that ? I do seem to remember that G2 is a torsion subgroup but if so I would expect the right check to be that `pPowQ` should be 0 (i.e. the point at infinity.)

delehef: You are absolutely right, I wrongly translated [the Go code](https://github.com/Consensys/zk-evm/blob/main/zeroknowledge/modules/ec_data/trace.go#L454).

arithmetization/src/main/java/net/consensys/linea/zktracer/module/ec_data/EcDataOperation.java
```diff
@@ -104,12 +107,24 @@ private static int ecTypeToNRows(int ecType, final Bytes input) {
     };
   }
 
-  private static boolean isOnG2(Bytes arg) {
-    return true; // TODO:
+  private static BigInteger extractParameter(final Bytes input) {
```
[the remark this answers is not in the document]
delehef: Fixed.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/ec_data/EcDataOperation.java
yelhousni: This looks correct after [aae0488](https://github.com/Consensys/besu-sequencer-plugins/pull/566/commits/aae048873280789156899a309eb47bc36885309e) and the [go code](https://github.com/Consensys/zk-evm/blob/f26595169dd65f78f6deb795db77df3e5a215203/zeroknowledge/modules/ec_data/trace.go#L446) too. However it is quite an optimised (in terms of constraints and also pure go performance). [Here](https://github.com/Consensys/gnark-crypto/blob/eb75782795d2d3ce6172797615f8eada77a6a3dd/ecc/bn254/g2.go#L410) we implement the membership test in Go with a cost dominated with a scalar multiplication with a 64 bits (instead of 254 bits in the naive method). The circuit version of this test can be found [here](https://github.com/Consensys/gnark/blob/2fbf083268f800d8a9c6079a5328ee774a350406/std/algebra/emulated/sw_bn254/pairing.go#L288).

yelhousni (approved):

PR #570, opened 2024-01-24
perf(EUC): one less column

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/euc/Trace.java
```diff
@@ -50,18 +49,17 @@ public class Trace {
...
     return List.of(
-        new ColumnHeader("euc.CEIL", 1, length),
+        new ColumnHeader("euc.CEIL", 8, length),
         new ColumnHeader("euc.CT", 1, length),
         new ColumnHeader("euc.CT_MAX", 1, length),
-        new ColumnHeader("euc.DIVIDEND", 1, length),
-        new ColumnHeader("euc.DIVIDEND_BYTE", 1, length),
-        new ColumnHeader("euc.DIVISOR", 1, length),
+        new ColumnHeader("euc.DIVIDEND", 8, length),
+        new ColumnHeader("euc.DIVISOR", 8, length),
```
delehef, line 56: There is an issue either with Corset or the constraints, those ought to be 1-byte wide, not 8.

delehef (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/euc/Trace.java
letypequividelespoubelles: nope, `DIVIDEND` and `DIVISOR` are eight bytes long, `DIVISOR_BYTE` is a byte

PR #571, opened 2024-01-24
Avoid reprocessing txs that go over line count limit

fab-10:
Besu has a cache to identify txs that have been already seen via p2p, so that is also a protection to avoid reprocessing bad txs, but that cache does not apply to txs sent by RPC, and that is the case of the txs sync script, and thus the script can resubmit bad txs at every run.
To be compatible with the txs sync script and prevent the reprocessing of bas txs, a cache has been introduced in the line count selector, to remember about these bad txs, and removed them from the pool without the need to process them again.
The cache is in memory only, so a restart cleans the cache, persistence of the cache can be added later if needed.

There is a new option to customize the mac size of the cache: `plugin-linea-over-line-count-limit-cache-size`; with a default of 10_000 txs.


.gitignore
```diff
@@ -32,3 +32,6 @@ site/
 /.envrc
 reference-tests/src/test/java/net/consensys/linea/generated/*
 !reference-tests/src/test/java/net/consensys/linea/generated/.gitkeep
+arithmetization/bin/
+acceptance-tests/bin/
+reference-tests/bin/
```
delehef, line 37: Do these folders exist on your computer?

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorConfiguration.java
```diff
@@ -15,128 +15,22 @@
 
 package net.consensys.linea.sequencer.txselection;
 
+import lombok.Builder;
+import lombok.Getter;
+
 /** The Linea configuration. */
+@Getter
+@Builder
 public final class LineaTransactionSelectorConfiguration {
```
delehef, line 24: Should this `class` be a `record`?

.gitignore
fab-10: yes I have these, not sure which gradle task creates them

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorConfiguration.java
fab-10: yes it can be, just found that Lombok builder works with records too, so we can have the benefit of the "named parameters constructor" offered by builders

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
```diff
@@ -151,6 +171,17 @@ public BlockAwareOperationTracer getOperationTracer() {
     return zkTracer;
   }
 
+  private void rememberOverLineCountLimitTransaction(final Transaction transaction) {
+    if (overLineCountLimitCache.size() > overLimitCacheSize) {
+      overLineCountLimitCache.iterator().remove();
```
jonesho, line 176: Why not `overLineCountLimitCache.size() >= overLimitCacheSize`, otherwise, it implies the `overLineCountLimitCache.size()` would not always be under or equal to `overLimitCacheSize`

jonesho: Also I see here's the only place we calls remove entry from `overLineCountLimitCache`, why we're not doing any removal in `onTransactionNotSelected` and `onTransactionSelected`?

jonesho: Oh, yes, they shouldn't be removed, we need to keep them in memory to avoid the sync scripts for resubmitting those txns

fab-10: Will fix the size check to be >≠.

About the second point, the behavior is what we need here. Because a bad tx in the cache will never be select, and there is no reason to remove in ‘onTransactionSelected’, and we need to remember it for the future blocks, so we cannot remove it in ’onTrandactionNotSelected’.

So we only remove a bad tx when the cache is full.

jonesho: Yes, @fab-10, I understood after I read it more carefully, that's why I marked as off topic, thanks for the explain anyway

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorConfiguration.java
fab-10: done

Gabriel-Trintinalia:
would you mind adding this new option to here: https://github.com/Consensys/besu-sequencer-plugins/blob/main/PLUGINS.md?

fab-10:
> would you mind adding this new option to here: https://github.com/Consensys/besu-sequencer-plugins/blob/main/PLUGINS.md? just some basic doco

Done

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
```diff
@@ -151,6 +171,21 @@ public BlockAwareOperationTracer getOperationTracer() {
     return zkTracer;
   }
 
+  private void rememberOverLineCountLimitTransaction(final Transaction transaction) {
+    if (overLineCountLimitCache.size() >= overLimitCacheSize) {
+      final var it = overLineCountLimitCache.iterator();
+      if (it.hasNext()) {
```
garyschulte, line 177: non-blocking, but IMO it would be more robust to make this a while loop on the off chance that the cache is more than one entry over the limit for any reason.

garyschulte (approved):
This LGTM, just one minor comment about only removing a single element from the cache when it is over the limit

arithmetization/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
fab-10: done

PR #581, opened 2024-02-05
feat[hub]: implement the precompile call mechanism & plug the OOB

delehef:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/call/oob/precompiles/ModexpCds.java
```diff
@@ -0,0 +1,41 @@
...
+package net.consensys.linea.zktracer.module.hub.fragment.imc.call.oob.precompiles;
+
+import static net.consensys.linea.zktracer.module.oob.Trace.OOB_INST_modexp_cds;
+import static net.consensys.linea.zktracer.types.Conversions.booleanToBytes;
+
+import net.consensys.linea.zktracer.module.hub.fragment.imc.call.oob.OobCall;
+import net.consensys.linea.zktracer.module.oob.OobDataChannel;
+import org.apache.tuweni.bytes.Bytes;
+
+public record ModexpCds(long callDataSize) implements OobCall {
```
powerslider, line 25: Can we rename it to `ModexpCallDataSize`?

delehef: All those are named to coincide with the spec

powerslider (approved):

PR #585, opened 2024-02-06
Linea estimate gas endpoint

fab-10:
Implements https://github.com/Consensys/protocol-misc/issues/867

This requires the update Linea Besu from this PR https://github.com/Consensys/linea-besu/pull/36

The estimate gas logic is taken from the standard `eth_estimateGas` method.

New options:

- `plugin-linea-estimate-gas-min-margin` implements: MINIMUM_MARGIN_ESTIMATE_GAS to recommend a specific gas price when using linea_estimateGas in the frontend nodes (Default "1.0")
- `plugin-linea-tx-compression-ratio` The ratio between tx serialized size and its compressed size (Default 5)

Some refactor of the way cli options and configuration are managed in the plugin, since now some of the options are used by multiple plugins and so that required to make them shared.
The profitability calculator has been moved to a class so it can be also shared between the new RPC endpoint and the tx selector plugin.

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
+++ b/arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
@@ -1,0 +1,130 @@
+package net.consensys.linea.bl;
+
+import java.math.BigDecimal;
+
+import lombok.extern.slf4j.Slf4j;
+import net.consensys.linea.config.LineaTransactionSelectorConfiguration;
+import org.hyperledger.besu.datatypes.Transaction;
+import org.hyperledger.besu.datatypes.Wei;
+import org.slf4j.spi.LoggingEventBuilder;
+
+@Slf4j
+public class TransactionProfitabilityCalculator {
+
+  private final LineaTransactionSelectorConfiguration conf;
+  private final double preComputedValue;
+
+  public TransactionProfitabilityCalculator(final LineaTransactionSelectorConfiguration conf) {
+    this.conf = conf;
+    this.preComputedValue =
+        conf.getEstimateGasMinMargin() * conf.getGasPriceRatio() * conf.getVerificationGasCost();
+  }
+
+  public Wei profitablePriorityFeePerGas(
+      final Transaction transaction, final Wei minGasPrice, final long gas) {
+    final double txSize = Math.max(0, transaction.getSize() + conf.getAdjustTxSize());
+    final double compressedSerializedSize = txSize / conf.getTxCompressionRatio();
+
+    final var profitAt =
+        preComputedValue
+            * txSize
+            * minGasPrice.getAsBigInteger().doubleValue()
+            / (gas * conf.getVerificationCapacity());
+
+    final var profitAtWei = Wei.ofNumber(BigDecimal.valueOf(profitAt).toBigInteger());
+
+    log.atDebug()
+        .setMessage(
+            "Estimated profitable priorityFeePerGas: {}; estimateGasMinMargin={}, gasPriceRatio={}, "
+                + "verificationGasCost={}, txSize={}, compressedTxSize={}, minGasPrice={}, gas={}, verificationCapacity={}")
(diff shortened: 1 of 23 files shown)


arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
```diff
@@ -0,0 +1,143 @@
...
+        .addArgument(conf.gasPriceRatio())
+        .addArgument(conf.verificationGasCost())
+        .addArgument(txSize)
+        .addArgument(compressedSerializedSize)
+        .addArgument(minGasPrice::toHumanReadableString)
+        .addArgument(gas)
+        .addArgument(conf.verificationCapacity())
+        .log();
+
+    return Wei.ofNumber(BigDecimal.valueOf(profitAt).toBigInteger());
```
jpnovais, line 65:
```suggestion
    return profitAtWei;
```

arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
```diff
@@ -0,0 +1,143 @@
...
+
+  public TransactionProfitabilityCalculator(final LineaTransactionSelectorConfiguration conf) {
+    this.conf = conf;
+    this.preComputedValue =
+        conf.estimateGasMinMargin() * conf.gasPriceRatio() * conf.verificationGasCost();
+  }
+
+  public Wei profitablePriorityFeePerGas(
+      final Transaction transaction, final Wei minGasPrice, final long gas) {
+    final double txSize = Math.max(0, transaction.getSize() + conf.adjustTxSize());
```
jpnovais, line 39: is there a scenario where `transaction.getSize()` or `conf.adjustTxSize()` can be negative?

Anyway it does not hurt.

arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
```diff
@@ -0,0 +1,143 @@
...
+
+  public boolean isProfitable(
+      final String step,
+      final Transaction transaction,
+      final double minGasPrice,
+      final double effectiveGasPrice,
+      final long gas) {
+    final double revenue = effectiveGasPrice * gas;
+
+    final double l1GasPrice = minGasPrice * conf.gasPriceRatio();
```
jpnovais, line 76: Note: my interpretation of `l1GasPrice = minGasPrice * conf.gasPriceRatio();` is that we consider the cost on L1 our minL2Gas price * ratio. I this this will yield Optimistic lowest cost on L1, which may lead false positive profit?

Given that L2 Gasprice (`effectiveGasPrice` here I think) flutuates regarding the L1Gasprice (updated by the coordinator) shouldn't we consider `effectiveGasPrice` for l1GasPrice? 

Or is this formula to amortize fluctuations?

** @fab-10  I am aware that this fucntion was already like this, just migrated to this new class. I am just raising this thought **

cc: @julien-marchand

jpnovais (approved):

arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
julien-marchand: conf.adjustTxSize has historically be a negative constant -45, if the config change to another value, it's safer to use the Max here

arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
julien-marchand: `effectiveGasPrice` is the actual gas price of the transaction (computed from base fee and priority fee), which doesn't inform much about the actual L1 gas price.

From what I understand minGasPrice is the actual value that fluctuates based on the coordinator info

julien-marchand (approved):

PR #590, opened 2024-02-09
Use compressed tx size also when selecting txs from block creation

fab-10:
@julien-marchand this PR uses the compressed tx size also when calculating the profitability during block creation, so it is aligned with the `linea_estimateGas` endpoint, is that correct?

FYI: @macfarla @pinges check `private double getCompressedTxSize(final Transaction transaction) {` for the integration of the `gnarkCompression`


fab-10:
relates to https://github.com/Consensys/protocol-misc/issues/860

julien-marchand:
> this PR uses the compressed tx size also when calculating the profitability during block creation, so it is aligned with the linea_estimateGas endpoint, is that correct?

Yes, txpool acceptance criteria (using `eth_sendRawTransaction` or `p2p`), `linea_estimateGas` and block creation should all use the same compression mechanism. It is just the margin config that should change.

From what I understand, this PR aligns the behavior, and the PR to transform this ratio (`size/5`) into the actual compression is pending

arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
```diff
@@ -55,8 +54,8 @@ public Wei profitablePriorityFeePerGas(
         .addArgument(conf.estimateGasMinMargin())
         .addArgument(conf.gasPriceRatio())
         .addArgument(conf.verificationGasCost())
-        .addArgument(txSize)
-        .addArgument(compressedSerializedSize)
+        .addArgument(transaction::getSize)
```
gauravahuja, line 57: This value will be different from the earlier logged `txSize` which also added `conf.adjustTxSize()`. Is that expected?

gauravahuja (approved):
lgtm, just the one question.

arithmetization/src/main/java/net/consensys/linea/bl/TransactionProfitabilityCalculator.java
fab-10: good point, I have aligned the fields in the logs

Issue #480, opened 2023-12-08
CLI options for L2L1 bridge


delehef:
L2L1 bridge contract address & log topic ought to be specified by CLI arguments rather than environment variables.

delehef:
@fab-10 could you please help with this one?

We would need the `ZkTracer` to potentially receive two CLI arguments:
 - `L2L1topic`: hex string;
 - `L2L1_CONTRACT_ADDRESS`: hex string.

The new `ZkTracer(final String l2l1ContractAddress, final String l2l1Topic)` constructor is ready on [this PR](https://github.com/Consensys/besu-sequencer-plugins/pull/591), would you mind taking it from there and adding the CLI switch?

delehef:
No

fab-10:
ok, so will make them mandatory

PR #591, opened 2024-02-12
Get L2L1 settings from ZkTracer constructor

delehef:
Addresses #480

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
@@ -58,6 +58,10 @@ public class ZkTracer implements ConflationAwareOperationTracer {
   private Hash hashOfLastTransactionTraced = Hash.EMPTY;
 
   public ZkTracer() {
+    new ZkTracer("0xDEADBEEF", "0x12345");
+  }
+
+  public ZkTracer(final String l2l1ContractAddress, final String l2l1Topic) {
     // Load opcodes configured in src/main/resources/opcodes.yml.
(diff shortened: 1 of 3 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -53,11 +55,21 @@ public class ZkTracer implements ConflationAwareOperationTracer {
...
   private final Map<String, Integer> spillings = new HashMap<>();
   private Hash hashOfLastTransactionTraced = Hash.EMPTY;
 
   public ZkTracer() {
+    this(
+        LineaL1L2BridgeConfiguration.builder()
+            .contract(Address.fromHexString("0xDEAD"))
+            .topic(Bytes.fromHexString("0xBEEF"))
+            .build());
+  }
```
fab-10, line 68: this should not be needed anymore

```suggestion
```

fab-10 (commented):
LGTM, just the removal of dead code

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
delehef: This is still used by all the places where the L2L1 bridge settings are useless, e.g. in the tracing endpoints, the continuous tracer, etc.

fab-10: replaced with a constant `LineaL1L2BridgeConfiguration.EMPTY`

fab-10 (approved):

PR #598, opened 2024-02-15
Avoid returing an estimated priority fee that is less than the min gas price

fab-10:
This is a safe net to avoid that the estimated priority fee + base fee returned can be less than the current min gas price


gauravahuja (approved):
lgtm

PR #604, opened 2024-02-19
move compress native into plugin repo

pinges:
This moves the native implementation here.

The native lib is built locally and added as a resource to the jar file (as well as the dict file).

When building the besu-plugins jar file now some additional tools are needed (like the go compiler)

Currently only the locally built native library is added to the jar file. For the release we need to make sure that the build is running on the right platform (I assume linux x86-64)


native/compress/build.gradle
```diff
@@ -0,0 +1,101 @@
...
+}
+
+dependencies {
+    implementation 'net.java.dev.jna:jna:5.12.1'
+    testImplementation 'net.java.dev.jna:jna:5.12.1'
+    implementation 'org.apache.tuweni:tuweni-bytes:2.2.0'
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.10.2'
+    testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.10.2'
+    testImplementation 'org.assertj:assertj-core:3.22.0'
+    testImplementation 'org.mockito:mockito-core:4.4.0'
```
fab-10, line 46: versions should go into `dependency-management.gradle`

fab-10 (commented):
LGTM, just one comment, and there are failing CI jobs

native/compress/build.gradle
pinges: done

fab-10 (approved):

PR #628, opened 2024-02-27
Use the right classloader to load the native library

fab-10:
LibCompress needs to use the classloader that knows about the plugin location when reading the native lib from the resources.
Use of log to print native lib loading status.


ahamlat (approved):
LGTM

PR #634, opened 2024-02-28
linea_estimateGas compatibility switch

fab-10:
Fixes https://github.com/Consensys/protocol-misc/issues/881

When the new option `plugin-linea-estimate-gas-compatibility-mode-enabled` is set to `true` then the `linea_estimateGas` API, instead of returning the result of the profitability formula, returns the min mineable gas price. By default the flag is set to `false`.


macfarla (approved):
LGTM

PR #650, opened 2024-03-21
Improve linea_estimateGas error response

fab-10:


native/Dockerfile-win-dockcross
```diff
@@ -1,9 +1,10 @@
-FROM dockcross/windows-shared-x64
+ARG IMAGE
+FROM dockcross/$IMAGE
 
 ARG GO_VERSION=1.22.1
 ARG GO_ARCHIVE_CHECKSUM=aab8e15785c997ae20f9c88422ee35d962c4562212bb0f879d052a35c8307c7f
 
-ENV DEFAULT_DOCKCROSS_IMAGE native-windows-cross-compile
+ENV DEFAULT_DOCKCROSS_IMAGE native-${IMAGE}-cross-compile
```
macfarla, line 7: @fab-10 is this change related?

fab-10: not required for the feature, but it enables building the linux native lib on Windows so it can be to run the plugin under WSL, will move it in a separate PR

arithmetization/src/main/java/net/consensys/linea/rpc/linea/LineaEstimateGas.java
```diff
@@ -200,26 +237,29 @@ private Long estimateGasUsed(
               return lowResult
                   .map(
                       lr -> {
-                        // if with the low estimation gas is successful the return this
-                        // estimation
+                        // if with the low estimation gas is successful the return this estimation
```
macfarla, line 240:
```suggestion
                        // if with the low estimation gas is successful then return this estimation
```

macfarla (approved):

PR #654, opened 2024-03-27
Rename to TransactionPoolValidator to make it clear it only apply to the txpool

fab-10:
Just renaming `TransactionValidator` to `TransactionPoolValidator` to make it clear that the validation is only applied when a tx is evaluated in the txpool


Gabriel-Trintinalia (approved):
Lgtm

PR #655, opened 2024-03-28
feat(shakira-data): implement SHAKIRA_DATA module

### LINKED ISSUE #624: Implement SHAKIRA_DATA module
There are in this module
- 3 different regimes (KECCAK, SHA2-256 and RIPEMD-160)
- every activation of one of these goes through 3 sub phases
  - DATA: variable number of rows (but at least 1)
  - RESULT: 2 rows always
  - EXTRA: 3 rows explained below
- the EXTRA phase is a little stupid it's just to make sure every activation produces at least 4 + 2 = 6 rows in total. Since DATA can be as short as 1 row and RESULT is alwasy 2 rows we need 3 extra rows
- let me explain the "4 + 2 = 6" comment, it comes from
  - 4 bytes required to justify smallness of ID increments
  - 2 bytes required to show smallness of the final nBytes value
- a lot of the time the EXTRA rows are just wasted space

Also INDEX_MAX of the data phase in only **implicitly defined in the spec**. It must be equal to
- 0 when `0 < SIZE <=16 `
- 1 when `16 < SIZE <= 32`
- 2 when `32 < SIZE <= 48`
- etc ...

In other words the INDEX_MAX of the data phase must be set to `(SIZE - 1) / 16`. Also this module is only activated when SIZE is nonzero.

### LINKED ISSUE #684: The SHAKIRA module requires at least 2 padding rows
# Main point

The lookup from the HUB to the SHAKIRA module which the HUB requires in order to extract
- the output of the `KECCAK` opcode (for unexceptional, nontrivial cases)
- the hash of the init code so as to serve the correct hash to the CREATE2 induced lookup to the RLP_ADDR module
- the code hash for `RETURN` opcodes (associated with unexceptional, nontrivial deployments)

requires the SHAKIRA module to have 2 padding rows since it extracts values from two consecutive rows of the SHAKIRA module, and expects to find, when its HUB selector is zero, two empty rows in SHAKIRA.

**Note.** Isse #685 claimed that the same should hold for the ADD module. This is a mistake on my end and is wrong. The reason it doesn't hold is that the two rows in question live in the **source** of the lookup rather than in the **target**.

powerslider:
Resolves: #624 
Signed-off-by: Tsvetan Dimitrov <tsvetan.dimitrov@consensys.net>


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blake2fmodexpdata/Blake2fModexpDataOperation.java
```diff
@@ -200,7 +200,7 @@ private UnsignedByte[] getHubStampDiffBytes(Bytes currentHubStamp) {
     BigInteger hubStampDiff = hubStampBigInt.subtract(prevHubStampBigInt).subtract(BigInteger.ONE);
 
     Preconditions.checkArgument(
-        hubStampDiff.compareTo(BigInteger.valueOf(256 ^ 6)) < 0,
+        hubStampDiff.compareTo(BigInteger.valueOf((long) Math.pow(256, 6))) < 0,
```
letypequividelespoubelles: not sure it's usefull as it will break anyway if the hubStamp is greater than 256**6

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/Sha256Blocks.java
```diff
@@ -45,6 +43,17 @@ public String moduleKey() {
   private static final int SHA256_PADDING_LENGTH = 64;
   private static final int SHA256_NB_PADDED_ONE = 1;
 
+  private final Hub hub;
+  private final Stack<Integer> counts = new Stack<>();
```
letypequividelespoubelles: It seems a Deque is preferable

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/RipeMd160Blocks.java
```diff
@@ -38,12 +50,9 @@ public String moduleKey() {
...
 
-  private static final int PRECOMPILE_BASE_GAS_FEE = 600;
-  private static final int PRECOMPILE_GAS_FEE_PER_EWORD = 120;
-  private static final int RIPEMD160_BLOCKSIZE = 64 * 8;
-  // If the length is > 2⁶4, we just use the lower 64 bits.
-  private static final int RIPEMD160_LENGTH_APPEND = 64;
-  private static final int RIPEMD160_ND_PADDED_ONE = 1;
+  @Getter private final ShakiraData data;
+
+  private int lastDataCallHubStamp = 0;
```
letypequividelespoubelles: It's very strange that this info is stored here ... Amha, the last hubStamp should be in ShakiraOperation only, with somethink like `previousHubStamp = shakiraStamp ==1 ? 0 : shakiraOperations.get(shakiraStamp-2).hubStamp()`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -54,14 +62,16 @@ public String moduleKey() {
 
   @Override
   public void enterTransaction() {
-    this.deployedCodesizes.push(new ArrayList<>(5));
+    this.deployedCodeSizes.push(new ArrayList<>(8));
```
letypequividelespoubelles: Why is the ArrayList given a fixed size ? Where does it come from ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -78,18 +88,39 @@ private static int numKeccak(long x) {
...
+    codeSizes.peek().add(codeSize);
+
+    if (codeSize != 0) {
+      final long codeOffset = Words.clampedToLong(frame.getStackItem(codeOffsetStackItemOffset));
+      final Bytes byteCode = frame.shadowReadMemory(codeOffset, codeSize);
 
-    // Capture contract deployment
-    if (opCode == OpCode.RETURN && hub.currentFrame().underDeployment() && frame.stackSize() > 1) {
-      final long codeSize = clampedToLong(frame.getStackItem(1));
-      this.deployedCodesizes.peek().add(codeSize);
```
letypequividelespoubelles, line 92: Where do you store the deployedCodeSize ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -78,18 +88,39 @@ private static int numKeccak(long x) {
   public void tracePreOpcode(final MessageFrame frame) {
     final OpCode opCode = this.hub.opCode();
 
-    // Capture calls to SHA3
-    if (opCode == OpCode.SHA3) {
-      if (frame.stackSize() > 1) {
-        final long sha3Size = Words.clampedToLong(frame.getStackItem(1));
-        this.sha3Sizes.peek().add(sha3Size);
```
letypequividelespoubelles, line 85: same, we still need to store the called size to meet the prover's limit

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/RipeMd160Blocks.java
```diff
@@ -19,17 +19,29 @@
...
+public final class RipeMd160Blocks implements Module {
+  private static final int PRECOMPILE_BASE_GAS_FEE = 600;
+  private static final int PRECOMPILE_GAS_FEE_PER_EWORD = 120;
+  private static final int RIPEMD160_BLOCKSIZE = 64 * 8;
+  // If the length is > 2⁶4, we just use the lower 64 bits.
+  private static final int RIPEMD160_LENGTH_APPEND = 64;
+  private static final int RIPEMD160_ND_PADDED_ONE = 1;
+
   private final Hub hub;
   private final Stack<Integer> counts = new Stack<>();
```
letypequividelespoubelles: prefer a Deque

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/RipeMd160Blocks.java
```diff
@@ -80,39 +87,43 @@ public static long gasCost(final Hub hub) {
...
-            this.counts.push(this.counts.pop() + blockCount);
-          }
+    if (opCode.isCall()) {
+      final Address target = Words.toAddress(frame.getStackItem(1));
+      if (target.equals(Address.RIPEMD160)) {
+        final long dataByteLength = hub.transients().op().callDataSegment().length();
+
+        if (dataByteLength == 0) {
+          return;
+        } // skip trivial hash TODO: check the prover does skip it
```
letypequividelespoubelles, line 104: did we check it ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/Sha256Blocks.java
```diff
@@ -45,6 +43,17 @@ public String moduleKey() {
...
+
+  @Getter private final ShakiraData data;
+  private int lastDataCallHubStamp = 0;
```
letypequividelespoubelles: Same, it's strange that it's stored here

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraData.java
```diff
@@ -0,0 +1,78 @@
...
+    return Trace.headers(this.lineCount());
+  }
+
+  public void call(final ShakiraDataOperation operation) {
+    this.state.add(operation);
+
+    operation.currentId(operation.hubStamp() + 1);
+    operation.prevId(lastDataCallId);
+
+    lastDataCallId = operation.currentId();
```
letypequividelespoubelles: What happens when you pop transaction ? I'm pretty sure it'll break. This must be computed at `commit` time

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -0,0 +1,284 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+import net.consensys.linea.zktracer.container.ModuleOperation;
+import net.consensys.linea.zktracer.types.UnsignedByte;
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.crypto.Hash;
+
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
```
letypequividelespoubelles: We don't need it, we can't have duplicate

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -0,0 +1,284 @@
...
+@Accessors(fluent = true)
+public class ShakiraDataOperation extends ModuleOperation {
+  private static final int PHASE_EXTRA = 0;
```
letypequividelespoubelles: Those constants should come from the lisp amha

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -0,0 +1,284 @@
...
+
+  @Override
+  protected int computeLineCount() {
+    return totalNumberOfRows();
+  }
+
+  void trace(Trace trace, int stamp) {
+    var tracerBuilder = ShakiraTraceHelper.builder().trace(trace).currentId(currentId).stamp(stamp);
+
+    switch (precompileType) {
```
letypequividelespoubelles: I don't understand why we duplicate for each precompile, we trace exactly the same for the three operation (only the flag and the way we compute the result differ)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -0,0 +1,284 @@
...
+            Trace.PHASE_KECCAK_RESULT,
+            Trace.INDEX_MAX_KECCAK_RESULT,
+            TOTAL_SIZE_RESULT_PHASES,
+            deltaBytesResultPhases),
+        PHASE_EXTRA,
+        new PhaseInfo(
+            PHASE_EXTRA, INDEX_MAX_EXTRA, TOTAL_SIZE_EXTRA_PHASES, deltaBytesExtraPhases));
+  }
+
+  private Bytes extractLimbRowData(Integer rowIndex) {
```
letypequividelespoubelles: An easier way could be to just store the rightPadded (to LLARGE * (input.size +LLARGEMO) / LLARGE) input so after you can do the slicing by LLARGE bytes without checking it at each row (or if you prefer, just check it for the last row, it's useless before)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -0,0 +1,284 @@
...
+
+    return !limbSizeIsLessThan16
+        ? hashInput.slice(startLimbOffsetIndex, Trace.LLARGE)
+        : rightPadTo(
+            hashInput.slice(startLimbOffsetIndex, inputDataSize - startLimbOffsetIndex),
+            Trace.LLARGE);
+  }
+
+  private UnsignedByte[] getHubStampDiffBytes(int currentHubStamp) {
+    final BigInteger prevHubStampBigInt = BigInteger.valueOf(prevId);
```
letypequividelespoubelles: you can just use long as hubStamp is & 6 bytes int at max

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -0,0 +1,284 @@
...
+        hubStampBigInt.subtract(prevHubStampBigInt).subtract(BigInteger.ONE);
+
+    Preconditions.checkArgument(
+        hubStampDiff.compareTo(BigInteger.valueOf((long) Math.pow(256, 4))) < 0,
+        "Hub stamp difference should never exceed 256 ^ 4");
+
+    return bytesToUnsignedBytes(leftPadTo(bigIntegerToBytes(hubStampDiff), 4).toArray());
+  }
+
+  private UnsignedByte[] getDataPhaseDeltaBytes() {
```
letypequividelespoubelles: Why do we do all of this ? DEltaByte will be always 0 during data phase, no ?

letypequividelespoubelles (changes requested):
I'm afraid the computation of the deltaByte might break if we pop transaction, not sure though. I think we don't have test for this right ? 

I'm pretty sure we could simplify a lot the tracing but not mandatory.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blake2fmodexpdata/Blake2fModexpDataOperation.java
OlivierBBB: Why would it break if the HUB stamp is larger than that ? Isn't it an int ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
OlivierBBB: IIRC they are minimum sizes to achieve:
- avoiding to copy the underlying array  too often as you add stuff in it (with the expectation that 8 items should be enough most of the time)
- not allocating too much space either.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -78,18 +88,39 @@ private static int numKeccak(long x) {
...
+        callShakira(frame, 1, 0, this.deployedCodeSizes);
+      }
+
+      if (opCode == OpCode.CREATE2 && pch.aborts().none()) {
+        callShakira(frame, 2, 1, this.create2Sizes);
       }
     }
+  }
+
+  private void callShakira(
```
OlivierBBB, line 108: Does anyone have her number ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/EcPairingCallEffectiveCall.java
```diff
@@ -104,18 +101,16 @@ public static long gasCost(final Hub hub) {
...
-            return 0;
-          }
-
-          return PRECOMPILE_BASE_GAS_FEE + PRECOMPILE_MILLER_LOOP_GAS_FEE * nMillerLoop;
+    if (opCode.isCall()) {
+      final Address target = Words.toAddress(frame.getStackItem(1));
+      if (target.equals(Address.ALTBN128_PAIRING)) {
+        final long length = hub.transients().op().callDataSegment().length();
+        final long nMillerLoop = (length / ECPAIRING_NB_BYTES_PER_MILLER_LOOP);
+        if (nMillerLoop * ECPAIRING_NB_BYTES_PER_MILLER_LOOP != length) {
```
OlivierBBB, line 109: I'm curious as to why we don't rather start with e.g.
```java
if (length % ECPAIRING_NB_BYTES_PER_MILLER_LOOP != 0) { return 0; }

// then resume operations
final long nMillerLoop = (length / ECPAIRING_NB_BYTES_PER_MILLER_LOOP) // etc ..
```

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/RipeMd160Blocks.java
OlivierBBB: I concur with @letypequividelespoubelles in the observation that it's the Shakira class that ought to know the previous ID (rather than objects related to either KECCAK, SHA2-256 or RIPEMD-160. It's module wide shared information and bears no particular connection to the nature of the operation.

On a different note I believe we don't care about the `previousHubStamp` as much as we care about the `previousId`. In any case the SHAKIRA module knows only about the ID.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/RipeMd160Blocks.java
OlivierBBB: The prover should only prove what we ask it to prove. Do we need to check anything ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/Sha256Blocks.java
OlivierBBB: There really should only be a single kind of SHAKIRA operation with some field with values in an `enum` that specifies the underlying operation i.e. either `KECCAK`, `SHA2-256` or `RIPEMD-160`. Everything is derivable from that.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
OlivierBBB: @powerslider we discussed the fact that this should be unified on our last call.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
OlivierBBB: > DEltaByte will be always 0 during data phase, no ?

No, `deltaBytes` is an array of length `L` where
```java
L = ceil(size / 16) + 2 + 3
// ceil(size / 16) many data rows
// 2 result rows
// 3 'extra' rows
```
The contents of `deltaBytes` are all zero except (potentially) for the first 4 bytes and the last 2. The first byte (if not more) is always on a data row.

**Note:** Since we deal with nonempty data `L >= 6` so there is always enough room. This is the purpose of the 'extra' rows.

OlivierBBB (changes requested):
We should unify the tracing + have a single source of the previous ID (which would likely be the only natural thing to do when tracing is unified.)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blake2fmodexpdata/Blake2fModexpDataOperation.java
letypequividelespoubelles: HUB STAMP / CONTEXT_NUMBER columns are typed as an `:i48`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/EcPairingCallEffectiveCall.java
letypequividelespoubelles: indeed, done

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
letypequividelespoubelles: yes, but the 8, 100, 4, where this values come from ? @powerslider  ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blake2fmodexpdata/Blake2fModexpDataOperation.java
```diff
@@ -57,14 +57,20 @@ public class Blake2fModexpDataOperation extends ModuleOperation {
...
+          PHASE_MODEXP_BASE,
+          new PhaseInfo(PHASE_MODEXP_BASE, Trace.INDEX_MAX_MODEXP_BASE),
           PHASE_MODEXP_EXPONENT,
-              new PhaseInfo(PHASE_MODEXP_EXPONENT, Trace.INDEX_MAX_MODEXP_EXPONENT),
-          PHASE_MODEXP_MODULUS, new PhaseInfo(PHASE_MODEXP_MODULUS, Trace.INDEX_MAX_MODEXP_MODULUS),
-          PHASE_MODEXP_RESULT, new PhaseInfo(PHASE_MODEXP_RESULT, Trace.INDEX_MAX_MODEXP_RESULT),
-          PHASE_BLAKE_DATA, new PhaseInfo(PHASE_BLAKE_DATA, Trace.INDEX_MAX_BLAKE_DATA),
-          PHASE_BLAKE_PARAMS, new PhaseInfo(PHASE_BLAKE_PARAMS, Trace.INDEX_MAX_BLAKE_PARAMS),
-          PHASE_BLAKE_RESULT, new PhaseInfo(PHASE_BLAKE_RESULT, Trace.INDEX_MAX_BLAKE_RESULT));
+          new PhaseInfo(PHASE_MODEXP_EXPONENT, Trace.INDEX_MAX_MODEXP_EXPONENT),
```
OlivierBBB, line 63: Purely formatting ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -119,7 +119,7 @@ public boolean callLT(final Bytes arg1, final Bytes arg2) {
     return this.callLT(Bytes32.leftPad(arg1), Bytes32.leftPad(arg2));
   }
 
-  public boolean callLT(final int arg1, final int arg2) {
+  public boolean callLT(final long arg1, final long arg2) {
```
OlivierBBB, line 122: There are places where we want to `callLt` with stack arguments (e.g. in the OOB module), should we not allow for `BigIntegers` here ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -159,6 +159,10 @@ public boolean callLEQ(final Bytes32 arg1, final Bytes32 arg2) {
     return arg1.compareTo(arg2) <= 0;
   }
 
+  public boolean callLEQ(final long arg1, final long arg2) {
```
OlivierBBB, line 162: itou

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -78,18 +87,40 @@ private static int numKeccak(long x) {
...
+      final MessageFrame frame,
+      final int codeSizeStackItemOffset,
+      final int codeOffsetStackItemOffset,
```
OlivierBBB, line 110: On the stack the order is always  `offset, size` it would be more natural to order things accordingly in `callShakira`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -78,18 +87,40 @@ private static int numKeccak(long x) {
...
+    final PlatformController pch = this.hub.pch();
+
+    if (pch.exceptions().none()) {
+      // Capture calls to SHA3.
+      if (opCode == OpCode.SHA3) {
+        callShakira(frame, 1, 0, this.sha3Sizes);
+      }
+
+      // Capture contract deployment
+      // TODO: compute the gas cost if we are under deployment.
```
OlivierBBB, line 97: IIRC in the MXP module we already compute the gas cost which the deployment would incur. So `(pch.exceptions().none())` should be enough. If it isn't part of the MXPX exception detection / OOGX exception detection then the issue is with the platform controller.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/EcAddEffectiveCall.java
```diff
@@ -59,13 +59,11 @@ public void popTransaction() {
...
     final OpCode opCode = hub.opCode();
 
-    switch (opCode) {
-      case CALL, STATICCALL, DELEGATECALL, CALLCODE -> {
-        final Address target = Words.toAddress(frame.getStackItem(1));
-        if (target.equals(Address.ALTBN128_ADD)
-            && hub.transients().op().gasAllowanceForCall() >= PRECOMPILE_GAS_FEE) {
-          this.counts.push(this.counts.pop() + 1);
-        }
+    if (opCode.isCall()) {
```
OlivierBBB, line 62: In actuality we should also check that CALL_DATA_SIZE is nonzero. Indeed the arithmetization doesn't trigger ECDATA otherwise.

OlivierBBB: The same comment applies for
- ECRECOVER
- ECMUL
- ECPAIRING

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  private Bytes computeResult() {
+    return switch (precompileType) {
+      case SHA256 -> Hash.sha256(hashInput);
+      case KECCAK -> Hash.keccak256(hashInput);
+      case RIPEMD -> Hash.ripemd160(hashInput);
+    };
+  }
+}
```
OlivierBBB, line 167: Nice and clean

OlivierBBB (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blake2fmodexpdata/Blake2fModexpDataOperation.java
letypequividelespoubelles: Idk, anyway it'll be deleted in the PR fixing issues in blakemodexp

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
letypequividelespoubelles: we have sereveral `callLT`, for long, Nytes, Bytes32, if needed fot a BigInt, we just add it

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/EcAddEffectiveCall.java
letypequividelespoubelles: we have to review all the precompile triggering anyway, and maybe trigger them with the HUB, not at `tracePreOpcode` ... cf https://github.com/Consensys/linea-arithmetization/issues/686

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
letypequividelespoubelles: same, cf https://github.com/Consensys/linea-arithmetization/issues/686

letypequividelespoubelles (approved):

PR #657, opened 2024-03-29
Temporary hacky way to return eth_gasPrice in linea_estimateGas

fab-10:
Returning the `eth_gasPrice` value when `linea_estimateGas` is in compatibility mode, making an internal call to local RPC endpoint once per block.


julien-marchand (approved):

powerslider:
@fab-10 is it ok to close this PR in linea-arithmetization?

fab-10:
yes, closing it since no more relevant

PR #693, opened 2024-04-24
feat(rlpaddr): add lookup to trm

### LINKED ISSUE #690: `RLPADDR` must trim address
spec: https://github.com/Consensys/linea-spec/issues/412

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpaddr/RlpAddr.java
```diff
@@ -168,10 +182,10 @@ private void traceCreate2(int stamp, RlpAddrChunk chunk, Trace trace) {
   }
 
   private void traceCreate(int stamp, RlpAddrChunk chunk, Trace trace) {
-    final int RECIPE1_CT_MAX = 8;
+    final int RECIPE1_NB_ROWS = MAX_CT_CREATE + 1;
```
powerslider: This should be camelCase, because it is a local variable.

powerslider (approved):

PR #699, opened 2024-05-03
feat: add mechanism to count rows not yet in traced operation

### LINKED ISSUE #698: add mechanism to count rows that are not yet traced
If  `traceEndConflation` or `traceEndBlock` add rows in a module, the sequencer won't know it as he computes the number of lines after `traceEndTransaction`, and thus could lead to overflow.

So we must have a separate counter for rows added after `traceEndTransaction`

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -34,12 +34,20 @@
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
 import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.worldstate.WorldView;
 
-@RequiredArgsConstructor
 public class Wcp implements Module {
   private final StackedSet<WcpOperation> operations = new StackedSet<>();
+  /** count the number of rows that could be added after the sequencer counts the number of line */
+  public final Stack<Integer> additionalRows = new Stack<>();
```
ahamlat: Consider using `Deque<Integer> stack = new ArrayDeque<>();`, Stack extends Vector, thus not efficient at all.

ahamlat (changes requested):
Does it make sens to store the additional lines as additionalRows grow to avoid iterating over it when calculating line counting ?

letypequividelespoubelles:
> Does it make sens to store the additional lines as additionalRows grow to avoid iterating over it when calculating line counting ?

yes absolutely it does, done.

powerslider (approved):

ahamlat (approved):
LGTM

PR #709, opened 2024-05-15
fix(exp): kill duplicate

### LINKED ISSUE #695: `EXP` creates doublon instruction in its trace

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/Exp.java
lorenzogentile404, line 1: Can we delete `tracePreOpcode` because of the presence of `callExpLogCall` and `callModExpLogCall`?

lorenzogentile404, line 1: Can we delete `tracePreOpcode` because of the presence of `callExpLogCall` and `callModExpLogCall`?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ModExpLogChunk.java
lorenzogentile404, line 1: Thanks for fixing the lookup usage. The mistake of the call without assignment and duplication of logic was a consequence of my misunderstanding of the purpose of the call (i.e., I believed it was for tracing the lookup, but I can see it is simply for delegating computation). I likely need to fix that also in other modules I implemented.

lorenzogentile404 (changes requested):
- ~~If I understand correctly, in case of identical usage of the `EXP` opcode, we do not recreate from scratch the same `ExpLogChunk` and trace it only once?~~
- ~~Why we did this optimisation specifically here and not whenever we haveidentical usages of opcodes (or precompiles)? Is this simply a work in progress in other places?~~ 

These questions were already addressed during call.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ExpLogChunk.java
```diff
@@ -74,11 +65,9 @@ public void preCompute() {
     pPreprocessingWcpArg2Hi[0] = Bytes.EMPTY;
     pPreprocessingWcpArg2Lo[0] = Bytes.EMPTY;
     pPreprocessingWcpInst[0] = UnsignedByte.of(ISZERO);
-    pPreprocessingWcpRes[0] = this.exponent.hi().isZero();
-    final boolean expnHiIsZero = pPreprocessingWcpRes[0];
-
-    // Lookup
-    wcp.callISZERO(this.exponent.hi());
+    final boolean expnHiIsZero = wcp.callISZERO(this.exponent.hi());
+    ;
```
OlivierBBB, line 69: is this lone semicolon a typo ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ModExpLogChunk.java
```diff
@@ -29,15 +30,15 @@
...
 import lombok.RequiredArgsConstructor;
 import net.consensys.linea.zktracer.module.hub.fragment.imc.call.oob.ModExpLogCall;
 import net.consensys.linea.zktracer.module.wcp.Wcp;
-import net.consensys.linea.zktracer.opcode.OpCode;
 import net.consensys.linea.zktracer.types.EWord;
 import net.consensys.linea.zktracer.types.UnsignedByte;
 import org.apache.tuweni.bytes.Bytes;
-import org.hyperledger.besu.evm.frame.MessageFrame;
 
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
```
OlivierBBB, line 41: Does this work, then ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/Exp.java
letypequividelespoubelles: yes, the hub already triggers `EXP` when needed. So if you trigger `EXP` at every opcode and if it's the right one, etc, you just do everything twice

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ModExpLogChunk.java
letypequividelespoubelles: it is indeed for the lookup, but if you A calls several time  B with the excat same input, then you only need it once in B

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ExpLogChunk.java
letypequividelespoubelles: yes, deleted

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ModExpLogChunk.java
letypequividelespoubelles: now yes (stackedList -> StackedSet)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ModExpLogChunk.java
```diff
@@ -92,83 +93,9 @@ public static ModExpLogChunk fromExpLogCall(final Wcp wcp, final ModExpLogCall c
     return modExpLogChunk;
   }
 
-  public static ModExpLogChunk fromFrame(final Wcp wcp, final MessageFrame frame) {
```
OlivierBBB, line 95: Where did this Chunk builder function go ?

letypequividelespoubelles: deleted, it was never called. The hub fully triggers exp (as before it was triggered by the hub AND by tracePreOpcode() for exp)

powerslider (approved):

PR #714, opened 2024-05-15
feat: replace all places where global constants are referred with references to newly generated GlobalConstants class

powerslider:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/call/oob/opcodes/DeploymentReturn.java
lorenzogentile404, line 1: Does `deployment` replace `return`?

lorenzogentile404 (commented):
LGTM, specifically `oob` and `mxp` where I am most familiar. Do you have a list of the substitution you made (e.g., `Trace.oob.EVM_INST_L`T -> `GlobalConstants.EVM_INST_LT`). It would be way easier to review it if you made something like that or you made a little script for substitution.

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/call/oob/opcodes/DeploymentReturn.java
OlivierBBB: Yes,  `OOB_INST_deployment` the name in the spec, too.

PR #720, opened 2024-05-20
Improve ZkTracer initialization time

ahamlat:
When doing load testing on eth_sendRawTransaction with transaction simulation enabled, we've noticed that ZkTracer takes more than 20% of the time, on loading the opcode and the spillings.

image
This RP Improves ZkTracer initialization time by doing only once Opcodes and spillings loading from disk resources. This is done when the class is loaded, opcodes and spillings are initialized is a static way.

Fixes issue https://github.com/Consensys/linea-arithmetization/issues/716


ahamlat:
@letypequividelespoubelles @powerslider This PR is already installed on different environment with the other release process. Can we merge it with this new version ?

letypequividelespoubelles:
I'll let @powerslider or @OlivierBBB answer as I'm off and idk where we are in merging our hub PR which might creates new conflicts

PR #726, opened 2024-05-23
fix(shfreftable): turned MSHP into an UnsignedByte

### LINKED ISSUE #725: fix(shfreftable): regenerate Trace.java since MSHP column is 16 bits while in constraints is 8 bits only

lorenzogentile404:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/tables/shf/Trace.java
```diff
@@ -184,7 +184,7 @@ public Trace fillAndValidateRow() {
     }
 
     if (!filled.get(3)) {
-      mshp.position(mshp.position() + 2);
+      mshp.position(mshp.position() + 1);
```
OlivierBBB, line 187: What does this do ?
I imagine that the 2 -> 1 here is the same as the 2 -> 1 change in the ColumnHeader for MSHP ?

OlivierBBB (approved):
Downstream changes from having µMSHP going from `:i8` to `:byte` in the lisp. LGTM.
Also see https://github.com/Consensys/corset/issues/158

arithmetization/src/main/java/net/consensys/linea/zktracer/module/tables/shf/Trace.java
letypequividelespoubelles: yes, its is. Generated by corset.

letypequividelespoubelles (approved):

PR #735, opened 2024-05-27
fix(mxp): used constants

### LINKED ISSUE #734: fix(mxp): used constants

lorenzogentile404:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpData.java
```diff
@@ -380,11 +381,11 @@ private mxpExecutionPath getMxpExecutionPath() {
     return mxpExecutionPath.NON_TRIVIAL;
   }
 
-  public int maxCt() {
+  public int ctMax() {
```
letypequividelespoubelles, line 384: isn't it nb of rows instead ?

letypequividelespoubelles (approved):

PR #741, opened 2024-05-29
Fix/issue 729/few fixes in modexp data

### LINKED ISSUE #729: fix: previous hubStamp in MODEXP_DATA

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blake2fmodexpdata/Blake2fModexpData.java
```diff
@@ -40,35 +51,59 @@ public String moduleKey() {
...
+
+  @Override
+  public void traceStartTx(WorldView worldView, Transaction tx) {
+    this.numberOfOperationsAtStartTx = operations.size();
   }
 
   @Override
   public void popTransaction() {
-    this.state.pop();
+    this.sortedOperations.removeAll(this.operations.sets.getLast());
```
lorenzogentile404, line 84: Could you add a comment about what is the logic here?

lorenzogentile404 (changes requested):

powerslider (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blake2fmodexpdata/Blake2fModexpData.java
letypequividelespoubelles: logic is the same as in shakira:

- during the tx we store the op in the stackedSet
- at the end of the tx, we 1) sort (by frowing id) the op of this tx
-  2) add this tx to sortedOperation

3) trigger wcp to prove the sorting



- if we pop a tx, we have to delete the op added by this tx, this is done by `
    this.sortedOperations.removeAll(this.operations.sets.getLast());`

lorenzogentile404 (approved):

PR #744, opened 2024-05-30
perf: line count only once

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/set/StackedSet.java
```diff
@@ -39,15 +39,21 @@
 public class StackedSet<E extends ModuleOperation> implements StackedContainer, java.util.Set<E> {
   public final Deque<Set<E>> sets = new ArrayDeque<>();
   private final Map<E, Integer> occurrences = new HashMap<>();
+  private final Deque<Integer> cumulativeLineCount = new ArrayDeque<>() {{add(0);}};
```
ahamlat, line 42: Not sure if a Deque is needed here. What about keeping track of the last value and the cumulative one. If we need to rollback the last transaction, we can just remove the last value from the cumulative one.
This works also for the size.

ahamlat (approved):

PR #759, opened 2024-06-06
feat: Update `CorsetValidator` with Expansion Options

DavePearce:
This updates `CorsetValidator` to use a better set of defaults, including field arithmetic and expansion of constraints to a lower level.  This result is more of the `corset` tool being tested, helping us ensure its ready.   

NOTE: expansion does not (by default) go to the lowest level (because this has a large impact on performance).  Nevertheless, the option to do this is present in the updated `CorsetValidator` and can be enabled easily enough.


.github/workflows/gradle.yml
```diff
@@ -84,6 +84,7 @@ jobs:
         run: ./gradlew :arithmetization:test
         env:
           JAVA_OPTS: -Dorg.gradle.daemon=false
+          CORSET_FLAGS: fields;expand;expand;expand
```
DavePearce, line 87: @powerslider This is how you configure `corset` in a github action.

powerslider (approved):

PR #765, opened 2024-06-06
fix(exp and mxp and oob and ecdata): turned toBigInteger() into toUnsignedBigInteger

### LINKED ISSUE #762: unsignedBigInteger
Unless you're sure you want signed number, don't use `toBigInteger()` but  `.toUnsignedBigInteger()`

Seen in MXP, ECDATA, OOB

lorenzogentile404:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ExpOperation.java
```diff
@@ -91,7 +91,7 @@ final void traceComputation(int stamp, Trace trace) {
       pComputationMsb
       */
       // tanzb turns to 1 iff trimAcc is nonzero
-      tanzb = pComputationTrimAcc.slice(0, i + 1).toBigInteger().signum() != 0;
+      tanzb = pComputationTrimAcc.slice(0, i + 1).toUnsignedBigInteger().signum() != 0;
```
letypequividelespoubelles, line 94: it seems here you really want a signed bigInt as you ask for its sign no ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/OobOperation.java
```diff
@@ -191,7 +191,7 @@ public OobOperation(
   private void setOpCodeFlagsAndWghtSumAndIncomingInst(MessageFrame frame) {
     final OpCode opCode = OpCode.of(frame.getCurrentOperation().getOpcode());
     // In the case of CALLs and CREATEs this value will be replaced
-    wghtSum = UnsignedByte.of(opCode.byteValue()).toBigInteger();
+    wghtSum = BigInteger.valueOf(Byte.toUnsignedInt(opCode.byteValue()));
```
letypequividelespoubelles, line 194: sounds a complicate way of life to get it :)

letypequividelespoubelles (changes requested):
it seems you need to keep a signed BigInt for exp ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ExpOperation.java
lorenzogentile404: No, you just want to distinguish between `0` and `!= 0`. The alternative would be `!...toUnsignedBigInteger().compareTo(BigInteger.ZERO)` or creating some dedicated method for it.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/OobOperation.java
lorenzogentile404: Both of them seem to work and are comparable in terms of readability. Do you have suggestions for something better?

lorenzogentile404: at some point `wghtSum` is assigned to `oobInst`, that is expected to be a `bigInteger`, so even if I change the type of `wghtSum` to something more convenient, I need that conversion anyway later.

lorenzogentile404: Sorry, I am wrong. `oobInst` is now `Bytes`, so potentially we may convert both of them to `Bytes`.

lorenzogentile404: `wghtSum = Bytes.fromHexString("FF08").toUnsignedBigInteger();` due to this it cannot simply be a byte. I would suggest to leave it as it is.

letypequividelespoubelles (approved):
LGTM

PR #772, opened 2024-06-11
Update besu to 24.6-develop-711dff7

daniellehrner:


powerslider:
@daniellehrner This update will require migrating the project to java 21.

daniellehrner:
> @daniellehrner This update will require migrating the project to java 21.

I did the update now. Unfortunately it requires an update of the linting lib as well, which has changed some linting rules. Many files have some changed formatting, but are otherwise not affected

powerslider (approved):

PR #800, opened 2024-06-26
Refactor Exceptions class - reduce number of allocations

lu-pinto:
By looking at the `Exceptions` class I cannot see any change of state that's not through the `prepare` call  which is only called once. This leads me to believe that we don't need to create new `Exceptions` objects every time with `snapshot`.

Also `Exceptions` can only have 1 boolean true at once, by looking at `prepare` so an `Enum` is the most suitable data structure for this case.

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/constants/GlobalConstants.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/constants/GlobalConstants.java
@@ -25,7 +25,8 @@ import java.math.BigInteger;
  */
 public class GlobalConstants {
   public static final int BLOCKHASH_MAX_HISTORY = 0x100;
-  public static final int EIP_3541_MARKER = 0xef;
+  public static final byte EIP_3541_MARKER = (byte) 0xef;
+  public static final int MAX_CODE_SIZE = 24576;
   public static final BigInteger EMPTY_KECCAK_HI =
       new BigInteger("16434357337474432580558001204043214908");
   public static final BigInteger EMPTY_KECCAK_LO =
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
@@ -16,10 +16,8 @@
 package net.consensys.linea.zktracer.module.hub.signals;
 
 import static net.consensys.linea.zktracer.module.constants.GlobalConstants.EIP_3541_MARKER;
+import static net.consensys.linea.zktracer.module.constants.GlobalConstants.MAX_CODE_SIZE;
 
-import lombok.Getter;
-import lombok.RequiredArgsConstructor;
-import lombok.experimental.Accessors;
 import net.consensys.linea.zktracer.module.hub.Hub;
 import net.consensys.linea.zktracer.opcode.OpCode;
 import net.consensys.linea.zktracer.opcode.OpCodeData;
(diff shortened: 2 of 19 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
```diff
@@ -29,118 +27,35 @@
...
-    this.outOfMemoryExpansion = outOfMemoryExpansion;
-    this.outOfGas = outOfGas;
-    this.returnDataCopyFault = returnDataCopyFault;
-    this.jumpFault = jumpFault;
-    this.staticFault = staticFault;
-    this.outOfSStore = outOfSStore;
-    this.invalidCodePrefix = invalidCodePrefix;
-    this.codeSizeOverflow = codeSizeOverflow;
+/** Encode the exceptions that may be triggered by the execution of an instruction. */
+public enum Exceptions {
```
ahamlat: I suggest to make it singular, i.e Exception

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Signals.java
```diff
@@ -142,20 +142,23 @@ public void prepare(MessageFrame frame, PlatformController platformController, H
 
     switch (opCode) {
       case CALLDATACOPY, CODECOPY -> {
-        this.mxp = ex.outOfMemoryExpansion() || ex.outOfGas() || ex.none();
+        this.mxp =
+            ex == Exceptions.OUT_OF_MEMORY_EXPANSION || ex == Exceptions.OUT_OF_GAS || ex.none();
```
ahamlat: Maybe add parenthesis to make it more readable. 
Also, why not use Exceptions.NONE instead of ex.none() for more readability ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Signals.java
```diff
@@ -142,20 +142,23 @@ public void prepare(MessageFrame frame, PlatformController platformController, H
...
         this.mmu = ex.none() && !frame.getStackItem(2).isZero();
       }
 
       case RETURNDATACOPY -> {
-        this.oob = ex.none() || ex.returnDataCopyFault();
-        this.mxp = ex.none() || ex.outOfMemoryExpansion() || ex.outOfGas();
+        this.oob = ex.none() || ex == Exceptions.RETURN_DATA_COPY_FAULT;
+        this.mxp =
+            ex.none() || ex == Exceptions.OUT_OF_MEMORY_EXPANSION || ex == Exceptions.OUT_OF_GAS;
```
ahamlat: The same as above.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Signals.java
```diff
@@ -142,20 +142,23 @@ public void prepare(MessageFrame frame, PlatformController platformController, H
...
         this.mmu = ex.none() && !frame.getStackItem(2).isZero();
       }
 
       case EXTCODECOPY -> {
         final boolean nonzeroSize = !frame.getStackItem(3).isZero();
-        this.mxp = ex.outOfMemoryExpansion() || ex.outOfGas() || ex.none();
-        this.trm = ex.outOfGas() || ex.none();
+        this.mxp =
+            ex == Exceptions.OUT_OF_MEMORY_EXPANSION || ex == Exceptions.OUT_OF_GAS || ex.none();
```
ahamlat: The same as above.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Signals.java
```diff
@@ -168,7 +171,8 @@ public void prepare(MessageFrame frame, PlatformController platformController, H
       }
 
       case LOG0, LOG1, LOG2, LOG3, LOG4 -> {
-        this.mxp = ex.outOfMemoryExpansion() || ex.outOfGas() || ex.none();
+        this.mxp =
+            ex == Exceptions.OUT_OF_MEMORY_EXPANSION || ex == Exceptions.OUT_OF_GAS || ex.none();
```
ahamlat: The same as above.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Signals.java
```diff
@@ -179,10 +183,10 @@ public void prepare(MessageFrame frame, PlatformController platformController, H
       }
 
       case CALL, DELEGATECALL, STATICCALL, CALLCODE -> {
-        this.mxp = !ex.staticFault();
-        this.stp = ex.outOfGas() || ex.none();
-        this.oob = opCode.equals(OpCode.CALL) && ex.staticFault() || ex.none();
-        this.trm = ex.outOfGas() || ex.none();
+        this.mxp = ex != Exceptions.STATIC_FAULT;
```
ahamlat: I suggest to use parenthesis for more readability

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Signals.java
```diff
@@ -210,8 +214,8 @@ public void prepare(MessageFrame frame, PlatformController platformController, H
         final boolean nonzeroSize = !frame.getStackItem(2).isZero();
         final boolean isCreate2 = opCode == OpCode.CREATE2;
 
-        this.mxp = !ex.staticFault();
-        this.stp = ex.outOfGas() || ex.none();
+        this.mxp = ex != Exceptions.STATIC_FAULT;
+        this.stp = ex == Exceptions.OUT_OF_GAS || ex.none();
```
ahamlat: The same as above.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Signals.java
```diff
@@ -234,10 +239,13 @@ public void prepare(MessageFrame frame, PlatformController platformController, H
 
         // WARN: Static part, other modules may be dynamically requested in the hub
         this.mxp =
-            ex.outOfMemoryExpansion() || ex.outOfGas() || ex.invalidCodePrefix() || ex.none();
-        this.oob = isDeployment && (ex.codeSizeOverflow() || ex.none());
+            ex == Exceptions.OUT_OF_MEMORY_EXPANSION
```
ahamlat: The same as above

gradle/common-dependencies.gradle
```diff
@@ -30,6 +30,8 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-params'
   testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
 
+  testImplementation 'org.assertj:assertj-core'
```
ahamlat: Why do we need a new assertion dependency ? jupiter already exposes assertEquals

ahamlat (changes requested):
Nice PR @lu-pinto.
Just few comments for more readability and the add of the new dependency

OlivierBBB:
> Also Exceptions can only have 1 boolean true at once, by looking at prepare so an Enum is the most suitable data structure for this case.

Could you shed light on this please ? What makes it so that only one may be active ?

Having at most one active exception is indeed what is intended when _tracing_ exceptions at the end (i.e. when filling the stack columns for exceptions --- at most one may be on at any point in time) but AFAIK the exceptions _object_ may have several exceptions active at any point in time. E.g. a `staticException` + the `sstoreException` (when the available gas prior to executing an SSTORE is $\leq G_\text{callstipend} = 2300$) for an SSTORE instruction.

My understanding was that we were logging all possible exceptions and then deciding a posteriori which to act upon, in any case that was what I remember explaining to Franklin who implemented this stuff. With some caveats e.g. the `memoryExpansionException`, a subexception of the `outOfGasException`, an invention of our arithmetization, would very likely always be disjoint.

gradle/common-dependencies.gradle
lu-pinto: I think the biggest difference is that `assertj` provides more context in assertions so that's what I typically use. Another benefit is the enhanced dev experience in tests. note: other projects in this repo are already using this, just not this one in particular.

ref: https://medium.com/nerd-for-tech/junit-vs-assertj-choosing-the-right-testing-framework-for-your-java-project-84d2664736dc

EDIT: I actually ended up removing the dependency in my new changes as I don't need it anymore.

lu-pinto:
You're right @OlivierBBB, it was my oversight. Indeed `prepare` is being called before every opcode execution so all exceptions will be collected within `Exceptions` once the `MessageFrame` has finished executing. I reworked my code to now be (I believe) side effect free using a bit mask to record multiple exception types.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/constants/GlobalConstants.java
letypequividelespoubelles, line 1: Do not modify this file, it's generated automatically (see trace-files.gradle)

letypequividelespoubelles (changes requested):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/constants/GlobalConstants.java
letypequividelespoubelles: If you want to add a constant there, you have to add this constant in this file https://github.com/Consensys/linea-constraints/blob/master/constants/constants.lisp and then regenerate the GlobalConstants.java

lu-pinto:
I'm closing this PR to fix a few things to adhere to branch name and commit message standards after sync with @powerslider. Will reopen a new one.

lu-pinto:
.

PR #806, opened 2024-06-27
Perf/issue 804/exception allocations

lu-pinto:
During one of our tests, which was around a contract running a loop, we observed around 3M of these objects created in a single transaction.

Since `Exceptions` is implemented by compounding several booleans I realized that we can implement this using a bitmask that only takes a size of a `short` instead of multiple instances that are re-cloned every time with all the memory and cpu overhead of creating such structures. 

@ahamlat I didn't include your suggestions for the parenthesis in the logic as the format is now kept pretty much unchanged. Let me know if you still want me to do this.

Note: If you are reviewing this one, you might also want to approve https://github.com/Consensys/linea-constraints/pull/244 which adds a constant required here.


powerslider (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/constants/GlobalConstants.java
```diff
@@ -25,18 +25,19 @@
  */
 public class GlobalConstants {
   public static final int BLOCKHASH_MAX_HISTORY = 0x100;
+  public static final int CREATE2_SHIFT = 0xff;
```
ahamlat, line 28: @powerslider  I'm wondering if these should be part of this PR. Does it mean that constants.lisp was changed without generating GlobalConstants.java ?

ahamlat (approved):
LGTM.
I have just one question for @powerslider regarding GlobalConstants.java

arithmetization/src/main/java/net/consensys/linea/zktracer/module/constants/GlobalConstants.java
powerslider: @ahamlat It means that it was not recently regenerated, so that is fine.

PR #814, opened 2024-07-01
feat: modify linea_getBlockTracesCountersV2 to return line counts for all modules (including ref tables and precompiles)

powerslider:


OlivierBBB (approved):
LGTM

PR #819, opened 2024-07-04
Missing limits for ECPAIRING

### LINKED ISSUE #817: ECPAIRING has 1 missing limit
# Main point

We have two missing limits related to the ECPAIRING precompile:
- `PRECOMPILE_ECPAIRING_G2_MEMBERSHIP_TESTS`
  - count the number of times we have to invoke the G2 membership testing circuit in gnark
  - will be set to 64
- ~`PRECOMPILE_ECPARING_LINES`~
  - ~counts the raw number of lines dedicated to ECPAIRING stuff~
  - ~**proposal:** set it to e.g. the power of 2 following 16 * ( 4 * 192 + 2) i.e. 16384~

## `PRECOMPILE_ECPAIRING_G2_MEMBERSHIP_TESTS`

Recall that we must test of G2 membership whenever in the following cases:
- If INTERNAL_CHECKS_PASSED = 1 and SUCCESS_BIT = 0:
  - 1 time per ECPAIRING call with (it is only required for the first $B_k\notin G_2$)
- If INTERNAL_CHECKS_PASSED = 1 and SUCCESS_BIT = 1:
  - 0, 1, 2, ... times (once per pair $(A_k, B_k)$ with $A_k = \infty$ and $B_k \neq \infty$.)

## `PRECOMPILE_ECPARING_LINES`

~Is necessary because we currently don't 'price' ECPAIRING calls that fail INTERNAL_CHECKS 😅 so the sequencer would accept an ECPAIRING call that fails ICP but has arbitrary size (given enough gas of course ... but it could be enormous.)~

**Correction.** This limit is actually already taken care of (the ECDATA module has a limit of `262144` rows.) Also note that the following computation

How large can the ECDATA trace get
- ECRECOVER: 128 * 10 = 1280
- ECADD: 16384 * 12 = 196608
- ECMUL: 32 * 10 = 320
- ECPAIRING (unexceptional, nontrivial): 16 * (4 * 12 + 2) = 800
- sum total = 199008

which accounts for "full use of the GNARK capabilities" still leaves plenty of room (wrt the `262144` limit) for plenty of horsing around. IRL we will likely only use a few hundred rows in that module in 99.9% of cases.

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/EcPairingEffectiveCall.java
```diff
@@ -165,4 +223,106 @@ public List<ColumnHeader> columnsHeaders() {
...
+  }
+
+  private boolean callDataContainsMalformedLargePoint(MessageFrame frame, MemorySpan callData) {
+    long offset = callData.offset();
+    long nPairsOfPoints = callData.length() / ECPAIRING_NB_BYTES_PER_MILLER_LOOP;
+
+    for (long i = 0; i < nPairsOfPoints; i++) {
+      final Bytes largeCoordinates = frame.shadowReadMemory(offset + ECPAIRING_NB_BYTES_PER_SMALL_POINT, ECPAIRING_NB_BYTES_PER_LARGE_POINT);
+
+      if (!isOnG2(largeCoordinates)) {
```
OlivierBBB: And on the curve !!!

OlivierBBB: Actually ok: the subgroup membership tests happens in the `isOnG2` method.

PR #834, opened 2024-07-15
fix(rlprcpt): index overflows if many topics

### LINKED ISSUE #833: RLP_RCPT.INDEX overflows in .lt
shadow node - block 6569423

letypequividelespoubelles:


DavePearce (approved):
Looks Good to Me. 

I checked that the new test passes with the new `zkevm.bin`, and fails with the old `zkevm.bin`.

PR #838, opened 2024-07-17
Feat: Optimise Trace Generation (except hub)

### LINKED ISSUE #837: Optimise Trace Generation (Except Hub)
The `Trace.java` file produced by `corset` for all of the modules is very inefficient places.  Some examples:

* Every `i8` requires `2` bytes per element (i.e. `16` bits)
* Every `i32` requires `8` bytes per element (i.e. `64` bits)
* Every `i64` requires `32` bytes per element (i.e. `256` bits)

For more details, see [Corset issue#160](https://github.com/Consensys/corset/issues/160).

At this stage, we don't have to do all the modules in one go.  In particular, I will not update the hub with this issue, but will leave that for future.

DavePearce:
This regenerates `Trace.java` files for all modules except the `hub` using the latest version of `corset`.  This optimises the generated `lt` trace files by ensuring the right widths are used for each column.  For example, a column with type `i8` previously used `2` bytes per element in the `lt` file (but now uses only `1`).  Likewise, a column with type `i64` was previously using `32` bytes per element in the `lt` file (but now uses `8` as expected).  

This has led to some modest space savings in the generated traces.  For the `531` tests in this repository, the total size of all `lt` files generates goes from `6.7G` to `5.1G`.  Furthermore, it seems likely that the `hub` will offer further (perhaps significant) improvements given that it accounts for the largest number of columns.

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Trace.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Trace.java
@@ -15,7 +15,6 @@
 
 package net.consensys.linea.zktracer.module.add;
 
-import java.math.BigInteger;
 import java.nio.MappedByteBuffer;
 import java.util.BitSet;
 import java.util.List;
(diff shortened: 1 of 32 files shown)


letypequividelespoubelles (approved):
LGTM, TY

powerslider (approved):

PR #858, opened 2024-07-24
fix: make --plugin-linea-conflated-trace-generation-traces-output-path option required to avoid faulty registration of the trace generation RPC endpoint

powerslider:
…h option required to avoid faulty registration of the trace generation RPC endpoint


fab-10:
Maybe we can move this configuration in the private section, since now also the sequencer needs to set it to something in order to start.

see https://github.com/Consensys/linea-arithmetization/issues/861

PR #864, opened 2024-07-25
feat: improve design of shared and private CLI options

powerslider:


arithmetization/src/main/java/net/consensys/linea/plugins/AbstractLineaOptionsPlugin.java
```diff
@@ -15,60 +15,63 @@
...
-import net.consensys.linea.plugins.config.LineaTracerConfiguration;
 import org.hyperledger.besu.plugin.BesuContext;
+import org.hyperledger.besu.plugin.BesuPlugin;
+import org.hyperledger.besu.plugin.services.PicoCLIOptions;
 
 /**
  * In this class we put CLI options that are private to plugins in this repo.
  *
  * <p>For the moment is just a placeholder since there are no private options
  */
```
fab-10: javadoc needs a refresh

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/capture/CaptureEndpointServicePlugin.java
```diff
@@ -63,6 +65,11 @@ private void createAndRegister(
         method.getNamespace(), method.getName(), method::execute);
   }
 
+  @Override
+  public Map<String, LineaOptionsPluginConfiguration> getLineaPluginConfigMap() {
+    return Map.of();
```
fab-10, line 70: Maybe here doesn't make a difference, but in general should call `var superMap = super.getLineaPluginConfigMap()` and add local mapping to the `superMap`

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/tracegeneration/TracesEndpointServicePlugin.java
```diff
@@ -38,10 +39,17 @@
...
 @AutoService(BesuPlugin.class)
 @Slf4j
-public class TracesEndpointServicePlugin extends AbstractLineaPrivateOptionsPlugin {
+public class TracesEndpointServicePlugin extends AbstractLineaOptionsPlugin {
   private BesuContext besuContext;
   private RpcEndpointService rpcEndpointService;
 
+  @Override
+  public Map<String, LineaOptionsPluginConfiguration> getLineaPluginConfigMap() {
+    final TracesEndpointCliOptions tracesEndpointCliOptions = TracesEndpointCliOptions.create();
```
fab-10, line 48: same as above

fab-10 (commented):
looks good, just a javadoc update and a note about calling super

fab-10 (approved):

PR #880, opened 2024-07-26
set besuVersion=24.7-develop-f812936

### LINKED ISSUE #879: Update Linea-Besu to 24.7-develop-f812936

fab-10:
fixes #879


fab-10:
@daniellehrner could you help with this:
```
/home/runner/work/linea-arithmetization/linea-arithmetization/arithmetization/src/test/java/net/consensys/linea/zktracer/testing/ToyWorld.java:40: error: ToyWorld is not abstract and does not override abstract method setAuthorizedCodeService(AuthorizedCodeService) in WorldUpdater

public class ToyWorld implements WorldUpdater {
```

@powerslider is there any specific reason this class is redefined? could be possible to directly use the one from Besu?

powerslider:
@fab-10 the reason why it's named `ToyWorld` is because we started our testing framework implementation based on those classes from Besu. Eventually though, we evolved it quite a bit, so it is not an apples to apples comparison. We are again working on it to improve it even more and merge the reference-tests logic with the unit testing logic, so I would not touch it for now.

PR #893, opened 2024-07-28
fix(blockCapturer): non empty block becomes empty while in some conflation

### LINKED ISSUE #877: fix: Blocks with Zero (Traceable) Transactions
A number of replay tests fail conflation due to blocks with (effectively) zero transactions.  _That is, blocks which only contain invalid transactions that are never seen by `ZkTracer`._  An example is the replace test `6110035-6110064`.  This fails with the following exception:

```
Exceptions triggered while tracing:
  - Index -1 out of bounds for length 0


java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:385)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at net.consensys.linea.zktracer.container.stacked.list.StackedList.getLast(StackedList.java:63)
```

This exception is actually thrown on block `6110052` (the 18th block in the given set being conflated).  However, that block contains two invalid transactions:

```
12:21:05.294 [Test worker] DEBUG org.hyperledger.besu.ethereum.mainnet.MainnetTransactionProcessor -- Invalid transaction: transaction up-front cost 0x95694f8b30a exceeds transaction sender account balance 0x0
12:21:05.299 [Test worker] DEBUG org.hyperledger.besu.ethereum.mainnet.MainnetTransactionProcessor -- Invalid transaction: transaction up-front cost 0x95694f8b30a exceeds transaction sender account balance 0x0
```

These transactions are never seen by `ZkTracer` and, hence, the list of transactions for that block remains empty.  This crashes on the last line here (in `ToyExecutionEnvironment.java`):

```java
      tracer.traceStartBlock(header, body);

      for (Transaction tx : body.getTransactions()) {
        System.out.println("******* PROCESSING TX " + tx.getHash());
        transactionProcessor.processTransaction(
            overridenToyWorld.updater(),
            (ProcessableBlockHeader) header,
            tx,
            header.getCoinbase(),
            tracer,
            blockId -> {
              throw new RuntimeException("Block hash lookup not yet supported");
            },
            false,
            Wei.ZERO);
      }

      tracer.traceEndBlock(header, body);
```

This makes sense because `traceEndBlock()` assumes at least one transaction has been processed.  Specifically, module `TxnData` assumes at least one transaction here:

```java
 @Override
  public void traceEndBlock(final BlockHeader blockHeader, final BlockBody blockBody) {
    this.currentBlock().currentTx().setCallWcpLastTxOfBlock(this.currentBlock().getBlockGasLimit());
  }
``` 

Its not clear to me what the right solution is.  But if we adjust this method as follows, then the original exception is gone and we move onto running `corset check`:
```java
  @Override
  public void traceEndBlock(final BlockHeader blockHeader, final BlockBody blockBody) {
    if(this.currentBlock().getTxs().size() > 0) {
      this.currentBlock().currentTx().setCallWcpLastTxOfBlock(this.currentBlock().getBlockGasLimit());
    }
  }
```

@powerslider FYI

letypequividelespoubelles:


arithmetization/src/test/java/net/consensys/linea/zktracer/ReplayTests.java
```diff
@@ -104,4 +104,24 @@ void rlprcptManyTopicsWoLogData() {
   void multipleFailingCallToEcrecover() {
     replay("5000544.json.gz");
   }
+
+  @Test
+  void empty_6110052_makesBatchFailing___old() {
```
powerslider, line 109: Please don't use underscores in the naming of the tests. Switch to camelCase.

PR #924, opened 2024-08-12
Only add needed dependencies in the uber jar

### LINKED ISSUE #923: Remove the need for the uber jar
At the moment this plugin is release as a single uber jar, that packs all the dependencies inside, including most of the Besu's dependencies, this create friction and is error prone since it is easy to have different versions of the same class that results in errors at runtime. 
For this we want to remove the uber jar, and just create a standard library jar that only contains the classes of the plugin. 
Note that the final distribution of the plugin, could also contain additional dependencies that are not already provided by Besu, for example at the moment `jackson-dataformat-yaml` jar needs to be installed along with the plugin jar in the Besu `plugins` dir to work.

fab-10:
Optimize the produced jar to only contains dependencies that are not already provided by Besu.

When adding a new dependency, as first thing verify it is already provided by Besu BOM, if yes then no need to declare it in the `dependency-management.gradle` file, use it in the required project.

If a dependency that is not already provided by Besu is required, then declare it in the `dependency-management.gradle` file, and mostly important use `pluginImplementation` configuration to use it in the project, this way it will be also included in the final jar.

This approach avoid duplicating dependencies classes in the final jar, and so reduces the jar size, and reduces the risk of classes clashes and errors at runtime.

fixes #923


gradle/dependency-management.gradle
```diff
@@ -31,6 +31,7 @@ repositories {
     content { includeGroupByRegex('com\\.splunk\\..*') }
   }
   mavenCentral()
+  mavenLocal()
```
powerslider, line 30: I  suppose we should remove this before merging?

gradle.properties
```diff
@@ -1,4 +1,4 @@
-releaseVersion=0.3.0-rc2
+releaseVersion=0.3.0-rc2-local
```
powerslider: This should also be reverted back.

powerslider (approved):

DavePearce (approved):
LGTM

Issue #935, opened 2024-08-14
Add a mechanism to catch needed block hash in BlockCapturer


letypequividelespoubelles:

DavePearce:
@letypequividelespoubelles Do we have any block numbers we know invoke the `BLOCKHASH` bytecode?

DavePearce:
Ok, some blocks which use the `BLOCKHASH` bytecode I have found:

* Block [8718330](https://lineascan.build/block/8718330)
* Block [8718090](https://lineascan.build/block/8718090) 

They seem to be few and far between.  They were found in scan of all blocks from `8718000` through to `8718375`.

Issue #764, opened 2024-06-06
verify replayTest works as intended


letypequividelespoubelles:
originally posted [here](https://github.com/Consensys/linea-constraints/issues/223) :

>Does it capture everything ? Do we make snapshots of accounts for:
    aborted CALL's / CREATE's ?
    failure condition CREATE's ?
    exceptional BALANCE / EXTCODEXXX / ... ?
    other edge cases ?
    SSTORE / SLOAD

- [ ] @OlivierBBB go through the list of opcodes to see if the above captures everything.

OlivierBBB:
**Question:** do we capture the BLOCKHASH values too ?

letypequividelespoubelles:
> **Question:** do we capture the BLOCKHASH values too ?

No : #935

PR #936, opened 2024-08-14
fix(blockCapturer): missing handling of selfdestruct

### LINKED ISSUE #933: BlockCapturer misses `SelfDestruct`
Method `tracePreExecution(MessageFrame frame)` in `net.consensys.linea.blockcapture.BlockCapturer` does a case analysis on bytecode instructions which can access storage.  However, it is missing a case for `SELFDESTRUCT` which transfers any outstanding balance to a given target account upon destruction.

@letypequividelespoubelles FYI.

letypequividelespoubelles:


OlivierBBB (approved):

powerslider (approved):

OlivierBBB:
Good catch. BTW didn't we have an issue with all the opcodes that would need to be caught by the capture mechanism?

letypequividelespoubelles:
you mean this #764 ? Yes but selfdestruct was forgotte, too.
And I create #935 for missing block hash

PR #748, opened 2024-05-31
feat(hub): the long and winding road to 1 + 1 = 2

letypequividelespoubelles:
This class contains peripheral information of the transaction required for the tracing.

PR #938, opened 2024-08-14
feat(corset): enable `go-corset` in `gradle` action

### LINKED ISSUE #908: feat: Enable `go-corset` in Github Actions
The goal here is to actually enable `go-corset` to be used in the CI pipeline.  At this time, there remains an outstanding resource usage limitation which is preventing it being enabled.  Once this is resolved, we can use a supplementary action this initially:

```
   go-corset-tests:
    runs-on: ubuntu-latest
    steps:
      - uses: webfactory/ssh-agent@v0.7.0
        with:
          ssh-private-key: |
            ${{ secrets.CONSTRAINTS_SSH_KEY }}
      - name: Checkout repository
        uses: actions/checkout@v3
        with:
          submodules: recursive

      - name: Set up JDK 21
        uses: actions/setup-java@v3
        with:
          java-version: 21
          distribution: temurin

      - name: Install Go
        uses: actions/setup-go@v4

      - name: Install Go Corset
        run:
          go install github.com/consensys/go-corset/cmd/go-corset@latest

      - name: Install Rust
        uses: actions-rust-lang/setup-rust-toolchain@v1

      - name: Install Rust Corset
        run: RUSTFLAGS=-Awarnings cargo install --git ssh://git@github.com/ConsenSys/corset --tag v9.7.13 --locked --force

      - name: Run unit tests
        run: ./gradlew -i :arithmetization:test
        env:
          CORSET_FLAGS: fields,expand,expand,expand
```

DavePearce:
This enables `go-corset` in the `gradle` action workflow.  However, `go-corset` is only run as part of a job which runs separately from the main tests job.  As such, it will not prevent pull requests from being merged.  The goal here is really to enable testing of `go-corset` before (eventually) swapping out the existing Rust corset tool.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -105,3 +105,40 @@ jobs:
           java-version: 21
       - name: spotless
         run: ./gradlew --no-daemon --parallel clean spotlessCheck
+
+  go-corset-tests:
+    runs-on: ubuntu-latest-128
+    steps:
+      - uses: webfactory/ssh-agent@v0.7.0
+        with:
+          ssh-private-key: |
+            ${{ secrets.CONSTRAINTS_SSH_KEY }}
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v3
+        with:
+          java-version: 21
+          distribution: temurin
+
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - name: Install Corset
+        run: RUSTFLAGS=-Awarnings cargo install --git ssh://git@github.com/ConsenSys/corset --tag v9.7.13 --locked --force
+
+      - name: Install Go
+        uses: actions/setup-go@v4
+
+      - name: Install Go Corset
+        run:
+          go install github.com/consensys/go-corset/cmd/go-corset@latest
+
+      - name: Run unit tests
+        run: ./gradlew :arithmetization:test
+        env:
+          JAVA_OPTS: -Dorg.gradle.daemon=false
+          CORSET_FLAGS: disable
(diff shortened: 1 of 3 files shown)


.github/workflows/gradle.yml
```diff
@@ -105,3 +105,41 @@ jobs:
           java-version: 21
       - name: spotless
         run: ./gradlew --no-daemon --parallel clean spotlessCheck
+
+  go-corset-tests:
+    runs-on: ubuntu-latest
```
letypequividelespoubelles: maybe should use `ubuntu-latest-128` ? See with @FlorianHuc

letypequividelespoubelles (approved):
Maybe we need to use an other machine as they don't run anymore on `ubuntu-latest, see with @FlorianHuc

.github/workflows/gradle.yml
DavePearce: So, the currently `gradle.yml` file only uses `ubuntu-latest` (see [here](https://github.com/Consensys/linea-tracer/blob/arith-dev/.github/workflows/gradle.yml)).  I don't think Tsvetan has got it setup yet so that we can use e.g. `ubuntu-latest-16core`.

DavePearce: Ok, I see that on the long running branch it has been changed to `ubuntu-latest-128`.

letypequividelespoubelles: yes, by @FlorianHuc  : https://github.com/Consensys/linea-tracer/pull/748/files#diff-0d634959c8276ae976e39ba475e63d0b9ec27824470c79f1971f58fe37c46325

DavePearce: Ok, yeah will check with Florian now.

DavePearce: So, this has resolved itself with the merge from the long running branch.  I am now using `ubunutu-latest-128`.

PR #946, opened 2024-08-17
fix: swap ordering of r/s fields in tx snapshot

DavePearce:
The ordering of these two fields was incorrect, and should be swapped. This then means the TX hashes are computed correctly.  Since this happens on the capture side, changes will only be noticed once the fix is deployed and new blocks captured.  Existing captures contain the `r` and `s` fields in the wrong order, and need to be updated.


letypequividelespoubelles (approved):
LGTM

PR #947, opened 2024-08-17
fix: Use correct CHAINID in EVM

### LINKED ISSUE #940: Incorrect `CHAINID` used in `ToyExecutionEnvironment`
_(this issue arises from investigations in #877)_

Whilst studying various replay tests, I have observed an unexpected difference from mainnet.  Specifically, the `CHAINID` is set incorrectly for execution in `ToyExecutionEnvironment` (and possibly other places perhaps).  This is despite considerable effort to set the `CHAINID` correctly.

Whilst observing the execution trace for the first transaction in block `6110045` I noticed this:

```
.. CHAINID  => 0x0000000000000000000000000000000000000000000000000000000000000539 
```

Here, `0x539` corresponds to the (default) chainID of 1337.  This is actually a BESU default.  The problem arises with the following line:

```
private final EVM evm = MainnetEVMs.london(EvmConfiguration.DEFAULT);
```

The problem here is that the `CHAINID` for the EVM (which is what will be reported by the `CHAINID` bytecode) is set by default, unless [explicitly passed in the constructor](https://github.com/hyperledger/besu/blob/main/evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java).

Suggested fix is:

1. Remove `evm` field altogether.
2. Instead, create an EVM within `getMainnetTransactionProcessor()` as follows:

```
private MainnetTransactionProcessor getMainnetTransactionProcessor() {
    EVM evm = MainnetEVMs.london(this.chainId, EvmConfiguration.DEFAULT);
    ...
``` 

This then correctly sets the `CHAINID` as reported by the corresponding bytecode.  I can't really judge whether this is a significant issue or not.

@letypequividelespoubelles FYI

DavePearce:
This updates the `ToyExecutionEnvironment` to use the correct CHAINID when constructing the EVM.  Previously, this was not done correctly and meant that the `CHAINID` bytecode was returning an invalid value, which in turn mean that storage addresses were being computed incorrectly (i.e. because they incorporated the CHAINID into the hash).

### DIFF
--- a/testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
+++ b/testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
@@ -84,7 +84,7 @@ public class ToyExecutionEnvironment {
       Hash.fromHexStringLenient("0xdeadbeef123123666dead666dead666");
 
   private final ToyWorld toyWorld;
-  private final EVM evm = MainnetEVMs.london(EvmConfiguration.DEFAULT);
+
   @Builder.Default private BigInteger chainId = CHAIN_ID;
   @Singular private final List<Transaction> transactions;
 
@@ -217,7 +217,9 @@ public class ToyExecutionEnvironment {
   }
 
   private MainnetTransactionProcessor getMainnetTransactionProcessor() {
-
+    // Construct EVM for executing transactions.
+    final EVM evm = MainnetEVMs.london(this.chainId, EvmConfiguration.DEFAULT);
+    //
     PrecompileContractRegistry precompileContractRegistry = new PrecompileContractRegistry();
 
     MainnetPrecompiledContracts.populateForIstanbul(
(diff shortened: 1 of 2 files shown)


DavePearce:
@powerslider I am not so certain about changes to the `ToyExecutionEnvironmentV2`.  As far as I can tell, this doesn't read the `chainID` out of the `tx` like the original `ToyExecutionEnvironment`.  So, it seems to me that it will always have the default value of `1337` for the `chainID` which will not be correct for some of the GeneralStateTests.

zkevm-constraints
letypequividelespoubelles, line 1: why updating the constraint submodule ?

DavePearce: Oh whoops, I will fix that (later today).

DavePearce: Fixed now, and tests all passing.

Issue #949, opened 2024-08-19
Unexpected Side Effect from `ZkTracer.tracePreExecution`


DavePearce:
(see also #877 

Currently, we have this observation:

* [Block 6110045](https://lineascan.build/tx/0xc8574d2e389a6474f06e635870bb1a3b2c5fe73f515297733545f305861d4428#internal), second tx (`0xc8574d...`).  In the last `STATICCALL` gas goes from `238643` beforehand to `3572` within (where should be `234,816` within).

However, commenting out this code from `ZkTracer.tracePreExecution()` yields the expected gas usage:

```
if (frame.getCode().getSize() > 0) {
      try {
        this.hub.tracePreOpcode(frame);
        this.pin55.ifPresent(x -> x.tracePreOpcode(frame));
      } catch (final Exception e) {
        this.tracingExceptions.add(e);
      }
    }
```

Therefore, something somewhere is having an unexpected side effect on the EVM state.

DavePearce:
Looks like `this.hub.tracePreOpcode(frame);` specifically is the problem.

DavePearce:
In `Hub.java` this is where its coming from:

```
void processStateExec(MessageFrame frame) {
    ...
    this.handleStack(frame);
    this.triggerModules(frame);  <------
```

DavePearce:
And then here:

```
void triggerModules(MessageFrame frame) {
   ...
    if (this.pch.signals().oob()) {
      this.oob.tracePreOpcode(frame);  <----
    }
```

DavePearce:
Now here:

```
 public void tracePreOpcode(MessageFrame frame) { // This will be renamed to tracePreOp
    this.chunks.add(new OobOperation(frame, add, mod, wcp, hub, false, 0, 0));
    OpCode opCode = OpCode.of(frame.getCurrentOperation().getOpcode());

    if (opCode.isCall()) {
      Address target = Words.toAddress(frame.getStackItem(1));

     ...
        } else {
          // Other precompiles case
          this.chunks.add(new OobOperation(frame, add, mod, wcp, hub, true, 0, 0)); <---
        }
```

DavePearce:
And now narrowed to somewhere in here:

```
private void populateColumns(final MessageFrame frame) { 
   ...
   } else if (isPrecompile()) {
      ... <---
   }
}
```

DavePearce:
Ok, it seems to be this assignment which is causing the problem:

```
      final BigInteger callGas =
[...]

PR #950, opened 2024-08-19
fix: change implementation of gasAvailableForChildCall due to having side effects in TangerineWhistleGasCalculator

powerslider:


OlivierBBB:
Fixes the issue https://github.com/Consensys/linea-tracer/issues/949 by applying the same fix as https://github.com/Consensys/linea-tracer/commit/dc3a4fba869bc044e73b69118ff996acc51609df.

OlivierBBB (approved):

Issue #840, opened 2024-07-18
`ToyWorld` issues


OlivierBBB:
List of known issues with our tests. They seem to be related to the `ToyWorld` but we don't know exactly what's going wrong.
- [ ] coinbase address not recognized as such, in particular when it comes to paying the COINBASE address in gas fees at the end of the transaction it's not done for some reason, see #767
- [x] failures when regenerating `Trace.java` files (I believe this was settled by @DavePearce and @letypequividelespoubelles ?) see see https://github.com/Consensys/linea-arithmetization/issues/794#issuecomment-2232440935
- [x] deployment transactions (i.e. `tx.toAddress` is empty) don't run the provided initialization code https://github.com/Consensys/linea-tracer/blob/8fb4af6fa7713dea98ec351cd14626a181acd798/arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/ContractModifyingStorageTest.java#L180
- [ ] for deployments again there are more layers of `WorldUpdater`'s (4 initially, now 3 for some reason) than expected
- [x] #967
- [x] make the `ToyWorld` aware of precompiles (solved by @OlivierBBB in Tbilissi)
- [x] #997
  - [ ] re-enable `HUB` constraints (uncomment them in the constraints repo's `Makefile`
  - [ ] recompile the zkevm.bin object (`make zkevm.bin -B`)
  - [ ] run some of the simpler tests in the repo (add test, 1+1=2, ...)

letypequividelespoubelles:
@OlivierBBB @gauravahuja I think it's now all ok with @gauravahuja 's ToyV2 ?

PR #966, opened 2024-08-22
fix: `ToyWorld.commit()` parent update

### LINKED ISSUE #964: Incorrect implementation of `ToyWorld.commit()`
By comparing execution using `ToyWorld` vs `org.hyperledger.besu.evm.fluent.SimpleWorld`, we can see a problem with `ToyWorld.commit()`.  Currently, it is:

```
    if (parent != null) {
      parent.addressAccountMap.putAll(addressAccountMap);
    }
```
However, with this implementation, [Block 6110040](https://lineascan.build/tx/0xe3e410ddcd6286a9e67ea75f82c654c81c49ec5d2696b7059885af7bf5b2316c#internal) tx `0xe3e410ddcd6286a9e` incorrectly reverts.  However, using the implementation from `SimpleWorld` and it succeeds:

```
    addressAccountMap.forEach(
      (address, account) -> {
        if (!account.updateParent()) {
          parent.addressAccountMap.put(address, account);
        }
      });
```

**NOTE:** I would personally suggest using `SimpleWorld` instead of `ToyWorld` for the replay tests at least.  Ideally, we could drop `ToyWorld` altogether --- though I am unsure whether that is possible at this stage.  The reason being simply less code to maintain.

@letypequividelespoubelles @powerslider FYI --- next one identified.  Will make the PR now.

DavePearce:
This puts through a simple fix for `ToyWorld.commit()` and also tidies up `ToyWorld` a little by removing the `accounts` field which is not needed (as was incorrectly managed).

### DIFF
--- a/testing/src/main/java/net/consensys/linea/testing/ToyAccount.java
+++ b/testing/src/main/java/net/consensys/linea/testing/ToyAccount.java
@@ -176,6 +176,23 @@ public class ToyAccount implements MutableAccount {
     mutable = false;
   }
 
+  /**
+   * Push changes into the parent account, if one exists
+   *
+   * @return true if a parent account was updated, false if not (this indicates the account should
+   *     be inserted into the parent contact).
+   */
+  public boolean updateParent() {
+    if (parent instanceof ToyAccount toyAccount) {
+      toyAccount.balance = balance;
+      toyAccount.nonce = nonce;
+      toyAccount.storage.putAll(storage);
+      return true;
+    } else {
+      return false;
+    }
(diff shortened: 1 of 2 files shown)


OlivierBBB:
@DavePearce Does this fix this long standing issue wrt `accounts` already noted in #840 (and which I just turned into an issue #967) ?

DavePearce:
@OlivierBBB Yes, in a sense it does because it removes the `accounts` list altogether!

powerslider (approved):

PR #973, opened 2024-08-23
chore: switch to ubuntu-22.04-16core github runner on the tests stage

powerslider:


OlivierBBB (approved):
Not an expert but LGTM 🙂

PR #982, opened 2024-08-26
feat: update blockcapturer to capture outcomes

DavePearce:
Update `BlockCapturer` to capture outcomes.  These are then checked (when present) as part of a `replayTest()`.  The benefit of this is that it helps to ensure the testing environment reflects mainnet.


testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
```diff
@@ -211,7 +223,7 @@ private void execute() {
               Wei.ZERO);
 
       this.testValidator.accept(result);
-      this.zkTracerValidator.accept(tracer);
+      // this.zkTracerValidator.accept(tracer);
```
FlorianHuc: if it's not used, should we delete it?

arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
```diff
@@ -39,34 +43,79 @@
...
+    final List<StorageSnapshot> storage = txStorage.collapse(world);
+    // Construct the result snapshot
+    TransactionResultSnapshot resultSnapshot =
+        new TransactionResultSnapshot(
+            status, output.toHexString(), logStrings, gasUsed, accounts, storage, destructStrings);
+    // Record the outcome
+    txSnapshot.setTransactionResult(resultSnapshot);
+    // Reset for next transaction
+    txIndex++;
+    this.txStorage = new StorageReaper();
```
FlorianHuc: Given we initialize them in the enterTransaction, I we shouldn't need to initialize them here. Should we instead add a close method here, and in the enter check that the reapers are either null or close? This way we would be sure that tx and entered then exited one after the other?

DavePearce: Originally I did something like that, but I had a problem with serialisation.  Let me revisit that.

testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
DavePearce: That's a mistake.  It should not have been commented out.

arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
DavePearce: Ok, have resolved the problem here.  It was that, for some transactions, `traceStartTransaction()` is never called!  Instead, `tracePrepareTransaction()` is always called but we can go straight to `traceEndTransaction()` from there without calling `traceStartTransaction()`.

I've also refactored the way in which the `TransactionChecker` was integrated.  Previously, it was a wrapper around an existing `OperationTracer` (which was `ZkTracer`).  However, now its a standalone implementation and I use a separate mechanism for multiplexing tracers together (defined in `ConflationAwareOperationTracer`).  This has an additional benefit that I can relatively easily include the `BlockCapturer` for testing purposes.

testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
DavePearce: That's a mistake.  It should not have been commented out.

however, in fact, the enclosing `execute()` method is no longer used anywhere as all the existing tests now use `ToyExecutionEnvironmentV2`.  So we could actually drop that, and I would suggest doing so (perhaps in a separate PR though).

arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
```diff
@@ -68,22 +68,27 @@ public void traceStartBlock(BlockHeader blockHeader, BlockBody blockBody) {
     this.reaper.enterBlock(blockHeader, blockBody);
   }
 
+  @Override
+  public void tracePrepareTransaction(WorldView worldView, Transaction transaction) {
+    this.reaper.prepareTransaction(transaction);
+  }
+
   @Override
   public void traceStartTransaction(WorldView worldView, Transaction transaction) {
```
letypequividelespoubelles: we should plug only on `tracePrepareTransaction` besu hook. `traceStartTransaction` is just after, when the upfront gascost (tx + dtat + access list + deployment ) is paid by the sender.

letypequividelespoubelles (changes requested):

arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
DavePearce: We need to use `tracePrepareTransaction()` here.  For some transactions, `traceStartTransaction()` is never called and execution goes straight to `traceEndTransaction()`.  I'm not sure exactly the circumstances, but its either for an invalid transaction or perhaps one that has no code associated with it.  Eitherway, without this, we will get a `NullPointerException` in `traceEndTransaction()` as `txStorage` and `txAddresses` will be `null`.

DavePearce: I only found this out from testing, tbh.

DavePearce: Anyway, sorry, I guess the point is: I stuck with `traceStartTransation()` because that is what the `BlockCapturer` was originally doing.  However, yes, am happy to just merge into `tracePrepareTransaction()`.

letypequividelespoubelles: yes, we should be plugged only at `tracePrepareTransaction()` and remove `traceStartTransaction()`

DavePearce: Yeah sorry for some reason I read you original message the other way around.  Have made the change and just waiting for tests to go green.

letypequividelespoubelles (approved):
LGTM, TY

PR #992, opened 2024-08-27
Make `ReturnSection`'s `resolveAtContextReEntry` method use the createe frame

### LINKED ISSUE #991: Fix for `resolveAtContextReEntry` in `ReturnSection`
# Main point

There is a bug with `resolveAtContextReEntry` for (deployment) `ReturnSection`'s. The code was using the current frame at that point in time (that is: the _creator_ frame which resumes execution after RETURN'ing from the _createe_ frame) to trigger ROM_LEX.

OlivierBBB:


letypequividelespoubelles (approved):
LGTM, reviewed together.

PR #1001, opened 2024-08-27
Feat: bulk capture and bulk replay

gusiri:
This PR  adds `bulk capture` script and `bulk replay` feature.

----

**bulk capture**: bulk downloads multiple replay files specified in range.csv

Usage: 
```
0. copy `bulk_capture.sh` and `range.csv` to your `src/test/resources` folder. (optional, but this will make your life easier) 
1. add some ranges to range.csv
2. run `./bulk_capture.sh range.csv range2.csv range3.csv` 
3. check `./replays` folder

```

-----

**bulk replay**: `replayBulk` run a bulk replay of multiple replay files specified by a directory. 

- `*.json.gz` in the target directory will be replayed all at once. 
-  The conflated traces will be moved to `conflated` directory once replayed.
-  The replay files will be moved to `replayed` directory once completed.
-  Note: `CORSET_VALIDATOR.validate()` is disabled by default for replayBulk

Usage:
```
1. modify the target directory for replayBulk() in ReplayTests.java
2. check ./conflated folder
```
----
**replayAt**: replay with any file path. The conflated trace will be written to the same directory as the replay file instead of the system tmp folder.


OlivierBBB:
Amazing @gusiri 🙂 I started writing a bash script for a similar purpose (and with help from chatgpt) because I was getting tired of capturing a single range all the time 😄

OlivierBBB:
Could you provide an example `range.csv` file ?

scripts/range.csv
```diff
@@ -0,0 +1,3 @@
+7566651-7566655
+7566656-7566663
```
gusiri, line 2: @OlivierBBB looks something like this.

OlivierBBB: So no commas but dashes. Got it 👍

gusiri: a sample range.csv is included in `/scripts`

scripts/bulk_capture.sh
```diff
@@ -0,0 +1,124 @@
...
+EOF
+    )
+
+    # Append the command to full_cmd
+    full_cmd+="curl -X POST 'http://localhost:8545' --data '$payload' | jq '.result.capture' -r | jq . | gzip > $remote_filename; echo 'Captured $start-$end'; sleep 3; "
+done < "$range_file"
+
+# Set the compressed file name based on the very first and very last ranges
+compressed_file="/tmp/replays_${very_first}_${very_last}.tar.gz"
+local_compressed_file="replays_${very_first}_${very_last}.tar.gz"
```
OlivierBBB, line 98: Those seem like slightly odd names for a collection of conflated ranges, maybe it would be a better convention to call it something like
```bash
compressed_file="/tmp/replays_${number_of_ranges}_conflations_between_${very_first}_and_${very_last}.tar.gz"
local_compressed_file="replays_${number_of_ranges}_conflations_between_${very_first}_and_${very_last}.tar.gz"
```
Where `number_of_ranges` would be the number of rows in the `.csv` file. Or something similar.

scripts/bulk_capture.sh
```diff
@@ -0,0 +1,124 @@
...
+    # Debug: Show extracted start and end values
+    echo "Debug: Extracted start: '$start', end: '$end'"
+
+    # Check if start and end are properly extracted
+    if [ -z "$start" ] || [ -z "$end" ]; then
+        echo "Debug: Skipping invalid line: '$line'"
+        continue
+    fi
+
+    # Track the very first and very last ranges
```
OlivierBBB: Given the names below wouldn't it make sense to track the smallest `$start` and the largest `$end` (if both are well-formed and in the correct order, of course) ?

scripts/bulk_capture.sh
OlivierBBB: Those seem like slightly odd names for a collection of conflated ranges, maybe it would be a better convention to call it something like

```bash
compressed_file="/tmp/replays_${number_of_ranges}_conflations_between_${very_first}_and_${very_last}.tar.gz"
local_compressed_file="replays_${number_of_ranges}_conflations_between_${very_first}_and_${very_last}.tar.gz"
```

Where `number_of_ranges` would be the number of rows in the `.csv` file. Or something similar.

OlivierBBB (approved):

scripts/bulk_capture.sh
gusiri: Agree. This way the file name will be more  meaningful. Will fix this.

scripts/bulk_capture.sh
gusiri: I got your point. I originally used these ranges so avoid duplicate names, but now we had that uuid.  I think we can add more context to the compressed file name. Let me fix this.

scripts/bulk_capture.sh
gusiri: updated

scripts/bulk_capture.sh
gusiri: Updated

gusiri:
@OlivierBBB 

Now we have this compressed file names as you suggested. Also ${very_first} and ${very_last} represent the smallest $start and the largest $end, respectively.

```
compressed_file="/tmp/replays_${number_of_ranges}_conflations_between_${very_first}_and_${very_last}.tar.gz"
local_compressed_file="replays_${number_of_ranges}_conflations_between_${very_first}_and_${very_last}.tar.gz"
```


Added more logs as well.
```
Replay files for 4 conflations between 7566651 and 7766656 have been processed and downloaded.
Skipped ranges due to invalid configurations: 3.
```

PR #1007, opened 2024-08-28
add tests for exceptions.

FlorianHuc:


lorenzogentile404 (changes requested):
Could you add a comment describing the tests cases? Especially, what is the meaning of these numbers?

```
    Mockito.when(frame.stackSize()).thenReturn(7);
    Mockito.when(frame.getStackItem(2)).thenReturn(Bytes.ofUnsignedShort(1));
```
Also, if I understand correctly, what `Mockito` does it actually continues the execution of the test until the `Assertion` whenever all the `when` instructions return the `thenReturn` values, otherwise the execution is "filtered out"? Some comment about it too, without explaining the framework, but just the reason why we use it would be useful.

FlorianHuc:
Mockito creates an object, and every 'when' define the behaviour for a specific call. Other calls with unspecified behviour will return null.

lorenzogentile404 (approved):

Issue #1014, opened 2024-08-28
Fix: MXP / ROMLEX problem with "deployment number 0"


DavePearce:
_(to recreate this requires the fix for #998 to have been applied --- hopefully this will be merged shortly)_

There is some kind of problem affecting the `fatMxp()` and `testMxpRandomAdvanced()` tests which only appears after the fix for #998.  I have marked these tests as `@Disabled` temporarily until this issue is resolved.  The trace is:

```
Exceptions triggered while tracing:
  - RomChunk with address 0x6ff1019c622e4641f86f4bb7232b7901b8d20db6, deployment number 0 and deploymentStatus true not found


java.lang.RuntimeException: RomChunk with address 0x6ff1019c622e4641f86f4bb7232b7901b8d20db6, deployment number 0 and deploymentStatus true not found
	at net.consensys.linea.zktracer.module.romlex.RomLex.getCodeFragmentIndexByMetadata(RomLex.java:99)
	at net.consensys.linea.zktracer.module.hub.fragment.account.AccountFragment.resolvePostConflation(AccountFragment.java:197)
	at net.consensys.linea.zktracer.module.hub.defer.DeferRegistry.resolvePostConflation(DeferRegistry.java:142)
	at net.consensys.linea.zktracer.module.hub.Hub.traceEndConflation(Hub.java:467)
```

@letypequividelespoubelles @lorenzogentile404 FYI

letypequividelespoubelles:
test `testRandomMxpInstructionsFollowedByTriggeringRoob()` to reenable too

OlivierBBB:
@DavePearce You should try my latest PR --- I've solved similar issues there, mayb this is sufficient ?

OlivierBBB:
I am cautiously optimistic that this bug was fixed in the latest commit to #1027

Issue #1008, opened 2024-08-28
fix: update existing replays to ensure correct TX hash


DavePearce:
The goal here is to redownload the replays for existing tests (e.g. `fatMxp`) to ensure they generate the correct `tx` hash.  That just helps with debugging.

DavePearce:
Some have been done as part of #1015.  Also it might make sense to wait until the new `BlockCapturer` is online which provides tx outcome data.

DavePearce:
Fixed as part of #1030

PR #1015, opened 2024-08-28
fix: `ToyAccount.isStorageEmpty()`

DavePearce:
This puts through a straightforward fix for the method `ToyAccount.isStorageEmpty()`.  This unfortunately exposes a problem with MXP / ROMLEX (#1014).  Two tests have been marked as `@Disabled` until this problem is resolved.

Furthermore, some replay files have been updated to ensure they have the right tx hash.  More to be done later as part of #1008.


letypequividelespoubelles (approved):
LGTM

PR #1021, opened 2024-08-28
feat: update gradle test config to show better output

### LINKED ISSUE #1020: Improve output for `go-corset` in Github Action
The output we're getting from `go-corset` is not super helpful.  There are two aspects:

1. We're not seeing any `INFO` or `DEBUG` log information.  Whilst this matches what we get from Rust `corset`, I think we can do better.
2. We're not getting any logs being uploaded which can then be inspected.

DavePearce:
This updates the gradle test configuration to generate better output during testing.  In particular, so we can see why a particular test failed rather than just "FAILED".  This will be helpful for both Rust `corset` and `go-corset`.

### DIFF
--- a/gradle/tests.gradle
+++ b/gradle/tests.gradle
@@ -101,7 +101,7 @@ test {
     showStandardStreams true
 
     // set to false to hide passed standard out and error streams
-    showPassedStandardStreams true
+    showPassedStandardStreams false
 
     // set to false to hide skipped standard out and error streams
     showSkippedStandardStreams true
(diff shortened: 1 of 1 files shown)


DavePearce:
I wonder if makes sense to flip this to `false` as well:

```java
    // set to false to hide passed standard out and error streams
    showPassedStandardStreams true
```

Since we only care about the output when the test fails?

letypequividelespoubelles (approved):
LGTM

letypequividelespoubelles:
> I wonder if makes sense to flip this to `false` as well:
> 
> ```java
>     // set to false to hide passed standard out and error streams
>     showPassedStandardStreams true
> ```
> 
> Since we only care about the output when the test fails?

I do agree, it would be easier to read if we print only failing test

DavePearce:
Ok, will put that in place and see how it looks.

PR #1025, opened 2024-08-28
Fix for `BLOCKHASH` conversion error

### LINKED ISSUE #1024: Prevent Conversion Error for `BLOCKHASH` in `BlockCapturer`
Currently we have this in `BlockCapturer.java`:

```
case BLOCKHASH -> {
        if (frame.stackSize() > 0) {
          BigInteger peek0 = frame.getStackItem(0).toUnsignedBigInteger();
          // Determine block number requested
          final long blockNumber = peek0.longValueExact();
          // Use enclosing frame to determine hash
          Hash blockHash = frame.getBlockHashLookup().apply(blockNumber);
          // Record it was seen
          this.reaper.touchBlockHash(blockNumber, blockHash);
        }
      }
```

However, there is a risk that the conversion `longValueExact()` could fail in the event someone uses a large (invalid) block number as the argument.  That is permitted and, in this case, the EVM places are `0` on the stack.  Therefore, we should catch the case where there is a conversion error and ignore it (i.e. no need to touch the block hash in this case).

Thanks to @letypequividelespoubelles for spotting this problem!

DavePearce:
This puts in place a simple fix for ensuring captured block hashes are for valid blocks relative to the current block.  This is based on the implementation in [BlockHashOperation.java](https://github.com/Consensys/linea-besu/blob/zkbesu/evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java)

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
@@ -15,7 +15,6 @@
 
 package net.consensys.linea.blockcapture;
 
-import java.math.BigInteger;
 import java.util.List;
 import java.util.Set;
 
@@ -39,6 +38,10 @@ import org.hyperledger.besu.plugin.data.BlockBody;
 import org.hyperledger.besu.plugin.data.BlockHeader;
 
 public class BlockCapturer implements ConflationAwareOperationTracer {
+  public static final int MAX_RELATIVE_BLOCK = 256;
+
+  private static final int MAX_BLOCK_ARG_SIZE = 8;
+
   /**
    * The {@link Reaper} will collect all the data that will need to be mimicked to replay the block.
    */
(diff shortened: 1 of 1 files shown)


arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
```diff
@@ -39,6 +38,10 @@
 import org.hyperledger.besu.plugin.data.BlockHeader;
 
 public class BlockCapturer implements ConflationAwareOperationTracer {
+  public static final int MAX_RELATIVE_BLOCK = 256;
```
letypequividelespoubelles, line 41: you could use `BLOCKHASH_MAX_HISTORY` in `GlobalConstants.java`

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
DavePearce: Well maybe at some point I'll put that in

PR #1030, opened 2024-08-29
fix: expose true output from tx result failure

DavePearce:
This updates the `TransactionResultChecker` to work around the fact that the `MainnetTransactionProcessor` calls `traceEndTransaction()` twice when a `RuntimeException` is raised.  Previously, this was causing the true reason for a result failure.  Now, we can see the true reason.

Furthermore, this updates all the replay files to include tx outcomes, and blockhashes (where appropriate).


letypequividelespoubelles (approved):
LGTM

OlivierBBB:
This PR fixes the double dipping of the `traceEndTx` IIUC ?

PR #1034, opened 2024-08-29
perf: more efficient data storage structure

### LINKED ISSUE #743: perf: compute only once the number of line

letypequividelespoubelles:
Modified `StackedSet` and `StackedList` (which is the structure that stores all module infos, except for the `HUB` ?):
- instead of using a List<List<ModuleOperation>> (one list per tx) that enable to pop as many transaction we want, we use two list : one list for all the tx since the begining of the conflation, one list for the current tx, that we can pop
- enhance the lineCounting to not loop over all the tx of the batch, and just add to the counter each time we add an operation (and keep the possibility to pop the last tx)

Modified the lineCounting of the MMIO to not loop over each MMIO op each time we call `lineCount()`


arithmetization/src/main/java/net/consensys/linea/zktracer/module/ecdata/EcData.java
```diff
@@ -22,7 +22,7 @@
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import net.consensys.linea.zktracer.ColumnHeader;
-import net.consensys.linea.zktracer.container.stacked.list.StackedList;
+import net.consensys.linea.zktracer.container.stacked.StackedList;
```
lorenzogentile404, line 27: Here I guess you just changed folder structure?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/Exp.java
```diff
@@ -21,7 +21,7 @@
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import net.consensys.linea.zktracer.ColumnHeader;
-import net.consensys.linea.zktracer.container.stacked.set.StackedSet;
+import net.consensys.linea.zktracer.container.stacked.StackedSet;
```
lorenzogentile404, line 27: why can't we move to a `StackedList` here but in other modules we can?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/gas/Gas.java
```diff
@@ -67,9 +67,11 @@ private GasParameters extractGasParameters(MessageFrame frame) {
   @Override
   public void commit(List<MappedByteBuffer> buffers) {
     final Trace trace = new Trace(buffers);
-    for (int i = 0; i < this.chunks.size(); i++) {
-      GasOperation gasOperation = this.chunks.get(i);
-      gasOperation.trace(i + 1, trace);
+    int stamp = 0;
+    for (GasOperation gasOperation : this.chunks.getAll()) {
+      // TODO: I thought we don't have stamp for gas anymore ?
```
lorenzogentile404, line 74: You are right. The PR updating GAS has to still be merged.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/Exp.java
letypequividelespoubelles: StackedSet <=> we want to kill duplicate and it's not ordered (as in MOD, ADD, WCP, etc ...)
StackedList <=> ordered list of unique operation (RLP_TXN, MMU, etc)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/ecdata/EcData.java
letypequividelespoubelles: yes, and rewrite completely the StackedList / StackedSet ;)

lu-pinto:
I have a question about concurrency, not something introduced in this PR though. Correct me if I'm wrong but all of these modules act like singletons to be more performant but they don't look compatible with parallel transactions since there's only one tracer, correct? If we ever want to run parallel transactions here you would have to rethink the design or run multiple tracers.
context: parallel transaction were introduced to besu not so long ago.

letypequividelespoubelles:
> I have a question about concurrency, not something introduced in this PR though. Correct me if I'm wrong but all of these modules act like singletons to be more performant but they don't look compatible with parallel transactions since there's only one tracer, correct? If we ever want to run parallel transactions here you would have to rethink the design or run multiple tracers. context: parallel transaction were introduced to besu not so long ago.

Exactly, we have only one ZkTracer running at the time, and we're not able to parallelise. This PR doesn't change anything about this though.

ahamlat:
> context: parallel transaction were introduced to besu not so long ago.

Transaction parallelization is introduced on Besu Block processing (MainnetParallelBlockProcessor.java class). During block building, we still evaluate transactions sequentially, so there is no concurrency issue to worry about.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
```diff
@@ -67,19 +67,19 @@ public List<ColumnHeader> columnsHeaders() {
   public void commit(List<MappedByteBuffer> buffers) {
     final Trace trace = new Trace(buffers);
     int stamp = 0;
-    for (AddOperation op : this.chunks) {
+    for (AddOperation op : chunks.getAll()) {
```
ahamlat: Why this is removed here, knowing that it is used above. I think you can keep it across the whole class.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
```diff
@@ -67,19 +67,19 @@ public List<ColumnHeader> columnsHeaders() {
...
 
   @Override
   public int lineCount() {
-    return this.chunks.lineCount();
+    return chunks.lineCount();
   }
 
   public BigInteger callADD(Bytes32 arg1, Bytes32 arg2) {
-    this.chunks.add(new AddOperation(OpCode.ADD, arg1, arg2));
+    chunks.add(new AddOperation(OpCode.ADD, arg1, arg2));
```
ahamlat: The same as above.

ahamlat (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,111 @@
...
+  public int size() {
+    return thisTransactionOperation.size() + operationSinceBeginningOfTheConflation.size();
+  }
+
+  public int lineCount() {
+    return lineCounter.lineCount();
+  }
+
+  public E get(int index) {
+    return getAll().get(index);
```
ahamlat: I think there is a performance issue here. We should reconsider the way we get an element from this data structure

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
letypequividelespoubelles: we are with @OlivierBBB in war with `this.` and whenever we see it, delete it ;)

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: true, but we almost never use this (we only use in some of @lorenzogentile404 's test)

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
OlivierBBB, line 1: Is this just a copy and paste of `.../stacked/set/StackedSet.java` ? Same question applies to `.../stacked/list/StackedList.java`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
OlivierBBB: True ☠️

OlivierBBB: @ahamlat is there any benefit to keeping `this.bla` rather than writing `bla` directly ? When I look at the `.class` files it automatically inserts the omitted `this.`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
```diff
@@ -55,30 +55,30 @@ public String moduleKey() {
 
   @Override
   public void traceStartBlock(final ProcessableBlockHeader processableBlockHeader) {
-    this.batchUnderConstruction = true;
-    this.wcp.additionalRows.push(this.wcp.additionalRows.pop() + TIMESTAMP_BYTESIZE);
+    batchUnderConstruction = true;
+    wcp.additionalRows.add(TIMESTAMP_BYTESIZE);
```
OlivierBBB, line 53: Is `bla.add(x)` just an alias for `bla.push( bla.pop() + x )` ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/CountOnlyOperation.java
```diff
@@ -45,4 +44,9 @@ public void add(final int operationCount) {
   public int lineCount() {
```
OlivierBBB, line 46: Could we add a `Preconditions.CheckArgument(thisTransactionCount + operationCount >= 0)` to `add` ? Given that we sometimes `add` negative values.

arithmetization/src/test/java/net/consensys/linea/zktracer/containers/StackedListTests.java
```diff
@@ -52,22 +49,6 @@ protected int computeLineCount() {
...
-    assertThat(state.lineCount()).isEqualTo(1);
-
-    state.enter();
-    state.add(new IntegerModuleOperation(3));
-    assertThat(state.lineCount()).isEqualTo(4);
-
-    state.pop();
-    assertThat(state.lineCount()).isEqualTo(1);
-  }
-
```
OlivierBBB: What was the purpose of that test @letypequividelespoubelles ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmu/MmuOperation.java
```diff
@@ -88,12 +86,8 @@ protected int computeLineCount() {
...
-    if (mmioLineCount != -1) {
-      return mmioLineCount;
-    }
-
-    mmioLineCount = 0;
+  public int mmioLineCount() {
+    int mmioLineCount = 0;
     if (traceMe()) {
       for (int i = 0; i < mmuData().numberMmioInstructions(); i++) {
         mmioLineCount +=
```
OlivierBBB: In the function body that follows we use the method `numberOfRowOfMmioInstruction`. It would be great to rename it using the word `Line` instead of `Row` e.g. `lineCountOfMmioInstruction`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
OlivierBBB: Also see other comment about non-negativity check for `add`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/Mxp.java
```diff
@@ -71,12 +58,13 @@ public List<ColumnHeader> columnsHeaders() {
   @Override
   public void commit(List<MappedByteBuffer> buffers) {
     final Trace trace = new Trace(buffers);
-    for (int i = 0; i < this.mxpOperations.size(); i++) {
-      this.mxpOperations.get(i).trace(i + 1, trace);
+    int stamp = 0;
+    for (MxpOperation mxpOperation : mxpOperations.getAll()) {
+      mxpOperation.trace(++stamp, trace);
     }
```
OlivierBBB, line 54: What's the idea with the `blaOperatoins.getAll()` that appears everywhere ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/trm/Trm.java
```diff
@@ -100,9 +100,8 @@ public void commit(List<MappedByteBuffer> buffers) {
     final Trace trace = new Trace(buffers);
 
     int stamp = 0;
-    for (TrmOperation operation : this.operations) {
-      stamp++;
-      operation.trace(trace, stamp);
+    for (TrmOperation operation : operations.getAll()) {
+      operation.trace(trace, ++stamp);
     }
```
OlivierBBB, line 99: What is the reason for the `++stamp` rather than `stamp++`. IIRC it's something about updating a value prior / post something else ...

OlivierBBB: Nevermind, ChatGPT helped me with this 🙂

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxn/RlpTxn.java
```diff
@@ -1214,9 +1214,8 @@ public List<ColumnHeader> columnsHeaders() {
   public void commit(List<MappedByteBuffer> buffers) {
     final Trace trace = new Trace(buffers);
     int absTxNum = 0;
-    for (RlpTxnChunk chunk : this.chunkList) {
-      absTxNum += 1;
-      traceChunk(chunk, absTxNum, trace);
+    for (RlpTxnChunk chunk : operations.getAll()) {
+      traceChunk(chunk, ++absTxNum, trace);
```
OlivierBBB: We should get rid of the `chunk` terminology while we're at it (at least for modules other than ROM related ones) in favour of `operation` / `operations` / `traceOperations` / ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
```diff
@@ -55,31 +58,37 @@ public String moduleKey() {
...
+  }
 
   @Override
-  public int lineCount() {
-    int sum = 0;
-    for (MmuOperation o : mmu.mmuOperations()) {
-      sum += o.computeMmioLineCount();
+  public void traceEndTx(TransactionProcessingMetadata tx) {
+    for (MmuOperation o : mmu.mmuOperations().thisTransactionOperation()) {
+      lineCounter.add(o.mmioLineCount());
```
OlivierBBB, line 71: `o` 😅

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
```diff
@@ -55,31 +58,37 @@ public String moduleKey() {
...
-  public void enterTransaction() {}
+  public void enterTransaction() {
+    lineCounter.enter();
+  }
 
   @Override
-  public void popTransaction() {}
+  public void popTransaction() {
+    lineCounter.pop();
+  }
```
OlivierBBB, line 66: There seems to be a lot of repetition wrt the `enterTransaction` and `popTransaction` methods. Maybe we could have a default implementation that they inherit ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
OlivierBBB: I guess judging by the line count that went down quite a bit the answer is no 😅

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Implements a system of nested lists behaving as a single one, where the current context
+ * modification can transparently be dropped.
+ *
+ * @param <E> the type of elements stored in the list
+ */
+@Accessors(fluent = true)
+public class StackedList<E extends ModuleOperation> {
+  private final List<E> operationSinceBeginningOfTheConflation = new ArrayList<>();
+  @Getter private final List<E> thisTransactionOperation = new ArrayList<>();
```
OlivierBBB: Maybe rename to
```bash
operationsOfCurrentTransaction
operationsGeneratedByCurrentTransaction
```
But in all cases the name should start with `operations...`

**Personal preference:** `operationsGeneratedByCurrentTransaction`

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Implements a system of nested lists behaving as a single one, where the current context
+ * modification can transparently be dropped.
+ *
+ * @param <E> the type of elements stored in the list
+ */
+@Accessors(fluent = true)
+public class StackedList<E extends ModuleOperation> {
+  private final List<E> operationSinceBeginningOfTheConflation = new ArrayList<>();
```
OlivierBBB: Maybe one of
```bash
operationsPriorToCurrentTransaction
operationsGeneratedByPriorTransactions
operationsGeneratedByPriorTransactionsInConflation
```

**Personal preference:** `operationsGeneratedByPriorTransactions`

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
OlivierBBB: But in all cases the name should start with `operations...`

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,111 @@
...
+  @Getter private final List<E> thisTransactionOperation = new ArrayList<>();
+  private final CountOnlyOperation lineCounter = new CountOnlyOperation();
```
OlivierBBB, line 41:
```bash
operationsOfExternalCircuits
operationsLedgerForExternalCircuits
```

**Personal preference:** `operationsLedgerForExternalCircuits`

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
OlivierBBB: This will also apply to https://github.com/Consensys/linea-tracer/blob/d10cd5ea5dca08e25ccec4477c0ed2cc4ae8b398/arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/CountOnlyOperation.java#L21

OlivierBBB: Personal preference: `operationsGeneratedByCurrentTransaction`

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,111 @@
...
+
+  public List<E> getAll() {
+    final List<E> all = new ArrayList<>(operationSinceBeginningOfTheConflation);
+    all.addAll(thisTransactionOperation);
+    return all;
+  }
+
+  public boolean isEmpty() {
+    return size() == 0;
+  }
```
OlivierBBB, line 86: For what purposes is this being used ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,111 @@
...
+
+  public boolean contains(Object o) {
```
OlivierBBB, line 108: Why does this use `Object o` rather than say `E extends ModuleOperation` or so like in the next method ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,111 @@
...
+    return thisTransactionOperation.contains(o)
+        || operationSinceBeginningOfTheConflation.contains(o);
+  }
+
+  public boolean add(E e) {
+    lineCounter.add(e.lineCount());
+    return thisTransactionOperation.add(e);
+  }
```
OlivierBBB, line 115: I guess this `.add` operation in the function body is one of `List<E>` ? What is the definition of the returned `boolean` ?

OlivierBBB (changes requested):
Please address my comments 🙂

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
letypequividelespoubelles: yes, I just kept the same name

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: no, because the underlying has changesd. Before it was a stack of int, now it's a `CountOnlyOperation`

letypequividelespoubelles: > Also see other comment about non-negativity check for `add`

We want to add a neg at a time. True it'd be better to have a remove indeed. I'll do it.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
letypequividelespoubelles: ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: I don't follow you this. Which external circuits are you talking about ? This is really the lineCount of the StackedSet / StackedList. 

Rhe purpose of this is to kill a loop on all previous operations when querrying lineCount

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: returns `true` id really added, `false` if nothing was added. So in our case, always `true` unless something unexpected happened

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: implemented a better way to get wo using getAll

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: checks added

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/CountOnlyOperation.java
letypequividelespoubelles: added

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/Mxp.java
letypequividelespoubelles: it gives the exact same result

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxn/RlpTxn.java
letypequividelespoubelles: done

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: I did the change, but I don't like it, the two differets name are too close, I don't distinguish them with fast reading

letypequividelespoubelles: need to find more different names for the prior and current

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: because theoretically you can ask if smtg which is not E extends bla is in ? It'll return false of course

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: in fact I reverted the change, it was impossible for me to read. I'm open to new names of course :)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
OlivierBBB: It's just a very short name 🙂

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
OlivierBBB: My understanding (maybe wrong ?) was that CountOnlyOperations was for stuff like precompiles etc ... which we delegate to external circuits. Is that not the case ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
OlivierBBB: Do we want that ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: Not only. We use it too in StackedSet and StackedList to avoid a loop when line counting. 
We use it too in WCP to count the "potential additional rows".

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmu/MmuOperation.java
letypequividelespoubelles: done

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
letypequividelespoubelles: done, I now introduce two specalised interface, one for StatelessModule (ADD, WCP, etc), and one for normal module (RLP_*, etc)

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: afaik it's the standard way of doing it

arithmetization/src/test/java/net/consensys/linea/zktracer/containers/StackedListTests.java
letypequividelespoubelles: a very usefull test deleted by mistake

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: I Finally removed the possibility to remove lineCounting. It's only usage was for blockData to remove 4 rows at the end of the block, and it coyuld lead to issue if the last tx of the block was adding less than 4 rows in WCP (which is is theory possible ...)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
ahamlat: It is more for clarity especially when you have a parameter of a method with the same name as the class field. In this, this. is mandatory, but then you will need to add it everywhere to make it consistant. We can skip it if we are sure to not have to deal with this case.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
```diff
@@ -0,0 +1,49 @@
...
+import net.consensys.linea.zktracer.container.stacked.StackedSet;
+import org.hyperledger.besu.evm.worldstate.WorldView;
+
+/**
+ * A {@link StatelessModule} is a {@link Module} that contains unordered {@link E} where we don't
+ * keep trace of duplicates.
+ */
+public interface StatelessModule<E extends ModuleOperation> extends Module {
+
+  StackedSet<E> operations();
```
ahamlat: What is the difference between StatelessModule and StatefullModule, as I see that both have operations as a field ? I would imagine that StatelessModule doesn't have any field.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
lorenzogentile404: If there is a local variable with the same name of a class attribute, it is not only for clarity but that is how to refer to one or the other. I personally like using always `this` for class attributes since, if at some point, you declare a local variable with the same name then you'll change the behaviour in multiple places without having any feedback from Java.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,134 @@
...
+  public int size() {
+    return thisTransactionOperation.size() + operationSinceBeginningOfTheConflation.size();
+  }
+
+  public int lineCount() {
+    return lineCounter.lineCount();
+  }
+
+  public E get(int index) {
+    if (index < operationSinceBeginningOfTheConflation.size()) {
```
ahamlat: This implementation looks very specific for me. Are we sure that there're no edge cases here ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
```diff
@@ -0,0 +1,134 @@
...
+  public void clear() {
+    operationSinceBeginningOfTheConflation.clear();
+    thisTransactionOperation.clear();
+    lineCounter.clear();
+  }
+
+  public void finishConflation() {
+    conflationFinished = true;
+    operationSinceBeginningOfTheConflation.addAll(thisTransactionOperation);
+    thisTransactionOperation.clear();
```
ahamlat: What about the lineCounter here ? shall we call enter on lineCounter to be sure we add the line counting of the last transaction (in the block) ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,134 @@
...
+  public void clear() {
+    operationSinceBeginningOfTheConflation.clear();
+    thisTransactionOperation.clear();
+    lineCounter.clear();
+  }
+
+  public void finishConflation() {
+    conflationFinished = true;
+    operationSinceBeginningOfTheConflation.addAll(thisTransactionOperation);
+    thisTransactionOperation.clear();
```
ahamlat: The same comment as for StackedSet.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/bin/Bin.java
```diff
@@ -77,7 +66,7 @@ public List<ColumnHeader> columnsHeaders() {
   }
 
   @Override
-  public int lineCount() {
-    return this.chunks.lineCount();
+  public StackedSet<BinOperation> operations() {
```
ahamlat: This is a getter, could be better to make it like this : getOperations(). The comment works for all other modules.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  default void enterTransaction() {
+    operations().enter();
+  }
+
+  @Override
+  default void popTransaction() {
+    operations().pop();
```
ahamlat: A stateless class shouldn't change the internal state.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxrcpt/RlpTxnRcpt.java
```diff
@@ -46,45 +48,36 @@
...
             txMetaData.getBesuTransaction().getType(),
             txMetaData.statusCode(),
             txMetaData.getAccumulatedGasUsedInBlock(),
             txMetaData.getLogs());
-    this.chunkList.add(chunk);
+    operations.add(operation);
   }
 
-  public void traceChunk(final RlpTxrcptChunk chunk, int absTxNum, int absLogNumMax, Trace trace) {
+  public void traceChunk(
```
ahamlat: I guess we should rename traceChunk methods with traceOperation, right ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
```diff
@@ -50,14 +52,17 @@
...
-    implements Module, PostOpcodeDefer, ImmediateContextEntryDefer, ContextExitDefer {
-  private static final RomChunkComparator ROM_CHUNK_COMPARATOR = new RomChunkComparator();
+    implements StatelessModule<RomChunk>,
+        PostOpcodeDefer,
+        ImmediateContextEntryDefer,
+        ContextExitDefer {
 
   private final Hub hub;
 
   @Getter private final StackedSet<RomChunk> chunks = new StackedSet<>();
```
ahamlat: Why the name is different here from other modules ? sometimes I see `chunks`, sometimes `operations`.
If the name of the attribute is `operations`, the getter should be : `getOperations`. If the name is `chunks`, the getter should be `getChunks()`

ahamlat (changes requested):
Could you please address my few comments / questions ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
letypequividelespoubelles: One is using a StackedSet, the other one a StackedList

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
letypequividelespoubelles: we could add it for clarity, but there is no need, as lineCount always return the sum; I can do it for clarity if you prefer.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
letypequividelespoubelles: no, it does, it deals with a StackSet, cf previous commet. We call them StateLess because they don't interact with Besu state, but computes operations (addition, etc ...)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxrcpt/RlpTxnRcpt.java
letypequividelespoubelles: yes

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
letypequividelespoubelles: forgotten renaming, will switch to operations as all others

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
ahamlat: Why do you call it StatelessModule in this case ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
ahamlat: There is no need in this case. May be just add a comment for the class, regarding enter and finishConflation because it is not clear without context (callers context).

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: we just want to retrieve the object at index i in a concatenation of two lists. I don't see any edge case, but I might be wrong

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
ahamlat: I think you need to review the naming in this case, stateless in a very specific word in Java that means that the class  doesn't change the internal state.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
letypequividelespoubelles: done

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: done too

arithmetization/src/main/java/net/consensys/linea/zktracer/module/bin/Bin.java
letypequividelespoubelles: done

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
letypequividelespoubelles: yes I don't like the name either, but in this name we refer to the "evm" state

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: done

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: done

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
letypequividelespoubelles: cf next comment. Renamed

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/StatelessModule.java
letypequividelespoubelles: done

ahamlat (approved):
LGTM. Once the refactoring is shipped, we can do a performance analysis and see if there is any issue related to performance

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
OlivierBBB: ok

OlivierBBB (approved):

PR #1027, opened 2024-08-28
fix: many things ...

OlivierBBB:
WIP

Issue #1006, opened 2024-08-28
range 6032696  6032699


FlorianHuc:
## Ranges

```rust
6032696-6032699
```

## Issue

time curl --location --request POST 'localhost:15480' --header 'Content-Type: application/json' --data-raw '{"jsonrpc": "2.0", "method": "linea_generateConflatedTracesToFileV2", "params":[{"startBlockNumber": 6032696 , "endBlockNumber":6032699 , "expectedTracesEngineVersion": "0.5.0-beta"}], "id": 1}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Plugin internal error","data":"Exceptions triggered while tracing:\n  - null\n"}}

Preconditions.checkArgument(
    hub.currentFrame().isDeployment() == deploymentStatus); // sanity check
deploymentStatus
deploymentStatus is true  but  hub.currentFrame().isDeployment() is false
this is in StopSection()

PR #1045, opened 2024-08-30
Use `ReferenceTestWorldState` instead of `ToyWorld`

### LINKED ISSUE #969: Replace `ToyWorld` with `SimpleWorld` for `ReplayTests`
The goal here is to use the `SimpleWorld` implementation which is part of BESU.  Since we have had some problems with `ToyWorld` this should hopefully resolve them.

I have prototyped this already, and the `ReplayTests` at least it is possible to do.  Potentially, for other tests it might also be possible.

DavePearce:
<strike>This puts in place a temporary fix which converts `ToyWorld` over into `ReferenceTestWorldState` to use when executing transactions.  The conversion could be done more elegantly, but the goal here is to get something out for people to use ASAP.  Eventually, we would then remove `ToyWorld` altogether.</strike>

**UPDATE**: to get the fix out there more quickly, it was cherry-picked into #1027.  What's left of this PR is relatively straightforward.  It just updates some of the replay files so they now all include tx result / blockhash information.  It also marks all tests which pass tx result checking.  Only a small number remain which fail for various reasons (e.g. problems with the balance or `nonce`).  Those issues will be resolved separately.


DavePearce:
Will wait for the PR for  #1006 to be merged before merging this, since that now contains a commit from this PR (which was cherry-picked).

letypequividelespoubelles (approved):

Issue #1054, opened 2024-08-30
Incorrect Nonce for New Contracts in Replays


DavePearce:
For replay tests, currently the initial contract nonce is set as `0`.  In fact, it should be `1` and this is causing some tests to fail.

DavePearce:
Fixed with #1056

PR #1056, opened 2024-08-30
fix: initial contract nonce (#1054)

DavePearce:
The initial contract nonce was incorrectly set as `0` when it should be `1`.  This also enables tx result checking for a small number of tests which now pass as a result of this fix.  The remaining replays fail because of an incorrect balance.

### DIFF
--- a/testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
+++ b/testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
@@ -254,7 +254,7 @@ public class ToyExecutionEnvironment {
         new MessageCallProcessor(evm, precompileContractRegistry);
 
     final ContractCreationProcessor contractCreationProcessor =
-        new ContractCreationProcessor(evm, false, List.of(), 0);
+        new ContractCreationProcessor(evm, false, List.of(), 1);
 
     return new MainnetTransactionProcessor(
         gasCalculator,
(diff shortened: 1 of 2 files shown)


testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
```diff
@@ -254,7 +254,7 @@ private MainnetTransactionProcessor getMainnetTransactionProcessor() {
         new MessageCallProcessor(evm, precompileContractRegistry);
 
     final ContractCreationProcessor contractCreationProcessor =
-        new ContractCreationProcessor(evm, false, List.of(), 0);
+        new ContractCreationProcessor(evm, false, List.of(), 1);
```
OlivierBBB, line 257: The only question I have with respect to that is: when we query an account that is about to being deployed will we get a zero nonce ?

OlivierBBB (approved):
LGTM

testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironment.java
OlivierBBB: We'll find out soon enough when connecting HUB constraints 😄

OlivierBBB, line 257: I implemented the same fix on my working branch for the `ToyExecutionEnvironmentV2` BTW. It's not solving our issues with deployments.

PR #1077, opened 2024-09-03
Fix/issue 1064/cn not found for call data copy

letypequividelespoubelles:


OlivierBBB:
But I'm not understanding why we want a context for empty call data ? The MMU/MMIO instructions don't require it, don't they ?

letypequividelespoubelles:
> But I'm not understanding why we want a context for empty call data ? The MMU/MMIO instructions don't require it, don't they ?

just to avoid java exceptions : 

If someons does a CALLDATALOAD for exemple, the source is the callFrame CN = h (with empty data I agree). But when the mmuCall search for the callFrame with CN = h, it doesn't find it. That's why we must add in any case an (empty) callFrame with this CN.

OlivierBBB:
But shouldn't we just consult the `currentFrame` and ask whether its `callDataSpan.isEmpty()` or so ?

OlivierBBB (commented):
Shouldn't we just consult the `currentFrame` and ask whether its `callDataSpan.isEmpty()` or so ? We have a `CallDataInfo` object in the `CallFrame` where we can do that kind of check.

letypequividelespoubelles:
I don't get your point, in the arithmetization, if we calldatacopy from a deployment root context, the source CN will be h, so why not creating an empty callFrame at CN = h ? 
Note : I am only talking about creating the empty callFrame only to store the empty callData of the ROOT context, when `shouldCopyTxData ==false`.

letypequividelespoubelles:
closed and new fix is https://github.com/Consensys/linea-tracer/pull/1081

PR #1081, opened 2024-09-04
fix(callDataCopy): source ID

### LINKED ISSUE #1064: 67050-67059
## Ranges

```rust
67050-67059
```

## Issue

Failed to fetch or parse JSON data for block range =67050-67059: 'result'
Data received before error for block range =67050-67059: {'jsonrpc': '2.0', 'id': 1, 'error': {'code': -32603, 'message': 'Plugin internal error', 'data': 'Exceptions triggered while tracing:\n  - call frame CN 56640 not found\n'}}

07:25:49.694 [vert.x-worker-thread-8] ERROR net.consensys.linea.zktracer.ZkTracer - First exception that was caught while tracing:
java.lang.ClassCastException: class org.apache.tuweni.bytes.ArrayWrappingBytes cannot be cast to class org.apache.tuweni.bytes.Bytes32 (org.apache.tuweni.bytes.ArrayWrappingBytes and org.apache.tuweni.bytes.Bytes32 are in unnamed module of loader 'app')
        at net.consensys.linea.zktracer.module.hub.section.halt.SelfdestructSection.<init>(SelfdestructSection.java:86) ~[?:?]
        at net.consensys.linea.zktracer.module.hub.Hub.traceOpcode(Hub.java:1035) ~[?:?]
        at net.consensys.linea.zktracer.module.hub.Hub.processStateExec(Hub.java:958) ~[?:?]
        at net.consensys.linea.zktracer.module.hub.Hub.tracePreExecution(Hub.java:731) ~[?:?]
        at net.consensys.linea.zktracer.ZkTracer.tracePreExecution(ZkTracer.java:258) ~[?:?]
        at org.hyperledger.besu.evm.EVM.runToHalt(EVM.java:207) ~[besu-evm-24.8-develop-915fcb01.jar:24.8-develop-915fcb01]
        at org.hyperledger.besu.evm.processor.AbstractMessageProcessor.codeExecute(AbstractMessageProcessor.java:175) ~[besu-evm-24.8-develop-915fcb01.jar:24.8-develop-915fcb01]

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -560,30 +560,27 @@ public void traceContextEnter(MessageFrame frame) {
...
           recipientAddress,
           new Bytecode(
-              recipientAddress == null
-                  ? transients.tx().getBesuTransaction().getData().orElse(Bytes.EMPTY)
-                  : Optional.ofNullable(frame.getWorldUpdater().get(recipientAddress))
-                      .map(AccountState::getCode)
-                      .orElse(Bytes.EMPTY)),
+              Optional.ofNullable(frame.getWorldUpdater().get(recipientAddress))
+                  .map(AccountState::getCode)
+                  .orElse(Bytes.EMPTY)),
```
OlivierBBB: Does this work for deployments ? I'm not convinced this will work for deployment transactions.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/mmu/MmuCall.java
```diff
@@ -212,16 +210,8 @@ public static MmuCall callDataCopy(final Hub hub) {
...
-  }
-
-  public static int callDataContextNumber(final Hub hub) {
-    final CallFrame currentFrame = hub.callStack().current();
-
-    return currentFrame.isRoot()
-        ? currentFrame.contextNumber() - 1
-        : hub.callStack().parent().contextNumber();
+        .referenceOffset(callDataInfo.memorySpan().offset())
+        .referenceSize(callDataInfo.memorySpan().length());
```
OlivierBBB, line 214: That's the right way of doing it.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
letypequividelespoubelles: we never entered in the other case, as recipientAddress is set as the effectiveTo address of the tx, so imho, yes.
(Actually, it was a simplification proposed by Intellij, I just verified it was ok)

OlivierBBB: How does the transaction's init code enter the picture i.e. `transients.tx().getBesuTransaction().getData()` ?

OlivierBBB: I think that the issue is with the initial condition. It should be replaced with `tx.to.isEmpty ? ... : ...`

OlivierBBB: Could this explain our ongoing issue with deployment transactions in the ToyWorld v1 ?

OlivierBBB: Hm, no, Besu itself didn't care to execute an instructions. So that's not the explanation.

letypequividelespoubelles: added the initCode in the callFrame in case of deployment

Issue #1047, opened 2024-08-30
Incorrect Balance on some Replays


DavePearce:
For example, for `rlprcptManyTopicsWoLogData()` we get this when result checking is enabled:

```
   java.lang.RuntimeException: tx 0xfc101cd3e88876de3176b428226014d06c88fefdd854c0624752947e5aec514f balance of account 0x0000000000000000000000000000000000000000 (776153018587787272) does not match expected value (775913456386696799)
      at net.consensys.linea.zktracer.ReplayTests.replay(ReplayTests.java:75)
```

DavePearce:
So, two blocks illustrating the problem:

* [50022](https://lineascan.build/block/50022)
* [5104826](https://lineascan.build/block/5104826)
* [5104842](https://lineascan.build/block/5104842)
* [5105647](https://lineascan.build/block/5105647)
* [5105680](https://lineascan.build/block/5105680)
* [6020028](https://lineascan.build/block/6020028)
* [6569423](https://lineascan.build/block/6569423)
* [8718468](https://lineascan.build/block/8718468)
* [8718673](https://lineascan.build/block/8718673) (tx [0x053f451](https://lineascan.build/tx/0x053f45108d5b645f4979074f3df942346d5f3431f27d165735f8054f499015c4))
* [8718853](https://lineascan.build/block/8718853) (tx [0xe17a78bf9](https://lineascan.build/tx/0xe17a78bf9a3255d222e9e1c97a44d7831cd63fc7648e31d5dd764e0152ef6472))

DavePearce:
Observations:

- Block [6020028](https://lineascan.build/block/6020028) has this at the end:
```
0x7b5780d6df85a7df96a3e1a019639a1dbde937db => 0x9899f25bdd0785b2b47c1bd29711d88406c3b54d[0x508ca82df566dcd1b0de8296e70a96332cd644ec] (gas: 357730)
0x508ca82df566dcd1b0de8296e70a96332cd644ec => 0x0000000000000000000000000000000000000000[0x0000000000000000000000000000000000000000] (gas: 295991)

tx 0x90d947da9f51698a99cc2e178dfb7e53bde0c9a92c40ff27492c524fd4335750 balance of account 0x8f81e2e3f8b46467523463835f965ffe476e1c9e (6718674392457320274) does not match expected value (6718801508324469274)
```

- [6569423](https://lineascan.build/block/6569423) has this:
```
[...]

PR #1091, opened 2024-09-04
feat: update replay tests with tx outcomes

DavePearce:
This updates all replay files with outcome information.  This also enables result checking on a reasonable number of tests, but there are actually a lot which fail still for various reasons.  Furthermore, this improves debugging information for `go-corset`.


arithmetization/src/test/java/net/consensys/linea/zktracer/ReplayTests.java
OlivierBBB, line 1: I take it the tests were you re-enabled the check are those that succeeded the "final state check" ?

OlivierBBB (approved):
LGTM but I'm not qualified to review the `gradle` business.

arithmetization/src/test/java/net/consensys/linea/zktracer/ReplayTests.java
DavePearce: Yes, they pass the tx result checking (which includes checks against the state).  However, what you should notice is that still quite a few fail.  I have been compiling the evidence I have here: #1047

PR #1096, opened 2024-09-04
chore: add validation and usage for bulk capture

gusiri:
This PR
- Updated usage instructions in the script header.
- Added validation to skip lines that don't match the `integer-integer` format in input CSV files.
- Introduced the option to keep or remove compressed replay files after decompression via the `KEEP_LOCAL_COMPRESSED_FILE` constant. Keep the compressed replay files by default. 
- Echo absolute path of the replays folder
- Echo absolute path of local compressed file


OlivierBBB (approved):
LGTM

PR #1107, opened 2024-09-05
fix(oob): inst modexp pircing f of max

### LINKED ISSUE #1106: fix(oob): inst modexp pricing f of max

lorenzogentile404:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/OobOperation.java
```diff
@@ -1074,16 +1074,14 @@ private void setModexpPricing(ModexpPricingOobCall prcModexpPricingOobCall) {
         callToISZERO(1, BigInteger.ZERO, prcModexpPricingOobCall.getExponentLog());
 
     // row i + 2
-    final BigInteger fOfMax =
+    final BigInteger squareRootFOfMax =
```
OlivierBBB: We shouldn't call it that, maybe `ceilingOfMaxDividedBy8` or so ... But certainly not `squareRoot`

OlivierBBB (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/OobOperation.java
lorenzogentile404: Let's wait for the tests to finish. If they pass, then I'll fix the name and we force the merge

OlivierBBB (approved):

Issue #1049, opened 2024-08-30
Reference tests support tool


lorenzogentile404:
First of all, let's summarise the involved in the tool:

- https://github.com/Consensys/linea-tracer/tree/arith-dev/reference-tests: reference tests we want to run.
- https://github.com/Consensys/linea-constraints: constraints that needs to be satisfied by the traces generated by each references tests 
-  https://github.com/Consensys/corset: software verifying if the traces satisfy the above constraints.

The following is a pseudocode describing the tool:

```python
# here it is assumed these lists are already populated
list reference_tests # extracted from linea-tracer repo
list modules # extracted from linea-constraints repo
list constraints # extracted from linea-constraints repo

map module_to_failing_tests = {}
map constraint_to_failing_test = {}

# init maps
for module in modules:
    module_to_failing_tests.put(module, [])

for constraint in constraints:
    constraint_to_failing_tests.put(constraint, [])

for reference_test in reference_tests:
    list failing_constraints = reference_test.run()
    for failing_constraint in failing_constraints:
        module_to_failing_tests.get(failing_constraint.module()).add(reference_test)
        constraint_to_failing_tests.get(failing_constraint).add(reference_test) 

# later we want to have the chance to execute again tests based on these maps
module_to_failing_tests("hub").runAll()
constraint_to_failing_test("hub.name-annoying-constraint").runAll()
```

Informally, what we want is:
1) For each module, show the list of tests that make fail at least one of its constraints and the length of this list;
2) For each constraint, show the list of tests in which it fails and the length of this list;
3) Allow to selectively run only the tests associated to a failing module (case 1) or a failing constraint (case 2);

OlivierBBB:
I like the idea.

Unimportant question but what is gained by having a `map_of_failing_modules` when you already have a `map_of_failing_constraints` ? I would imagine that the constraints are of the form `module.name-of-constraint` already, no ?

Also shouldn't the present repo https://github.com/Consensys/linea-tracer be mentioned, too ?

I would also be interested in a `map_of_tests_to_java_exceptions` and a `map_of_tests_to_corset_exceptions`.

lorenzogentile404:
@OlivierBBB indeed `constraint_to_failing_test` is enough due to naming of constraints following `module.name-of-constraint`. The `module_to_failing_tests` is just a convenient way to aggregate all constraints of a module.

OlivierBBB:
[...]

PR #1108, opened 2024-09-05
1049 - Reference test tool

bradbown:
Implements the first PR for issue #1049 
Updates templates to extend the tests with a `TestWatcher`, the test watcher intercepts failed tests and records to an output file a mapping of the failed test to its module and constraint pair.

### Example
```json
[
    {
        "moduleName": "blockdata",
        "constraints": {
            "value-constraints": [
                "callcodecallcodecall_110_d0g0v0_London[London]",
                "callcodecallcode_11_d0g0v0_London[London]",
                "callcodecallcode_11_SuicideEnd_d0g0v0_London[London]",
                "callcodecallcode_11_OOGE_d0g0v0_London[London]",
                "exp_d0g0v0_London[London]"
            ]
        }
    },
    {
        "moduleName": "txndata",
        "constraints": {
            "txndata-into-blockdata": [
                "callcodecallcodecall_110_d0g0v0_London[London]",
                "callcodecallcode_11_d0g0v0_London[London]",
                "callcodecallcode_11_SuicideEnd_d0g0v0_London[London]",
                "callcodecallcode_11_OOGE_d0g0v0_London[London]",
                "exp_d0g0v0_London[London]"
            ]
        }
    }
]
```


reference-tests/src/test/resources/templates/BlockchainReferenceTest.java.template
lorenzogentile404, line 1: These tests classes may be generated automatically. Could you check it to be sure that the annotation for the `TestWatcher` is permanent and does not disappear whenever we re-generate them on a new machine?

reference-tests/src/main/java/net/consensys/linea/MapFailedReferenceTestsTool.java
```diff
@@ -0,0 +1,170 @@
...
+
+  public static List<ModuleToConstraints> getModuleToConstraints(
+      String jsonString, JsonConverter jsonConverter) {
+    List<ModuleToConstraints> constraintToFailingTests = new ArrayList<>();
+    if (!jsonString.isEmpty()) {
+      constraintToFailingTests =
+          new ArrayList<>(
+              Arrays.asList(jsonConverter.fromJson(jsonString, ModuleToConstraints[].class)));
+    }
+    return constraintToFailingTests;
```
lorenzogentile404: the name of the method seems inconsistent with the returned value. Wouldn't it be cleared `getConstraintsToFailingTests`?

reference-tests/src/main/java/net/consensys/linea/MapFailedReferenceTestsTool.java
```diff
@@ -0,0 +1,170 @@
...
+        getModuleToConstraints(jsonString, jsonConverter);
+
+    mapFailedConstraintsToTestsToModule(failedConstraints, constraintToFailingTests, testName);
+
+    jsonString = jsonConverter.toJson(constraintToFailingTests);
+    writeToJsonFile(jsonString, jsonOutputPath);
+  }
+
+  private static void mapFailedConstraintsToTestsToModule(
+      Set<String> failedConstraints,
```
lorenzogentile404, line 57: are you creating here the different parts of the generated JSON? The name of the method does not look self-explanatory. I suggest to consider adding some comments or splitting the method into smaller ones.

lorenzogentile404 (changes requested):

reference-tests/src/test/resources/templates/BlockchainReferenceTest.java.template
bradbown: The `genereateReferenceBlockchainTest` gradle task uses this template to generate the these tests though right? Unless there is another way to generate these too?
If this is the only way then deleting the generated directory and rerunning this task regenerates tests with the `@ExtendWith(ReferenceTestWatcher.class)` again.

reference-tests/src/main/java/net/consensys/linea/MapFailedReferenceTestsTool.java
bradbown: I've renamed it to `getModuleToConstraints()` as it's responsible for converting the json string to a `List<ModuleToConstraints>` or if the string is empty then creating a new empty `List<ModuleToConstraints>`, renamed the corresponding variables which call this method too.

reference-tests/src/main/java/net/consensys/linea/MapFailedReferenceTestsTool.java
bradbown: Yes this is the main method for mapping the failed tests, constraint and its parent module together.
I've split the method into smaller ones, so hopefully the method names help describe what's going on at each stage better now.

lorenzogentile404 (approved):

PR #1115, opened 2024-09-06
[997] Migrate Stp and OobCall test to v2

gauravahuja:
#997 

Fixes the failures in StpTest and OobCallTest after migrating the V2.

The errors were:
- StpTest: transaction up-front cost 0x604d4d0f5a0 exceeds transaction sender account balance <value>
- OobCallTest:Transaction gas limit higher that avaliable in block

To fix them I updated the sender account balance to a higher value in StpTest and reduced the Tx gaslimit in OobCalltest.


OlivierBBB (approved):
LGTM. Were the `randBigInt(true)` an issue ?

PR #1120, opened 2024-09-06
fix CountOnlyModule

### LINKED ISSUE #1119: new instance of `CountOnlyModule` doesn't create new `CountOnlyOperation`
spotted by @fab-10

letypequividelespoubelles:
Counters for precompiles wheren't reboot when creating a new ZkTracer, pointing to the same `CountOnlyModule`


fab-10 (approved):
LGTM, just a consideration about Lombok that hides that classes are overriding `counts()` method

letypequividelespoubelles:
> LGTM, just a consideration about Lombok that hides that classes are overriding `counts()` method

Yes, I do agree, but as they have the exact same behaviour and expected , I'm OK with it for now.

Issue #1085, opened 2024-09-04
Nightly builds and tests


OlivierBBB:
Move the long tests (e.g. the unnamed tests that use broken up ranges) to a nightly run to keep the tests short.

letypequividelespoubelles:
cf #439

PR #1127, opened 2024-09-06
[1085] Add nightly tests tag and improve ci test run time

gauravahuja:
#1085 

In this PR:
- A new tag "nightly" for tests that run in a nightly workflow. 
- For now all Replay tests are tagged with "nightly"
- Use junit parallelism to run the tests
- Non nightly tests now run in ~6-7mins on github runner.
- 
image


arithmetization/src/test/java/net/consensys/linea/zktracer/module/mxp/MxpTest.java
```diff
@@ -41,9 +41,12 @@
 import org.hyperledger.besu.datatypes.Wei;
 import org.hyperledger.besu.ethereum.core.Transaction;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
 
 // https://github.com/Consensys/linea-besu-plugin/issues/197
 
+@Execution(ExecutionMode.SAME_THREAD)
 public class MxpTest {
```
gauravahuja, line 50: For some reason one of the MxpTests fails if run concurrently.

.github/workflows/gradle-nightly-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Upload test report
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: unit-test-report
+          path: arithmetization/build/reports/tests/
+
+  go-corset-tests:
+    runs-on: ubuntu-latest-128
```
FlorianHuc, line 68: do we need this runner or can it use the standard one?

gradle/tests.gradle
```diff
@@ -114,21 +120,14 @@ test {
...
-
-  minHeapSize = "4g"
-  maxHeapSize = "8g"
-  maxParallelForks = Runtime.getRuntime().availableProcessors().intdiv(2) ?: 1
-
+tasks.test.configure {
   useJUnitPlatform {
-    excludeTags("AcceptanceTest")
-    exclude('**/**ReferenceTest*')
+    excludeTags("nightly")
```
FlorianHuc, line 122: why don't you exclude the referenceTest anymore?
Are they excluded through some other mechanism?

.github/workflows/gradle-nightly-tests.yml
gauravahuja: Eventually I would like to move all reference tests in this workflow by tagging them as nightly so we need a runner with high memory. 

However for the workflows that run on every commit can use a standard one.

gradle/tests.gradle
gauravahuja: If you run `./gradlew :test` it will run all tests except the ones tagged "nightly". But we use `./gradlew :arithmetization:test` in the CI for unit tests, which runs tests under the `arithmetization` module. Reference test are present under the `reference-test` module so excluding them is not required if we only use `./gradlew :arithmetization:test`

gauravahuja: You can run reference tests in the following ways:
- `./gradlew :reference-test:test`
- `./gradlew referenceGeneralStateTests`

arithmetization/src/test/java/net/consensys/linea/zktracer/ReplayTests.java
```diff
@@ -52,6 +53,7 @@
  * One can run this command: scripts/capture.pl --start xxx --end yyy --output my-test-case.json.gz
  */
 @Slf4j
+@Tag("nightly")
```
OlivierBBB, line 57: So is the idea that we add this at the top of any class that contains replay tests (once we're done debugging them) ?

OlivierBBB (approved):
I'm not the right person to review this but it looks good to me. And AFAICT the only action that is required by the arithmetization people is to tag all classes containing replay tests as `@Tag{"nightly"}` (once we're done with debugging them, that is.) My remaining questions are as follows:
- can we have several classes using this tag ? I imagine the answer is yes
- which branch runs `"nightly"` tests, is it only `arith-dev` ?
- will `"nightly"` runs only happen when there is a new merge into `arith-dev` ?

I then have some questions around Intellij and what the above change means for us locally (if at all.)
- is the "junit parallelism" already implicitly true on Intellij ? If not, how can we enable it ?
- does this tag have any effect on running tests from the `arithmetization/src/test/` folder in Intellij ? If we run tests from that folder will it ignore all `"nightly"` tests ?

Sorry for those basic questions. I'm just trying to understand the implications of your PR and if/how we could use stuff you've set up in the CI pipeline (parallelization and tags) to improve our local experience.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mxp/MxpTest.java
letypequividelespoubelles: could be because of https://github.com/Consensys/linea-tracer/pull/1120, fixed now with latest arith-dev

arithmetization/src/test/java/net/consensys/linea/zktracer/ReplayTests.java
gauravahuja: At the top of the class or specific tests inside classes.

gauravahuja:
> * can we have several classes using this tag ? I imagine the answer is yes

We can tag multiple test classes or test methods with this tag.

> * which branch runs `"nightly"` tests, is it only `arith-dev` ?

It runs on the default branch for the repo. I think `arith-dev` is set as the default branch.

> * will `"nightly"` runs only happen when there is a new merge into `arith-dev` ?

So everyday on the set schedule it will checkout the arith-dev. Whatever is merged into arith-dev by that time will be used. I also included a workflow-dispatch option which can be used to manually trigger the run at any time on arith-dev.

> * is the "junit parallelism" already implicitly true on Intellij ? If not, how can we enable it ?

Depending on how your intellij is configured it can use the gradle to run tests which is configured to run them in parallel.

> * does this tag have any effect on running tests from the arithmetization/src/test/ folder in Intellij ? If we run tests from that 
folder will it ignore all "nightly" tests ?

When you run tests in intellij it will ask you to choose a run config, for now it shows tests and nightlyTests. If the test is tagged as nightly you should choose nightlyTests. This can be modified in the test configuration in intellij.

PR #1133, opened 2024-09-06
feat: exceptions vs tracedexceptions

### LINKED ISSUE #860: Exception tracing fix
Currently we trace all exceptions present in the stack's `exceptions` object (a `short`):
```java
trace
        .peekAtStack(true)
        // ...
        .pStackOpcx(Exceptions.invalidCodePrefix(exceptions))
        .pStackSux(Exceptions.stackUnderflow(exceptions))
        .pStackSox(Exceptions.stackOverflow(exceptions))
        .pStackMxpx(Exceptions.memoryExpansionException(exceptions))
        .pStackOogx(Exceptions.outOfGasException(exceptions))
        .pStackRdcx(Exceptions.returnDataCopyFault(exceptions))
        .pStackJumpx(Exceptions.jumpFault(exceptions))
        .pStackStaticx(Exceptions.staticFault(exceptions))
        .pStackSstorex(Exceptions.outOfSStore(exceptions))
        .pStackIcpx(contextExceptions.invalidCodePrefix())
        .pStackMaxcsx(contextExceptions.codeSizeOverflow())
```
But this _could_ fly in the face of the design of those exceptions depending on how we populate this object. In any case: we should make sure that only one exception is present in the `exceptions` object. There are two ways to do this:
1. change (?) the way we populate `exceptions`
2. create, for every `TraceSection`, an `effectiveException` which represents at most one exception, that being the exception that we will effectively trace

### LINKED ISSUE #1118: `FullExceptions` vs `TracedException`
We need two `Exceptions` objects:
- the "full" exceptions object that records all exceptions happening for a given instruction
- the "traced" exceptions object that records the exception which we wish to act upon

The `TracedException` will be used in the tracing of the stack-rows. We must have the property that
- `TracedException` is always pure (i.e. a single power of 2 or zero i.e. it's a `short` with at most one nonzero binary digit)
- `TracedException = 0 <=> FullExceptions = 0`
- If `TracedException = (1 << k)` then the `k`-th bit of `FullExceptions` is `1`

Also the `Gas.java` should be using the `TracedException` to determine the `outOfGasException` flag.

-----

Example of what we would like to achieve for the `CallSection`, for instance:

```java
    // STATICX cases
    if (Exceptions.staticFault(exceptions)) {
      this.commonValues.tracedException == Exceptions.STATIC_FAULT;
      return;
    }

    // ...

    // MXPX case
    if (Exceptions.memoryExpansionException(exceptions)) {
      this.commonValues.tracedException == Exceptions.MEMORY_EXPANSION_EXCEPTION;
      return;
    }

    // ...

    // OOGX case
    if (Exceptions.outOfGasException(exceptions)) {
      this.commonValues.tracedException == Exceptions.OUT_OF_GAS_EXCEPTION;
      this.oogXCall(hub);
      return;
    }
```

The end goal is to replace `exceptions` below with `tracedExceptions` in the tracing of `StackFragmet.java`

```java
        // Exception flag // this is WRONG
        .pStackOpcx(Exceptions.invalidCodePrefix(exceptions))
        .pStackSux(Exceptions.stackUnderflow(exceptions))
        .pStackSox(Exceptions.stackOverflow(exceptions))
        .pStackMxpx(Exceptions.memoryExpansionException(exceptions))
        .pStackOogx(Exceptions.outOfGasException(exceptions))
        .pStackRdcx(Exceptions.returnDataCopyFault(exceptions))
        .pStackJumpx(Exceptions.jumpFault(exceptions))
        .pStackStaticx(Exceptions.staticFault(exceptions))
        .pStackSstorex(Exceptions.outOfSStore(exceptions))
        .pStackIcpx(contextExceptions.invalidCodePrefix())
        .pStackMaxcsx(contextExceptions.codeSizeOverflow())
```

lorenzogentile404:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
```diff
@@ -45,6 +45,21 @@ private Exceptions() {}
   public static final short INVALID_CODE_PREFIX = 512;
   public static final short CODE_SIZE_OVERFLOW = 2048;
 
+  public static boolean isException(short exception) {
```
letypequividelespoubelles: we already have a method `isAny`, we don't need this duplicates method

letypequividelespoubelles (changes requested):
LGTM for the logic but you did it only for some specific opcode, it seems it misses 
1) all StackOnly section
2) the SUX / SOX cases where we even don't create the opcode section

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
OlivierBBB: Also, if anything, the name should be `isPureException` / `isPure` as if you have confluent exceptions your `exception` would look like e.g. `0b00101100`, here you are checking if `exception` looks like e.g. `0b00000100`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
```diff
@@ -263,17 +269,17 @@ public Trace trace(Trace trace) {
...
+        .pStackSux(tracedException == Exceptions.STACK_UNDERFLOW)
+        .pStackSox(tracedException == Exceptions.STACK_OVERFLOW)
+        .pStackMxpx(tracedException == Exceptions.MEMORY_EXPANSION_EXCEPTION)
+        .pStackOogx(tracedException == Exceptions.OUT_OF_GAS_EXCEPTION)
+        .pStackRdcx(tracedException == Exceptions.RETURN_DATA_COPY_FAULT)
+        .pStackJumpx(tracedException == Exceptions.JUMP_FAULT)
+        .pStackStaticx(tracedException == Exceptions.STATIC_FAULT)
+        .pStackSstorex(tracedException == Exceptions.OUT_OF_SSTORE)
+        .pStackIcpx(tracedException == Exceptions.INVALID_CODE_PREFIX)
+        .pStackMaxcsx(tracedException == Exceptions.CODE_SIZE_OVERFLOW)
```
OlivierBBB: Here you can do a static import of `Exceptions` to get rid of all the `Exceptions.` prefixes

OlivierBBB: You should also add somwhere where it makes sense, e.g. in the tracing here, a check that `exceptions` and `tracedExceptions` satisfy the properties from the issue. So that we get neither false positives nor false negatives in the tracing.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/halt/SelfdestructSection.java
```diff
@@ -107,6 +107,7 @@ public SelfdestructSection(Hub hub) {
 
     // STATICX case
     if (Exceptions.staticFault(exceptions)) {
+      commonValues.setTracedException(Exceptions.STATIC_FAULT);
```
OlivierBBB: Also I would suggest adding a bit called `tracedExceptionWasAlreadySet` in the `commonValues` object that is initially set to `false` when `tracedExceptions` is set to `NONE`, that gets switched on whenever we use one of the nontrivial `setTracedException` methods and that panics whenever you attempt to set it a second time.

We in any case need some sort of mechanism to make sure we don't modify `tracedExceptions` twice in the same `commonFragment`

lorenzogentile404: I believe a precondition check to see if when we set it is still NONE should be enough.

letypequividelespoubelles (changes requested):
You need to add the stack row before returning. Here in case of an exception. You won't trace anything

OlivierBBB:
Good point @letypequividelespoubelles 🙂   https://github.com/Consensys/linea-tracer/pull/1133#pullrequestreview-2290090850

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
```diff
@@ -263,17 +300,17 @@ public Trace trace(Trace trace) {
...
+        .pStackSux(tracedException == PureException.STACK_UNDERFLOW)
+        .pStackSox(tracedException == PureException.STACK_OVERFLOW)
+        .pStackMxpx(tracedException == PureException.MEMORY_EXPANSION_EXCEPTION)
+        .pStackOogx(tracedException == PureException.OUT_OF_GAS_EXCEPTION)
+        .pStackRdcx(tracedException == PureException.RETURN_DATA_COPY_FAULT)
+        .pStackJumpx(tracedException == PureException.JUMP_FAULT)
+        .pStackStaticx(tracedException == PureException.STATIC_FAULT)
+        .pStackSstorex(tracedException == PureException.OUT_OF_SSTORE)
+        .pStackIcpx(tracedException == PureException.INVALID_CODE_PREFIX)
+        .pStackMaxcsx(tracedException == PureException.CODE_SIZE_OVERFLOW)
```
OlivierBBB: Again, you can make all the `PureException.XXX` static imports.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
```diff
@@ -213,6 +221,35 @@ public Trace trace(Trace trace) {
...
+        tracedException != PureException.STATIC_FAULT || Exceptions.staticFault(exceptions));
+    Preconditions.checkArgument(
+        tracedException != PureException.OUT_OF_SSTORE || Exceptions.outOfSStore(exceptions));
+    Preconditions.checkArgument(
+        tracedException != PureException.INVALID_CODE_PREFIX
+            || Exceptions.invalidCodePrefix(exceptions));
+    Preconditions.checkArgument(
+        tracedException != PureException.CODE_SIZE_OVERFLOW
+            || Exceptions.codeSizeOverflow(exceptions));
+
```
OlivierBBB, line 236: It would be better to do this as a switch case (so we only verify one condition rather than all of them.)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/PureException.java
```diff
@@ -0,0 +1,31 @@
...
+  STACK_OVERFLOW,
+  MEMORY_EXPANSION_EXCEPTION,
+  OUT_OF_GAS_EXCEPTION,
+  RETURN_DATA_COPY_FAULT,
+  JUMP_FAULT,
+  STATIC_FAULT,
+  OUT_OF_SSTORE,
+  INVALID_CODE_PREFIX,
+  CODE_SIZE_OVERFLOW
+}
```
OlivierBBB: I would rename that class to `TracedException` as this is what its purpose is

OlivierBBB: I'm also wondering if we should have both
```java
public enum PureException {
  UNDEFINED,
  NONE,
  ...
}
```

OlivierBBB: The idea being that by default we set `tracedException` to `UNDEFINED`. Then the first thing we do is check whether `Exceptions.none(...)` in which case we set it to `NONE`. Otherwise we change it to the relevant thing during the instruction processing.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
OlivierBBB: We should also have the `NONE` case where we `checkArgument(Exceptions.none(exceptions))`. And if we do use `UNDEFINED` we should make the default case be that and throw an exception.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/SstoreSection.java
```diff
@@ -86,6 +88,7 @@ public SstoreSection(Hub hub, WorldView worldView) {
     miscForSstore.callOob(new SstoreOobCall());
 
     if (sstoreException) {
+      commonValues.setTracedException(PureException.OUT_OF_SSTORE);
       return;
     }
```
OlivierBBB: Below you are missing the `OUT_OF_GAS_EXCEPTION` case.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
```diff
@@ -57,7 +57,7 @@ public static boolean noStackException(final short bitmask) {
   }
 
   /**
-   * @return true if any exception flag has been raised
+   * @return true if any exception flag has been raised, i.e., at least one exception flag is raised
```
OlivierBBB, line 62: We should also look into the stack exceptions that are caught by the HUB before it ever gets to instruction processing. AFAICT these aren't covered in the above. These coverage issues really push me in the direction of implementing the `UNDEFINED` value in the `TracedException` class.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/PureException.java
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package net.consensys.linea.zktracer.module.hub.signals;
+
+public enum PureException {
+  NONE,
+  INVALID_OPCODE,
+  STACK_UNDERFLOW,
+  STACK_OVERFLOW,
```
OlivierBBB: I may have missed it but I don't think you ever use these values
```java
public enum PureException {
  INVALID_OPCODE,
  STACK_UNDERFLOW,
  STACK_OVERFLOW,
  ...
}
```

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/JumpSection.java
```diff
@@ -41,6 +42,7 @@ public JumpSection(Hub hub) {
     this.addStackAndFragments(hub);
 
     if (Exceptions.outOfGasException(hub.pch().exceptions())) {
+      commonValues.setTracedException(PureException.OUT_OF_GAS_EXCEPTION);
       return;
     }
```
OlivierBBB: Somewhere below we should have the option of throwing `INVALID_JUMP_EXCEPTION`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -60,6 +62,7 @@ public class CommonFragmentValues {
   @Setter public int numberOfNonStackRows;
   @Setter public boolean TLI;
   @Setter public int codeFragmentIndex = -1;
+  @Getter private PureException tracedException = PureException.NONE;
```
OlivierBBB: This is where I would prefer we use `UNDEFINED` rather than `NONE`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -121,6 +123,7 @@ public CreateSection(Hub hub) {
     // MXPX case
     if (mxpCall.mxpx) {
       scenarioFragment.setScenario(CREATE_EXCEPTION);
+      commonValues.setTracedException(PureException.MEMORY_EXPANSION_EXCEPTION);
       return;
     }
```
OlivierBBB: You are missing the `OUT_OF_GAS_EXCEPTION` case below

OlivierBBB (changes requested):

OlivierBBB:
I added some of the things. In particular the `UNDEFINED` variant for `TracedExceptions`. It will for sure not pass the tests as there are many cases where the `TracedException` wasn't set.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/PureException.java
OlivierBBB: I think that was probably solved in the latest commit.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
OlivierBBB: I think that may also have been resolved in the latest commit.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/AccountSection.java
```diff
@@ -46,6 +48,9 @@ public AccountSection(Hub hub) {
     if (hub.opCode().isAnyOf(OpCode.SELFBALANCE, OpCode.CODESIZE)) {
       if (Exceptions.any(exceptions)) {
         // the "squash parent return data" context row is all there is
+        Preconditions.checkArgument(Exceptions.outOfGasException(exceptions));
+        // NOTE: no other exception beyond OUT_OF_GAS_EXCEPTION are supposed to be managed here
+        commonValues.setTracedException(TracedException.OUT_OF_GAS_EXCEPTION);
         return;
       }
```
letypequividelespoubelles, line 53: Same issue here as in tackOnlySection : in case of exception you don't trace anything. Can you check for all sections that yih didn't introduce too early return plz ?

letypequividelespoubelles (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
letypequividelespoubelles: I made the same comment before 😇

lorenzogentile404: @OlivierBBB is this solved by this in `CommonFragmentValues` or it is something we should check even earlier?

```
   tracedException = noException ? TracedException.NONE : TracedException.UNDEFINED;

    if (instructionFamily == INVALID) {
      checkArgument(Exceptions.invalidOpcode(exceptions));
      setTracedException(TracedException.INVALID_OPCODE);
    }

    if (Exceptions.stackUnderflow(exceptions)) {
      setTracedException(TracedException.STACK_UNDERFLOW);
    }

    if (Exceptions.stackOverflow(exceptions)) {
      setTracedException(TracedException.STACK_OVERFLOW);
    }
```

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/AccountSection.java
lorenzogentile404: This was already like this before I added the `tracedException` @letypequividelespoubelles so, if it is an issue, we have it also in other branches.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
lorenzogentile404: Let's see what @powerslider suggest. In general, keeping the prefix is the way I suggest due to consistency, readability and avoid silent shadowing of definitions.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/AccountSection.java
lorenzogentile404: In general, I did not add any return except for `StackOnlyFragment`, that is now fixed.

letypequividelespoubelles: You're right, my bad 👌

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
OlivierBBB: I imagine the above will have solved this. We have to trace a `CommonFragment` in all cases, whether we have a stack exception or not. So this should be sufficient.

letypequividelespoubelles: In case of an exception, we create a StackOnlySection, so imho, it's better to set for INVALID / SUX / SOX this tracedException before the OOGX. 
image

lorenzogentile404: @OlivierBBB @letypequividelespoubelles what should we agree on?

OlivierBBB: I am not understanding your last comment. Could you please rephrase it @letypequividelespoubelles ?

OlivierBBB: My understanding of our previous discussion was that doing it in the common section may not be the best since we would be doing the associated computations in execution phases where it doesn't matter (TX_INIT, TX_FINL, TX_SKIP). Was this what you meant @letypequividelespoubelles or did you mean something else ?

letypequividelespoubelles: > We should also look into the stack exceptions that are caught by the HUB before it ever gets to instruction processing.

In case of exceptions caught by the HUB before processing the opcode (INVALID, SUX and then SOX), we will create a `StackOnlySection` . IMHO, it would be better to set the tracedException in the `StackOnlySection` instead of setting it in the `CommonFragment`.
Impact : 
- more consistent with other BlaBlaSection : this is where we set the tracedException in all case
- tiny tiny better perf : we'll do the 3 checks check only for StackSection, and not for all section (so divide by ~ 2 the number of checks, as we have roughly half sectiuons which are StackOnlySection

letypequividelespoubelles: Note that I'm talking about StackOnlySection, not StackFragment

letypequividelespoubelles: And if we really want to reduce the number of check, we could introduce a `PrecoceExceptionSection`. Let me push what I have in mind.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
lorenzogentile404: @powerslider suggested to keep the prefix for enums too.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobCallTest.java
```diff
@@ -98,8 +99,45 @@ void TestRecursiveCallsWithBytecode() {
...
 
     assertTrue(Exceptions.none(hub.pch().exceptions()));
+  }
+
+  /**
+   * Same as {@link #TestRecursiveCallsWithBytecode()} but with an ADD opcode at the end triggering
+   * SUX
+   */
+  @Test
+  void TestRecursiveCallsWithBytecodeFollowedByStackUnderflow() {
```
powerslider, line 109: Please rename tests to start with a lower case letter.

lorenzogentile404: Let's do it in a separate PR project-wise.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
OlivierBBB: This is not resolved AFAICT @lorenzogentile404

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -60,18 +65,20 @@ public class CommonFragmentValues {
   @Setter public int numberOfNonStackRows;
   @Setter public boolean TLI;
   @Setter public int codeFragmentIndex = -1;
+  @Getter private TracedException tracedException;
```
OlivierBBB, line 68: I think by default we should set it to `UNDEFINED` @lorenzogentile404. Do you not agree ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/halt/StopSection.java
```diff
@@ -47,6 +48,11 @@ public StopSection(Hub hub) {
     // 3 = 1 + max_NON_STACK_ROWS in message call case
     // 5 = 1 + max_NON_STACK_ROWS in deployment case
     super(hub, hub.callStack().currentCallFrame().isMessageCall() ? (short) 3 : (short) 5);
+    final short exceptions = hub.pch().exceptions();
+    checkArgument(
+        Exceptions.none(exceptions),
+        "STOP is incapable of triggering an exception but exceptions = " + exceptions);
+
```
OlivierBBB, line 57: What would make sense is to write a `prettyPrint` method for exceptions where we print something

```java
underlying OpCode: <mnemonic>
raw exceptions:
  stackUnderflowException: ✅
  stackOverflowException:
  InvalidOpcodeException: 
  MemoryExpansionException ✅
  OutOfGasException: ✅
  ...
  None ✅
```
(we only print the present exceptions along with `None` (the above example makes no sense, it's just an example.)

OlivierBBB: See https://github.com/Consensys/linea-tracer/issues/1163

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/copy/ReturnDataCopySection.java
```diff
@@ -47,6 +48,7 @@ public ReturnDataCopySection(Hub hub) {
 
     // returnDataCopyException case
     if (returnDataCopyException) {
+      commonValues.setTracedException(TracedException.RETURN_DATA_COPY_FAULT);
```
OlivierBBB, line 51: I'm thinking that if we do a check for `RETURNDATACOPY` and the presence of a `RETURN_DATA_COPY_FAULT` (which, BTW, we should rename to `RETURN_DATA_COPY_EXCEPTION`) first then we can include the COPY family in the the families where we have the high level check.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
```diff
@@ -145,6 +146,7 @@ public CallSection(Hub hub) {
 
     // MXPX case
     if (Exceptions.memoryExpansionException(exceptions)) {
+      commonValues.setTracedException(TracedException.MEMORY_EXPANSION_EXCEPTION);
```
OlivierBBB, line 149: After re-reading this I believe we can include the `CALL` family in the high level `TracedException` setting.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -132,6 +134,7 @@ public CreateSection(Hub hub) {
     // OOGX case
     if (Exceptions.outOfGasException(exceptions)) {
       scenarioFragment.setScenario(CREATE_EXCEPTION);
+      commonValues.setTracedException(TracedException.OUT_OF_GAS_EXCEPTION);
```
OlivierBBB, line 137: Like for `CALL`'s I think nothing is stopping us from including the `CREATE` family in the high-level `TracedException` setting.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/AccountSection.java
```diff
@@ -46,6 +47,7 @@ public AccountSection(Hub hub) {
     if (hub.opCode().isAnyOf(OpCode.SELFBALANCE, OpCode.CODESIZE)) {
       if (Exceptions.any(exceptions)) {
         // the "squash parent return data" context row is all there is
+        Preconditions.checkArgument(Exceptions.outOfGasException(exceptions));
```
OlivierBBB, line 50: This is true because we never enter here in case of a `STACK_OVERFLOW_EXCEPTION`, but I would include it as an explanatory comment.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -82,13 +89,45 @@ public CommonFragmentValues(Hub hub) {
...
-                    || instructionFamily == InstructionFamily.CREATE
-                    || instructionFamily == InstructionFamily.HALT
-                    || instructionFamily == InstructionFamily.INVALID)
+            && ((instructionFamily == CALL
+                    || instructionFamily == CREATE
+                    || instructionFamily == HALT
+                    || instructionFamily == INVALID)
                 || exceptionAhoy);
+
+    tracedException = noException ? TracedException.NONE : TracedException.UNDEFINED;
```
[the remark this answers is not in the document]
OlivierBBB: This way we don't waste computation in the following checks. And this will be the case 99% of the time.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
OlivierBBB: I think by default we should set it to `UNDEFINED` @lorenzogentile404. Do you not agree ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
OlivierBBB: I would suggest (to not waste computation) adding
```java
if (tracedException == TracedException.NONE) {
  return;
}
```
here. This way we don't waste computation in the following checks. And this will be the case 99% of the time.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -82,13 +89,45 @@ public CommonFragmentValues(Hub hub) {
...
+    if (Exceptions.staticFault(exceptions)) {
+      checkArgument(opCode.mayTriggerStaticException());
+      setTracedException(TracedException.STATIC_FAULT);
+      return;
+    }
+
+    // For these instruction families we set the traced exception below manually
+    if (instructionFamily.isAnyOf(COPY, CALL, CREATE, HALT, STORAGE)) {
+      return;
+    }
```
OlivierBBB, line 130: So here, given my other comments, I believe we can include
```java
InstructionFamily.CALL
InstructionFamily.CREATE
```
Also if we do a special check for `RETURNDATACOPY` and for `SSTORE` between the static and this exclusion of instruction families we could also include the following instruction families.
```java
InstructionFamily.COPY
InstructionFamily.STORAGE
```

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/AccountSection.java
OlivierBBB, line 51: This is true because we never enter here in case of a `STACK_OVERFLOW_EXCEPTION`, but I would include it as an explanatory comment.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
OlivierBBB: Like for `CALL`'s I think nothing is stopping us from including the `CREATE` family in the high-level `TracedException` setting.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
OlivierBBB: After re-reading this I believe we can include the `CALL` family in the high level `TracedException` setting.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/SloadSection.java
```diff
@@ -85,6 +86,10 @@ public SloadSection(Hub hub, WorldView worldView) {
     hub.state.updateOrInsertStorageSlotOccurrence(storageSlotIdentifier, doingSload);
 
     this.addStackAndFragments(hub, readCurrentContext, miscFragmentForSload, doingSload);
+
+    if (Exceptions.outOfGasException(exceptions)) {
+      commonValues.setTracedException(TracedException.OUT_OF_GAS_EXCEPTION);
+    }
```
OlivierBBB: We can include the `STORAGE` family in the high level `TracedException` setting, see review comment in
arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/SstoreSection.java

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/SstoreSection.java
```diff
@@ -86,11 +87,16 @@ public SstoreSection(Hub hub, WorldView worldView) {
     miscForSstore.callOob(new SstoreOobCall());
 
     if (sstoreException) {
+      commonValues.setTracedException(TracedException.OUT_OF_SSTORE);
```
OlivierBBB: We can include the `STORAGE` family in the high level `TracedException` setting if we detect the `SSTORE` specific  exception first.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/copy/CallDataCopySection.java
```diff
@@ -51,12 +52,14 @@ public CallDataCopySection(Hub hub) {
...
     // The MXPX case
     if (mxpCall.mxpx) {
+      commonValues.setTracedException(TracedException.MEMORY_EXPANSION_EXCEPTION);
       return;
     }
 
     // The OOGX case
     if (Exceptions.any(exceptions)) {
       checkArgument(exceptions == OUT_OF_GAS_EXCEPTION);
+      commonValues.setTracedException(TracedException.OUT_OF_GAS_EXCEPTION);
```
OlivierBBB: The setting would disappear if we do include the `COPY` family in the high-level processing in `CommonFragmentValues`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/copy/CodeCopySection.java
```diff
@@ -51,12 +53,14 @@ public CodeCopySection(Hub hub) {
...
     if (mxpCall.mxpx) {
+      commonValues.setTracedException(TracedException.MEMORY_EXPANSION_EXCEPTION);
       return;
     }
 
     // The OOGX case
     if (Exceptions.any(exceptions)) {
-      checkArgument(exceptions == Exceptions.OUT_OF_GAS_EXCEPTION);
+      checkArgument(exceptions == OUT_OF_GAS_EXCEPTION);
+      commonValues.setTracedException(TracedException.OUT_OF_GAS_EXCEPTION);
```
OlivierBBB: The setting would disappear if we do include the `COPY` family in the high-level processing in `CommonFragmentValues`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/copy/ExtCodeCopySection.java
```diff
@@ -98,6 +100,7 @@ public ExtCodeCopySection(Hub hub) {
               .makeWithTrm(accountBefore, accountBefore, rawAddress, doingDomSubStamps);
 
       this.addFragment(accountReadingFragment);
+      commonValues.setTracedException(TracedException.OUT_OF_GAS_EXCEPTION);
```
OlivierBBB: The setting would disappear if we do include the `COPY` family in the high-level processing in `CommonFragmentValues`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/copy/ReturnDataCopySection.java
OlivierBBB: I'm thinking that if we do a check for `RETURNDATACOPY` and the presence of a `RETURN_DATA_COPY_FAULT` (which, BTW, we should rename to `RETURN_DATA_COPY_EXCEPTION`) first then we can include the `COPY` family in the the families where we have the high level check.

The setting would disappear if we do include the `COPY` family in the high-level processing in `CommonFragmentValues`

OlivierBBB (changes requested):

letypequividelespoubelles (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -60,18 +65,20 @@ public class CommonFragmentValues {
   @Setter public int numberOfNonStackRows;
   @Setter public boolean TLI;
   @Setter public int codeFragmentIndex = -1;
+  @Getter private TracedException tracedException = UNDEFINED;
 
   public CommonFragmentValues(Hub hub) {
-    final boolean noStackException = !Exceptions.stackException(hub.pch().exceptions());
-    final InstructionFamily instructionFamily = hub.opCode().getData().instructionFamily();
+    short exceptions = hub.pch().exceptions();
```
letypequividelespoubelles: Could be final

lorenzogentile404: Fixed

PR #1143, opened 2024-09-09
feat: tidy up `ToyExecutionEnvironment` and remove `ToyWorld`

DavePearce:
This tidies up the `ToyExecutionEnvironment` in several ways.  Firstly, it updates the manner in which the `MainnetTransactionProcessor` is created by reusing as much from BESU as possible; secondly, it removes `ToyWorld` altogether (as this is no longer needed); finally, it removes `ToyExecutionEnvironment.execute()` since this is no longer being used.


arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/ConflationSnapshot.java
```diff
@@ -29,4 +34,44 @@ public record ConflationSnapshot(
...
+    List<BlockHashSnapshot> blockHashes) {
+
+  /**
+   * Construct a block hash map for any block hashes embedded in this conflation.
+   *
+   * @return
+   */
+  public BlockHashLookup toBlockHashLookup() {
+    BlockHashMap map = new BlockHashMap();
+    // Initialise block hashes.  This can be null for replays which pre-date support for block hash
```
OlivierBBB, line 46: It may be worth re-capturing them. We have lists and @gusiri 's `bulk_capture.sh` script.

arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/ConflationSnapshot.java
```diff
@@ -29,4 +34,44 @@ public record ConflationSnapshot(
...
+    // capture and, hence, we must support this case (at least for now).
```
OlivierBBB, line 47: It may be worth re-capturing them. We have lists and @gusiri 's `bulk_capture.sh` script.

testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
```diff
@@ -0,0 +1,286 @@
...
+  private static final CorsetValidator CORSET_VALIDATOR = new CorsetValidator();
+
+  /**
+   * Determines whether transaction results should be checked against expected results embedded in
+   * replay files. This gives an additional level of assurance that the tests properly reflect
+   * mainnet (or e.g. sepolia as appropriate). When this is set to false, replay tests will not fail
+   * even if the tx outcome differs from what actually occurred on the relevant chain (e.g.
+   * mainnet). Such scenarios have arisen, for example, when <code>ZkTracer</code> has an unexpected
+   * side-effect on tx execution (which it never should).
+   */
```
OlivierBBB, line 85: I guess we don't know what could cause any remaining side effects at this point do we ? It may be interesting to have more information than just a boolean.

testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
```diff
@@ -0,0 +1,286 @@
...
+      conflation = gson.fromJson(replayFile, ConflationSnapshot.class);
+    } catch (Exception e) {
+      log.error(e.getMessage());
+      return;
+    }
+    this.executeFrom(chainId, conflation);
+    this.checkTracer();
+  }
+
+  public void replay(BigInteger chainId, final Reader replayFile, String inputFilePath) {
```
OlivierBBB, line 131: These two methods should be mergeable. I don't know if this is a use case for `Optional<String>` or for `String... inputFilePath` but there ought to be a way to unite these.

testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  /**
+   * Initialise a world updater given a conflation. Observe this can be applied to any WorldUpdater,
+   * such as SimpleWorld.
+   *
+   * @param world The world to be initialised.
+   * @param conflation The conflation from which to initialise.
+   */
+  private static void initWorld(WorldUpdater world, final ConflationSnapshot conflation) {
```
OlivierBBB, line 243: Does a world that was thusly initialized know about precompiles ? I ask because in our old setup we had to invoke a special function to make the execution environment and the world aware of precompiles.

OlivierBBB (approved):
LGTM (however: I mostly skipped reading `getMainnetTransactionProcessor`)

arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/ConflationSnapshot.java
DavePearce: Yeah, I agree with that.  But, we need to retain the ability to run without for now, otherwise it will break stuff for others.

testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
DavePearce: We don't know what we don't know!  But, also, there are other scenarios you should consider:

1.  Someone from the team could accidentally reintroduce a call to a side-effecting method (such as `gasAvailableForChildCall()`).  
2. BESU itself could be changed in a way that makes a previously pure method now have side-effects.  

We want to get notified of this immediately, in the event anything like this happens.

testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
DavePearce: So, yeah, I do agree with this.  I looked into it briefly, but I need to investigate in more detail what exactly is going on with the `inputFilePath`.  For now, I didn't think it was worth investing time in.

gauravahuja (approved):

arithmetization/src/test/java/net/consensys/linea/zktracer/ReplayTests.java
```diff
@@ -200,54 +206,54 @@ void fatMxp() {
   @Test
   void bulkReplay() {
     // bulkReplay("./src/test/resources/replays");
```
gusiri, line 208: @DavePearce 
Can we also update this comment to inlcude LINEA_MAINNET?
// bulkReplay(LINEA_MAINNET, "./src/test/resources/replays");

DavePearce: Yes, we can do.  Create a separate issue and I will add support there.  I haven't completely grokked the `builkReplay` system, so I need to talk it over with you first, etc.

gusiri: Thank you 👍

DavePearce: It would be good also to merge the two replay methods somehow.  I will have a look at that also

PR #1150, opened 2024-09-10
feat: enable reporting in `go-corset`

### LINKED ISSUE #1147: Enable Reporting for `go-corset`
`go-corset` now supports human-readable error reports for failing constraints (roughly similar to what was possible in `corset`).  This should be enabled so that debug information can be extracted from the CI pipeline when checking fails.

DavePearce:
This enables reporting in `go-corset` so we can get more debug information about failing constraints.


gauravahuja (approved):

OlivierBBB (approved):
LGTM

PR #1152, opened 2024-09-10
1049 fix daily blockchain workflow

bradbown:


reference-tests/build.gradle
```diff
@@ -52,6 +58,11 @@ tasks.register('referenceBlockchainTests', Test) {
   description = 'Runs ETH reference blockchain tests.'
   dependsOn generateBlockchainReferenceTests
 
+  boolean isCiServer = System.getenv().containsKey("CI")
+  minHeapSize = isCiServer ? "8g" :"4g"
+  maxHeapSize = isCiServer ? "120g" : "8g"
```
gauravahuja: I'm not sure if this will work, there is 128GB ram on the runner and if you reserve 120G for jvm, there isn't any left for corset.

gradle/tests.gradle
```diff
@@ -110,7 +110,7 @@ tasks.withType(Test).configureEach {
     showSkippedStandardStreams false
 
     // set to false to hide failed standard out and error streams
-    showFailedStandardStreams true
+    showFailedStandardStreams false
```
gauravahuja, line 113: Why do you want to set this to false?

reference-tests/build.gradle
```diff
@@ -52,6 +58,11 @@ tasks.register('referenceBlockchainTests', Test) {
   description = 'Runs ETH reference blockchain tests.'
   dependsOn generateBlockchainReferenceTests
 
+  boolean isCiServer = System.getenv().containsKey("CI")
+  minHeapSize = isCiServer ? "8g" :"4g"
+  maxHeapSize = isCiServer ? "20g" : "8g"
+  jvmArgs '-Xmx512m', '-Xms128m'
```
gauravahuja: Wait, Xms and Xmx are used to set min and max heap size which are already set using minHeapSize and maxHeapSize. I am not sure which one will take effect.

reference-tests/src/test/java/net/consensys/linea/ReferenceTestWatcher.java
```diff
@@ -45,4 +52,22 @@ public void testFailed(ExtensionContext context, Throwable cause) {
     MapFailedReferenceTestsTool.mapAndStoreFailedReferenceTest(
         testName, logEventMessages, JSON_OUTPUT_FILENAME);
   }
+
+  @Synchronized
+  private static ch.qos.logback.classic.Logger getLogbackLogger() {
+    try {
+      org.slf4j.Logger slf4jLogger = null;
+      for (int i = 0; i < LOGBACK_POLL_ATTEMPTS; i++) {
```
gauravahuja, line 59: Why are these attempts needed? Is there way to alter the instantiating order of this Test watcher and the root logger so that the root logger will be available before the test watcher is instantiated.

reference-tests/src/test/resources/templates/BlockchainReferenceTest.java.template
```diff
@@ -33,23 +37,24 @@ import static org.junit.jupiter.api.Assumptions.assumeTrue;
...
 
   private static final String FAILED_MODULE = "%%FAILED_MODULE%%";
   private static final String FAILED_CONSTRAINT = "%%FAILED_CONSTRAINT%%";
-
   private static final String[] TEST_CONFIG_FILE_DIR_PATH = new String[] {%%TESTS_FILE%%};
-  private static final String FAILED_TESTS_FILE_PATH = "%%FAILED_TEST_FILE_PATH%%";
 
-  public static Stream<Arguments> getTestParametersForConfig() {
+  public static Stream<Arguments> getTestParametersForConfig()
+      throws ExecutionException, InterruptedException {
```
gauravahuja: Doew generateTestParametersForConfig throw these exceptions?

reference-tests/src/test/java/net/consensys/linea/ReferenceTestWatcher.java
bradbown: This was added because there were occasions when `substituteLogger` was being retrieved. Since adding this I have excluded some jars from conflicting dependencies so I'll test whether this is still required.

reference-tests/src/test/resources/templates/BlockchainReferenceTest.java.template
bradbown: This was due to calling `get()` on a completable future, I've reworked the usage and this is no longer required.

reference-tests/src/test/java/net/consensys/linea/ReferenceTestWatcher.java
bradbown: It looks like it is still required.

gradle/tests.gradle
bradbown: This was when I kept having the CI pipeline being cancelled without explanation so in an attempt to save memory I was removing the logs in the meantime.

reference-tests/build.gradle
bradbown: I've removed the xmx and xms args.

gauravahuja (approved):

PR #1168, opened 2024-09-12
[1140] Some optimizations for CI and Reference tests

gauravahuja:
#1140


reference-tests/src/test/java/net/consensys/linea/GeneralStateReferenceTestTools.java
```diff
@@ -81,11 +81,10 @@ private static ProtocolSpec protocolSpec(final String name) {
               (testName, fullPath, stateSpec, collector) -> {
                 final String prefix = testName + "-";
                 for (final Map.Entry<String, List<GeneralStateTestCaseEipSpec>> entry :
-                    stateSpec.finalStateSpecs().entrySet().stream()
-                        .filter(e -> e.getKey().equalsIgnoreCase("London"))
```
gauravahuja, line 85: - This filter results in 0 test configs for some GeneralStateReferenceTests. 
- Since those are parameterized tests and if there are no tests  parameters generated, junit throws and excpection `Configuration error: You must configure at least one set of arguments for this @ParameterizedTest`. 
- This change will mitigate this exception and simply skip the non london tests.

reference-tests/src/test/java/net/consensys/linea/GeneralStateReferenceTestTools.java
```diff
@@ -137,7 +137,17 @@ public static Collection<Object[]> generateTestParametersForConfig(final String[
...
+    final BlockHeader blockHeader =
+        BlockHeaderBuilder.createDefault()
+            .baseFee(Wei.of(LINEA_BASE_FEE))
+            .gasLimit(LINEA_BLOCK_GAS_LIMIT)
+            .difficulty(Difficulty.of(LINEA_DIFFICULTY))
+            .number(spec.getBlockHeader().getNumber())
+            .coinbase(spec.getBlockHeader().getCoinbase())
+            .timestamp(spec.getBlockHeader().getTimestamp())
+            .parentHash(spec.getBlockHeader().getParentHash())
+            .buildBlockHeader();
```
OlivierBBB: Is the idea that you are post-hoc modifying the `baseFee`, `gasLimit` and `difficulty` parameters of the block header to reflect expected values of the arithmetization ?

OlivierBBB: The thing that I'm slightly worried about is whether this could lead to a different outcome state for certain tests. I imagine that among the things that the reference tests check is whether the final state root hash conforms to a predicted one. E.g. if the bytecode contains one of these opcodes
```rust
BASEFEE
DIFFICULTY
```
and, say, `SSTORE`s one of these values or uses them in a nontrivial way that leads to different branching behaviour (say `If basefee < xxx Then bla Else bla'`, not that it makes sense) then we may end up diverging from the expected execution.

gauravahuja: Removed this part, as I am using the new zkevm binary for reference tests.

.github/workflows/gradle-ethereum-tests.yml
```diff
@@ -3,6 +3,8 @@ name: reference tests
 on:
   schedule:
     - cron: "0 23 * * *"
+  pull_request:
```
gauravahuja: I will remove pull_request before merging this PR. It is there only to run the reference test job for commits on this PR.

.github/workflows/gradle-ethereum-tests.yml
```diff
@@ -30,38 +31,84 @@ jobs:
       - name: Install Rust
```
OlivierBBB, line 31: Wouldn't we want to rename the job to `ethereum-tests-rust-corset` ?

.github/workflows/gradle-ethereum-tests.yml
```diff
@@ -30,38 +31,84 @@ jobs:
...
+
+      - name: Install Go Corset
+        run: go install github.com/consensys/go-corset/cmd/go-corset@latest
+
+      - name: Generate General State Reference Tests
+        run: ./gradlew generateGeneralStateReferenceTests -Dorg.gradle.parallel=true -Dorg.gradle.caching=true
         env:
           JAVA_OPTS: -Dorg.gradle.daemon=false
-          ZKEVM_BIN: zkevm_for_reference_tests.bin
           CORSET_FLAGS: fields,expand,expand,expand
```
OlivierBBB, line 99: Question: both `ethereum-tests` and `ethereum-tests-go-corset` seem to be using `CORSET_FLAGS` which as I understand it is for rust corset. Is that on puropse ? Also `CORSET_FLAGS` gets redefined in the next paragraph.

OlivierBBB (approved):

.github/workflows/gradle-ethereum-tests.yml
gauravahuja: I was following the existing convention, where had `tests` and `tests-go-corset`

.github/workflows/gradle-ethereum-tests.yml
gauravahuja: > Also CORSET_FLAGS gets redefined in the next paragraph

There are two steps and each step has to define its environment variables. There is no sharing of env variables between steps.
  1. Generate the tests `./gradlew generateGeneralStateReferenceTests` 
  2. Run the tests `./gradlew referenceGeneralStateTests`

Actually the test generating step does not require corset flags and can be removed. I will do that in a followup PR.

Corset flags used in the step that runs the tests are as follows:

- ethereum-tests
  - `CORSET_FLAGS: fields,expand,expand,expand` 
- ethereum-tests-go-corset 
  - `CORSET_FLAGS: disable`
  - `GOCORSET_FLAGS: -wd --ansi-escapes=false --report --air`

These are the same flags used in the workflow that runs on every PR commit.

PR #1170, opened 2024-09-12
feat: use .tmp file extension while writing a conflated trace in order to indicate that the trace is incomplete if it stays with the .tmp extension

powerslider:


arithmetization/src/main/java/net/consensys/linea/plugins/rpc/tracegeneration/GenerateConflatedTracesV2.java
```diff
@@ -109,32 +113,41 @@ private TraceService getTraceService() {
...
-    return fileName.toAbsolutePath().toString();
+    // Generate the original and final trace file name.
+    final String origTraceFileName = generateOutputFileName(traceRequestParams);
+    // Generate and resolve the original and final trace file path.
+    final Path origTraceFilePath = generateOutputFilePath(origTraceFileName + TRACE_FILE_EXTENSION);
+
+    // Write the trace at the original and final trace file path, but with the suffix .tmp at the
+    // end of the file.
+    final Path tmpTraceFilePath =
+        tracer.writeToTmpFile(tracesOutputPath, origTraceFileName + ".", TRACE_TEMP_FILE_EXTENSION);
```
FlorianHuc, line 127: don't we have a '.' too many?

I would have found it easier to follow with final_name = xxx-yyy.lt

and

temp_name = final_name + ".tmp"

FlorianHuc (approved):

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/tracegeneration/GenerateConflatedTracesV2.java
powerslider: No, this dot is needed because `Files.createTempFile(rootDir, prefix, suffix)` appends <timestamp>.<suffix> at the end and there is no dot symbol before the timestamp so with this approach I am setting the suffix to be `.lt.tmp` and the renamed file will end with only `.lt`. Here is a real example from my local testing:

```
// Temp file name
600-700.conflated.0.6.0-rc1.6783354941660875494.lt.tmp

// Final file name
600-700.conflated.0.6.0-rc1.lt
```

PR #1174, opened 2024-09-12
Rectification of `AccountSnapshot`'s for some of the `CallSection` after care methods

### LINKED ISSUE #1169: insufficient balance
## Ranges

```rust
2746060-2746060
```

# Issue

In mainet with release v0.6.0-rc1:

Name: BLOCK_NUMBER, Length: 84001, dtype: int64
Error fetching or parsing JSON: 'result' {'jsonrpc': '2.0', 'id': 1, 'error': {'code': -32603, 'message': 'Plugin internal error', 'data': 'Exceptions triggered while tracing:\n  - Insufficient balance\n\t\tAddress: 0x2fcb398bcb5bd2aec807941f9d7953e1135056b8\n\t\tBalance: 0x0000000000000000000000000000000000000000000000000000000000000000\n\t\tValue: 0x000000000000000000000000000000000000000000000000006a94d74f430000\n'}}
Failed to fetch or parse JSON data for blockNumber=2746060: 'result'

OlivierBBB:
The main point is that we cannot use the snapshot of caller / callee at `REVERT` time as the thing we revert to. Because that snapshot is "immutable" (always the same") and this is incompatible with the incremental undoing of past actions.


arithmetization/src/test/java/net/consensys/linea/zktracer/Issue1169Tests.java
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package net.consensys.linea.zktracer;
+
+import static net.consensys.linea.testing.ReplayExecutionEnvironment.LINEA_MAINNET;
+import static net.consensys.linea.zktracer.ReplayTests.replay;
+
+import org.junit.jupiter.api.Test;
+
+// @Tag("nightly")
```
letypequividelespoubelles, line 22: if they are replayTest we want to run at each commit, isn't better to move them in the ReplayTest.java ?

letypequividelespoubelles (approved):
LGTM

arithmetization/src/test/java/net/consensys/linea/zktracer/Issue1169Tests.java
OlivierBBB: We could. I don't have a strong opinion on this. I use those separate files to prevent the `ReplayTests.java` file from becoming too obese.

PR #1176, opened 2024-09-12
Preliminary selection of nightly tests

### LINKED ISSUE #1135: Mark all the replay tests as `@Tag{"Nightly"}`

OlivierBBB:


letypequividelespoubelles (commented):
@OlivierBBB could it make sense to remove the nightly replay test from replay test.Java to a new Java classes ? So that in replay test.java we have only replay tests tagged as replay ? It would be cleaner and we could have annotation at the class level only.

OlivierBBB:
@letypequividelespoubelles that is a good idea. Thanks 🙂, I'll do it.

OlivierBBB:
@letypequividelespoubelles The ReplayTests.java file was split in 3, see commit description for https://github.com/Consensys/linea-tracer/pull/1176/commits/0956c48597dca876125c49137ba384ffb2404e6b

Issue #996, opened 2024-08-27
Multi-block tests in the ToyWorld


OlivierBBB:
We need multi-block tests for the aforementioned reasons:
- integration tests with state manager (with @arijitdutta67 and @bogdanbear)
- testing of BLOCKHASH module (for @letypequividelespoubelles )

PR #1178, opened 2024-09-13
[996] Add features to write multi block tests

gauravahuja:
#996


testing/src/main/java/net/consensys/linea/testing/ToyAccount.java
```diff
@@ -48,7 +48,7 @@ public class ToyAccount implements MutableAccount {
   private Wei balance;
   private Bytes code;
   private Supplier<Hash> codeHash = Suppliers.memoize(() -> Hash.hash(code));
-  private final Map<UInt256, UInt256> storage = new HashMap<>();
+  final Map<UInt256, UInt256> storage = new HashMap<>();
```
DavePearce, line 51: I would think getter / setters here is better than making it package visible.

DavePearce (commented):
LGTM

DavePearce (approved):

OlivierBBB:
@gauravahuja is this PR ready to be merged ?

PR #1179, opened 2024-09-13
CI - Daily workflow run on weekdays

bradbown:
Changed schedule jobs to only execute on weekdays


letypequividelespoubelles:
Could we not run the nightly if nothing was merged since the last nightly run instead ?

OlivierBBB:
See https://github.com/Consensys/linea-tracer/issues/1195 @bradbown @letypequividelespoubelles

gauravahuja (approved):

Issue #823, opened 2024-07-05
Eword/UInt256 usage optimizations


lu-pinto:
I was looking at the usage of tuweni API, mostly focused on the `EWord`. I see that it's also extensively used in Besu core (`Wei`, `Gas` ) however I believe this creates a good amount of memory churn (short lived objects in the heap), and by itself also makes the CPU work harder, that could maybe be avoided if we mostly use static calls.

### Rationale
So looking at `EWord`, it extends from `BaseUInt256Value` which is simply a wrapper around a `UInt256` type that is the actual value. Inside `EWord` we have the `UInt256` stored value and `ctor` , a lambda function that defines the behaviour when a mathematical operation is done (add, subtract, modulus, etc). Currently it is set to `Eword::new` that means a new instance is created every time. Also when of methods are called new instances are created. In a nutshell, for each eword abstraction there has to be one `EWord` instance backed by 1 or less `UInt256` instances which is a waste. That's because `UInt256` is slightly optimised to pre-initialise the first 64 unsigned ints and cache them statically. If you think we are in this range (I don't think so) the optimisation would even yield greater results.
We have also saw these instances in the heap dump we took for the huge transaction case. `StorageFragment` uses 4 `EWords` and some other types of fragment also make use of 1. Calls to the `Oob` module also make good use of it, and that's all I looked at, there is surely more I believe.

### Proposal
[PREFERRED] We could drop usage of the `EWord` type and just declare a `UInt256` instead,. All instance methods would be replaced with statics. Drawbacks: You will loose the typed benefits of having `EWord` declared type, but as a counter-argument I think you can also name the variables accordingly if you really need it.
Make `EWord` a lambda and extend from `Supplier<UInt256>` . To me this looks a better API as `BaseUInt256Value` is not more than a value holder (supplier). In this case you could still declare the type as it is now and access the value through `EWord.get()`. I think I'm still able to make instance methods non static through default method in the functional interface but I need to check. This would simplify greatly the class hierarchy and remove things you aren't really using (like copy?) Drawbacks: Memory layout of the lambda is still better than `BaseUInt256Value` but there is some overhead since lambdas are instances in the heap. Example of memory layout of `Supplier<Long>`:
```
Test$$Lambda/0x0000000800000a00 object internals:
OFF  SZ   TYPE DESCRIPTION                             VALUE
  0   8        (object header: mark)                   N/A
  8   4        (object header: class)                  N/A
 12   4        (alignment/padding gap)
[...]

PR #1181, opened 2024-09-15
`MOD` constraint failure on mainnet

### LINKED ISSUE #1180: MOD constraint failure on mainnet
## Ranges

```rust
2321470-2321479
```

## Issue

This range breaks the constraint

```rust
mod.set-absolute-values
```

image

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/mod/ModOperation.java
```diff
@@ -127,8 +127,9 @@ private static BaseBytes getRes(OpCode op, Bytes32 arg1, Bytes32 arg2) {
   }
 
   private UInt256 absoluteValueIfSignedInst(Bytes32 arg) {
-    if (isSigned) {
-      return UInt256.valueOf(arg.toBigInteger().abs());
+    if (this.isSigned()) {
+      Bytes argBytes = Bytes.of(arg.toArray());
+      return UInt256.valueOf(argBytes.toBigInteger().abs());
```
OlivierBBB, line 114: This was the culprit. See screenshots here https://github.com/Consensys/linea-tracer/issues/1180#issuecomment-2351784017

letypequividelespoubelles: Why not using, in the case of signed op : `return UInt256.valueOf(arg.toUnsignedBigInteger());` ?

`.toUnsignedBigInteger()` is what we use everywhere

letypequividelespoubelles: LGTM. We have to deal with https://github.com/Consensys/linea-tracer/issues/823 to have less conversions everywhere ...

arithmetization/src/test/java/net/consensys/linea/zktracer/LeosNewRanges.java
letypequividelespoubelles, line 1: what the point of adding so many replayTests if they were all passing ?

letypequividelespoubelles (approved):
LGTM, but https://github.com/Consensys/linea-tracer/issues/823. Too many conversions everywhere in this class ;(

OlivierBBB:
I agree about the conversions. There will have to be a clean up

PR #1188, opened 2024-09-16
fix(abortingCondiftions): handle EIP2681 MAX_NONCE

### LINKED ISSUE #1187: Update AbortingConditions to handle MAX_NONCE

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/AbortingConditions.java
```diff
@@ -30,72 +33,79 @@
...
     }
 
-    this.insufficientBalance =
-        switch (hub.currentFrame().opCode()) {
+    final OpCode currentOp = hub.opCode();
+
+    insufficientBalance =
+        switch (currentOp) {
           case CALL, CALLCODE -> {
-            if (Exceptions.none(hub.pch().exceptions())) {
```
letypequividelespoubelles, line 63: I remove this unnecessary check, as we "prepare" the abortingConditions iff no Exceptions, so this checks was a duplicate

OlivierBBB: I agree in the sense that we don't care about aborting conditions if we have any exceptions. But leaving this check out means we will be doing useless computations at times. Very, very rarely, I agree.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/AbortingConditions.java
```diff
@@ -30,72 +33,79 @@
...
+    insufficientBalance = false;
+    maxNonceReached = false;
   }
 
   public void prepare(Hub hub) {
-    this.callStackOverflow = hub.callStack().wouldOverflow();
+    callStackOverflow = hub.callStack().wouldOverflow();
     if (this.callStackOverflow) {
       return;
     }
```
OlivierBBB, line 60: Are we 100% sure that we don't want the full picture of aborting conditions ? AFAIR returning after detecting 1 aborting condition should be ok. IIRC all aborting conditions are detected in one fell swoop by the OOB module so this should be ok but I don't remember in detail.

OlivierBBB (approved):
LGTM

lorenzogentile404 (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/AbortingConditions.java
letypequividelespoubelles: No, what I mean is that we were having two times the check  that we don't have exceptions :  one before calling the `prepare` and once inside. So now we do it only once.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/AbortingConditions.java
letypequividelespoubelles: As the OOB does the three aborting in the same call, it should be ok ... It was already implemented like this when we had only 2 aborting conditions

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/AbortingConditions.java
OlivierBBB: I see

OlivierBBB:
So many of the tests are failing due to `gas-into-wcp` @lorenzogentile404 
image

PR #1197, opened 2024-09-16
fix capturing the ChainId

letypequividelespoubelles:


letypequividelespoubelles:
#1201 is the right way to do it

PR #1210, opened 2024-09-17
add exception if we try to get chainId from a tx wo chainId (v = 27 o…

### LINKED ISSUE #1209: get the chainId for BlockData

letypequividelespoubelles:
Search for the chainId of the network in all tx of the conflation (instead of previously just the last tx). Will fail  if all tx are wo chainId (will be fixed by https://github.com/hyperledger/besu/issues/7630).

Add exception when trying to retrieve chainId of an unprotected tx (v == 27 or 28)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
```diff
@@ -93,7 +94,20 @@ public List<ColumnHeader> columnsHeaders() {
...
+
+    // TODO: this doesn't work if all transaction of the batch are WO ChainId
+    long chainId = -1;
+    for (RlpTxnOperation tx : rlpTxn.operations().getAll()) {
+      try {
+        chainId = getChainIdFromTransaction(tx.tx());
+      } catch (Exception e) {
+        continue;
+      }
+      if (chainId != -1) {
```
DavePearce: If the call succeeds in the try block, then chainId cannot be -1 right? So i think you can just return.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
```diff
@@ -93,7 +94,20 @@ public List<ColumnHeader> columnsHeaders() {
...
+        break;
+      }
+    }
+
```
DavePearce: Do you need a check after the loop which throws exception if chainId == -1 ?

DavePearce (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: indeed

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: we could, but anyway it would blow in the constraint / when writting the .lt file. But cleaner, I agree, I'll add it.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxn/RlpTxn.java
```diff
@@ -593,7 +593,7 @@ private void handlePhaseBeta(RlpTxnColumnsValue traceValue, Transaction tx, Trac
     checkArgument(bigIntegerToBytes(V).size() <= 8, "V is longer than 8 bytes");
     final boolean betaIsZero =
         V.equals(BigInteger.valueOf(27))
-            || V.equals(BigInteger.valueOf(28)); // beta = ChainId = 0 iff (V == 27 or V == 28)
+            || V.equals(BigInteger.valueOf(28)); // beta = 0 iff (V == 27 or V == 28)
```
OlivierBBB, line 596: My understanding is that if v $\in$ { 27, 28 } then _we cannot draw any conclusions as to the value of_ $\beta$

OlivierBBB: These transactions don't have replay protection and so can be replayed on any chain (that allows for such transactions.)

arithmetization/src/main/java/net/consensys/linea/zktracer/types/TransactionUtils.java
```diff
@@ -24,10 +24,10 @@ public static long getChainIdFromTransaction(final Transaction tx) {
         return tx.getChainId().get().longValueExact();
       }
       case FRONTIER -> {
-        // when ChainId != 0, V = 2 * ChainID + 35 or V = 2 * ChainId + 36;
+        // If chainId is specified, V = 2 * ChainID + 35 or V = 2 * ChainId + 36;
         final long V = tx.getV().longValueExact();
         if (V == 27 || V == 28) {
-          return 0;
+          throw new IllegalArgumentException("ChainId not specified in transaction");
```
OlivierBBB, line 30: What is the purpose of throwing an exception ? To print a message ? I see that we catch it and resume execution.

OlivierBBB: If there is any exception that would be problematic it's if some transaction defines a chainId which does not coincide with that of Linea mainnet / testnet.

OlivierBBB (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxn/RlpTxn.java
letypequividelespoubelles: we have a beta, but then beta is NOT the value of the chainId ...

arithmetization/src/main/java/net/consensys/linea/zktracer/types/TransactionUtils.java
letypequividelespoubelles: we call this functiun in `phaseBeta` of `RLP_TXN` and here we'll be happy to catch exception if needed

letypequividelespoubelles: > it's if some transaction defines a chainId which does not coincide with that of Linea mainnet / testnet. 

Those transactions are supposed to be blocked by the sequencer.

PR #1222, opened 2024-09-17
use CliqueProtocolSchedule instead of MainnetProtocolSchedule to extr…

daniellehrner:
…act the correct coinbase value


letypequividelespoubelles:
It seems it doesn't impact the coinbase address we get in `Block.java` ? Am I missing something ? 

During tracing, this is where we read the coinbase address to then write it where needed in .lt trace. 

Edit: my bad, when running test, It's now not 0x00 but smtg else.

DavePearce (approved):
LGTM

PR #1228, opened 2024-09-17
Alphabetical ordering except when other priorities prevail

### LINKED ISSUE #1227: List (and order) of modules to trace / count
Fix all issues related to the line counting ahead of the upcoming release as per @Filter94 and @jonesho's request.

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -347,7 +345,10 @@ public List<Module> getModulesToTrace() {
                 stp,
                 trm,
                 txnData,
-                wcp))
+                wcp
+                // l2Block
```
letypequividelespoubelles: L2block is a count onlyoduke, shouldn't be traced

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -361,36 +362,36 @@ public List<Module> getModulesToCount() {
...
-                romLex,
                 add,
                 bin,
                 blakeModexpData,
                 blockdata,
                 blockhash,
-                ext,
                 ecData,
+                exp,
+                ext,
```
letypequividelespoubelles, line 370: Can we add GAS too ? It's empty now, but at least coordinator won't have to change it's setting again and again

letypequividelespoubelles (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
OlivierBBB: I added it in comments to have the same list as in `modulesToCount` (same length.) We can remove it.

PR #1231, opened 2024-09-17
fix: for legacy transactions with chainID

### LINKED ISSUE #1214: Incorrect hash on some legacy transactions
Whilst tx hashes are calculated correctly on some legacy transactions, there are cases where they are not.  An example is Example is [block 296519](https://lineascan.build/block/296519) which is tested in `traceTxStartNotTheSameAsTxPrepare()`.

DavePearce:
There was a bug in the code for setting a chainID for legacy transactions in the case that they do have a chainID.  It was a simple logic error, where v was used in a comparison after it had been modified.


letypequividelespoubelles (approved):
LGTM

PR #1235, opened 2024-09-18
[1140] Update column sizes from linea-constraints

gauravahuja:
#1140 

Corresponding linea-constraints changes https://github.com/Consensys/linea-constraints/pull/384


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Trace.java
letypequividelespoubelles, line 1: if you change this file, it means you change the lisp file, and thus must point to the right commit. This change will be discarded if you don't change it

PR #1241, opened 2024-09-18
feat(req-limit): implement request limiting for trace generation and line counting

powerslider:


arithmetization/src/main/java/net/consensys/linea/plugins/rpc/RequestLimiter.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Builder
+  public RequestLimiter(int concurrentRequestsCount) {
+    this.semaphore = new Semaphore(concurrentRequestsCount);
+  }
+
+  public R execute(T request, Function<T, R> processingFunc) {
+    if (!semaphore.tryAcquire()) {
+      throw new PluginRpcEndpointException(
+          RpcErrorType.INVALID_REQUEST,
```
FlorianHuc, line 39: I believe we should use another error type, the request is valid, it's just the server which is busy

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/RequestLimiter.java
```diff
@@ -0,0 +1,50 @@
...
+public class RequestLimiter<T extends PluginRpcRequest, R> {
+
+  private final Semaphore semaphore;
+
+  @Builder
+  public RequestLimiter(int concurrentRequestsCount) {
+    this.semaphore = new Semaphore(concurrentRequestsCount);
+  }
+
+  public R execute(T request, Function<T, R> processingFunc) {
```
FlorianHuc: where is the associated test?

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/Requests.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package net.consensys.linea.plugins.rpc;
+
+import java.security.InvalidParameterException;
+
+import org.hyperledger.besu.plugin.BesuContext;
+import org.hyperledger.besu.plugin.services.TraceService;
+
+public class Requests {
```
FlorianHuc: I find the name missleading.
Maybe we can call it ServiceProvider?

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/Requests.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Initialize the TraceService.
+   *
+   * @return the initialized TraceService.
+   */
+  public static TraceService getTraceService(final BesuContext besuContext) {
```
FlorianHuc: why not making a method that gets the class as parameter to make it more generic ?
Something along the lines :
public static <T> T getTraceService(final BesuContext besuContext, Class<T> className)
?

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/linecounts/GenerateLineCountsV2.java
FlorianHuc, line 1: instead of deleting and adding a new file, it would be good to use the git rename feature to ease review

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/tracegeneration/GenerateConflatedTracesV2.java
```diff
@@ -38,16 +39,27 @@
...
-  private final BesuContext besuContext;
+  private final RequestLimiter<PluginRpcRequest, TraceFile> requestLimiter;
+
   private final Path tracesOutputPath;
   private TraceService traceService;
 
+  public GenerateConflatedTracesV2(
+      final BesuContext besuContext, final TracesEndpointConfiguration endpointConfiguration) {
+    this.traceService = Requests.getTraceService(besuContext);
+    this.requestLimiter =
```
FlorianHuc: this limiter shoul dbe the same as for the line counts.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/RequestLimiter.java
powerslider: I could not find a better error code, so I am open to suggestions.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/Requests.java
powerslider: I don't understand why it should be `ServiceProvider`. The whole idea behind the name was utility functionality related to the requests logic.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/Requests.java
powerslider: Sure I can do that.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/linecounts/GenerateLineCountsV2.java
powerslider: I don't know why it happened this way, since I just renamed the file and the package.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/tracegeneration/GenerateConflatedTracesV2.java
powerslider: There is a slight technical difficulty to do that, so we need to discuss this aspect.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/RequestLimiter.java
```diff
@@ -0,0 +1,57 @@
...
+    if (!semaphore.tryAcquire()) {
+      throw new PluginRpcEndpointException(
+          RpcErrorType.INVALID_REQUEST,
+          "Request still in progress, wait until available permits are available.");
+    }
+
+    try {
+      return processingFunc.apply(request);
+    } catch (Exception ex) {
+      boolean blockNotFound = ex.getMessage().toLowerCase().contains("block not found");
```
lu-pinto: are you 100% sure all exceptions have a message? If `getMesage() == null` this is a potential NPE

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/linecounts/GenerateLineCountsV2.java
lu-pinto: I think when the file is  x% difference is no longer a rename and it's shown as a new file

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/RequestLimiter.java
powerslider: I will add this check. Thanks!

powerslider: Done.

lu-pinto:
LGTM, I cannot ship it yet. Might be because I'm still not a maintainer for Besu

arithmetization/src/main/java/net/consensys/linea/plugins/readiness/TracerReadinessPlugin.java
```diff
@@ -0,0 +1,140 @@
...
+  private boolean isTracerReady() {
+    this.synchronizationService =
+        Optional.ofNullable(synchronizationService)
+            .orElse(BesuServiceProvider.getSynchronizationService(besuContext));
+
+    final RequestLimiter requestLimiter =
+        RequestLimiterDispatcher.getLimiter(
+            RequestLimiterDispatcher.SINGLE_INSTANCE_REQUEST_LIMITER_KEY);
+
+    return !requestLimiter.isNodeAtMaxCapacity() && synchronizationService.isInitialSyncPhaseDone();
```
lu-pinto, line 128: just as a side note to not forget `isInitialSynchPhaseDone` will not give what you are looking after but there's no other option right now.
I open up the `isInSynch` check that if the head of the chain is out of synch here: https://github.com/hyperledger/besu/pull/7665

PR #1249, opened 2024-09-19
perf(WCP): split StackedSet per opcode

### LINKED ISSUE #1245: Split WCP operation per opcode to reduce setSize + compute ctMAx only when needed

letypequividelespoubelles:
This PR :
- split WCP operations set in many different smaller set, one for each WCP operation type. The aim is to make the contains check faster, as we'll be looking in a smaller set, and we don't have to check the equality of the opcode.
- makes the computation of maxCt only if the ModuleOperation is added to the StackedSet of operations


arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -123,7 +157,7 @@ public boolean callGT(final int arg1, final int arg2) {
   }
 
   public boolean callEQ(final Bytes32 arg1, final Bytes32 arg2) {
-    operations.add(new WcpOperation(EQbv, arg1, arg2));
+    leqOperations.add(new WcpOperation(EQbv, arg1, arg2));
```
lu-pinto: should be `eqOperations`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -54,13 +75,17 @@ public String moduleKey() {
 
   @Override
   public void enterTransaction() {
-    OperationSetModule.super.enterTransaction();
+    for (StackedSet<WcpOperation> operationsSet : operations) {
```
ahamlat, line 78: One optimization here is to inline the loop, as we have only few of them, but let's not do premature optimization and see what the profiling shows.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -84,20 +117,21 @@ public void commit(List<MappedByteBuffer> buffers) {
     final Trace trace = new Trace(buffers);
 
     int stamp = 0;
-    for (WcpOperation operation : operations.getAll()) {
-      operation.trace(trace, ++stamp);
+    for (StackedSet<WcpOperation> operationsSet : operations) {
```
ahamlat, line 120: The same as above related to loop inlining.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/AddOperation.java
```diff
@@ -135,6 +132,7 @@ void trace(int stamp, Trace trace) {
 
   @Override
   protected int computeLineCount() {
-    return this.ctMax + 1;
+    ctMax = computeCtMax();
```
lu-pinto, line 135: if `res` is `final` why doing computations all the time when `computeLineCount` is called? 
Is this method only called once during line counting?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -154,11 +188,18 @@ public boolean callLEQ(final Bytes arg1, final Bytes arg2) {
...
+    geqOperations.add(new WcpOperation(GEQbv, arg1, arg2));
     return arg1.compareTo(arg2) >= 0;
   }
 
   public boolean callGEQ(final Bytes arg1, final Bytes arg2) {
     return callGEQ(Bytes32.leftPad(arg1), Bytes32.leftPad(arg2));
   }
+
+  @Override
+  public void traceEndConflation(final WorldView state) {
```
ahamlat, line 200: Why is this needed with this optimization ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/AddOperation.java
lu-pinto: I've seen same case for other `ModuleOperations` too

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
```diff
@@ -21,28 +21,49 @@
...
+  private final StackedSet<WcpOperation> sltOperations = new StackedSet<>();
+  private final StackedSet<WcpOperation> sgtOperations = new StackedSet<>();
+  private final StackedSet<WcpOperation> eqOperations = new StackedSet<>();
+  private final StackedSet<WcpOperation> isZeroOperations = new StackedSet<>();
+
+  /**
+   * For perf, we split the WcpOperations into different StackedSet in order to - have smaller
+   * set,thus faster to check for equality - remove the opcode value when checking the equality
+   */
+  private final List<StackedSet<WcpOperation>> operations =
```
ahamlat, line 57: I'm not sure this list is needed, we can inline all the calls and have bit a of code duplication.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/WcpOperation.java
```diff
@@ -244,6 +243,7 @@ private int maxCt() {
 
   @Override
   protected int computeLineCount() {
-    return this.ctMax + 1;
+    ctMax = computeCtMax();
```
ahamlat, line 246: In this case, we differ this calculation from initialization time to line counting time, correct ? what is the benefit from doing this ?

ahamlat (approved):
LGTM, I have few small comments, questions

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/WcpOperation.java
ahamlat: I think I have the answer to my question : makes the computation of maxCt only if the ModuleOperation is added to the StackedSet of operations

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
letypequividelespoubelles: good catch, thanks !

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/AddOperation.java
letypequividelespoubelles: you mean that we could compute res only when needed ? Yes that's true, modified

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
letypequividelespoubelles: Before (when having only one StackedSet), it was done in the `OperationSetModule` interface

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/Wcp.java
lu-pinto: He already has a method which acts on a switch/case depending on the opcode so I thought it could be automated with a method but `# opcodes != # StackedSets` and there's no direct mapping unfortunately. So unless we can have a loop to automate fetching collections, I think the extra field might be fine as it's an ArrayList and makes the code a lot cleaner.

PR #1251, opened 2024-09-19
test: test CI using smaller runner

gauravahuja:
- Use a smaller 8 core runner vs 32 core
- Remove spotless as a separate workflow and include it as part of the build workflow
  - The separate spotless workflow uncessary instantiated another runner, downloaded the repo and jdk.
- Initialize only the linea-constraints submodule and exclude the ethereum-tests submodule. The ethereum-test submodule is about 1-2GB and only needed for reference tests and thus not required for CI on every commit.


.github/workflows/gradle.yml
```diff
@@ -77,7 +66,10 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
         with:
-          submodules: recursive
+          submodules: false
+
+      - name: Get Submodules
+        run: git submodule update --init --recursive linea-constraints
```
letypequividelespoubelles, line 72: Does this means it will update the submodule to the latest master in every PR ? If yes, it shouldn't be the case, while debugging, we need to point to a specific constraint commit

gauravahuja: My understanding is that it will update the submodule to the version specified in the `.gitmodules` file

powerslider (approved):

PR #1273, opened 2024-09-22
feat(replayTest): parametric

### LINKED ISSUE #1272: feat(replayTests): split ranges into independent parametrised tests

lorenzogentile404:


arithmetization/src/test/java/net/consensys/linea/replaytests/ReplayTests.java
```diff
@@ -188,4 +197,43 @@ void mainnet1339346ContextRevertTwice() {
...
+  }
+
+  private static void add(int start, int end) {
+    for (int i = start; i <= end; i++) {
+      blockNumbers.add(Arguments.of(i));
+    }
+  }
+
+  private static void add(int start) {
+    blockNumbers.add(Arguments.of(start));
```
OlivierBBB: Have you looked at the python script ? It produces a `.csv` file which you can then download in bulk with @gusiri 's `bulk_capture` script.

lorenzogentile404: Yes the python one, but cannot find the `bulk_capture` one. However, since the purpose of this test is having one parametric test per block, I believe it makes sense to download them one by one and have separated files.

OlivierBBB: It's in `arithmetization/src/tests/resources/bulk_capture.sh`

letypequividelespoubelles (changes requested):
Your changes should be in `ReplayTestTools.java`.

letypequividelespoubelles (approved):
LGTM

PR #1283, opened 2024-09-23
Fix for spotless

OlivierBBB:


OlivierBBB:
Already solved on `arith-dev`

PR #1285, opened 2024-09-23
SHF fixes

### LINKED ISSUE #1280: Failing ranges for `SHF` module
## Ranges

```rust
3901959-3902032
4065349-4065420
4436397-443645
4736791-4736859
4981619-4981658
5042462-5042507
```

OlivierBBB:
- [x] `SHB_k_HI/LO` went from 16B to 1B tracing
- [x] debugging range
- [ ] `resultChecking` fails for

```java
  /** this range fails if you use the default resultChecking == true */
  @Test
  void issue1280_range_4065365_4065369() {
    replay(LINEA_MAINNET, "4065365-4065369.json.gz", false);
  }
```


DavePearce:
The result checking is failing because of the issues addressed in #1291.  This is the key error:

```
TRACE - Contract creation error: account has already been created for address 0xfce10048e89baa0bfab8757ab45e2d057bb917ee
```

DavePearce (approved):
LGTM (though keeping in mind I don't follow everything going on here)

Issue #1205, opened 2024-09-16
Advanced storage testing


OlivierBBB:
To properly test storage operations and the new maps one should write **multi-block** tests where
- in some transactions addresses and storage keys are pre-warmed, including
  - addresses and storage keys that the transaction will interact with
  - addresses and storage keys that the transaction won't interact with
  - in relation to deployments:
    - addresses that don't yet exist but may or may not get deployed during the transaction
    - storage keys in accounts that may or may already exist
  - accounts that don't require pre-warming (e.g. SENDER, RECIPIENT / DEPLOYMENT_ADDRESS, PRECOMPILES --- **Note:** London does not prewarm the COINBASE address)
  - accounts where pre-warming makes no sense (e.g. COINBASE)
  - storage keys in accounts whose storage is inaccessible (EOA's, precompiles, `EMPTY` accounts / gibbrerish address accounts)
- storage values are set multiple times, at times going back to the value originally in storage
- storage operations may or may not get undone by a `REVERT`
- storage operations may get performed during deployment
- storage operations may get performed after deployment
- storage slots may get wiped by `SELFDESTRUCT`
- having multiple cycles (in one or more transactions) of
  - `CREATE2` wherein you touch storage
  - `CALL` to the deployed contract and touch storage 
  - `REVERT` the deployment
  - recycle

OlivierBBB:
BTW I would take the opportunity to contact @thedarkjester and @gauravahuja on how to use the new testing framework. We will want to write many tests, bytecode won't be an option, especially when we start mixing `CREATE(2)`'s, `CALL`'s and `SELFDESTRUT`'s, see https://github.com/Consensys/linea-tracer/pull/1052

lorenzogentile404:
This is addressed in this PR https://github.com/Consensys/linea-tracer/pull/1192. We need to merge the tests only.

PR #1291, opened 2024-09-23
fix: BlockCapturer for Created Accounts

### LINKED ISSUE #1289: Fix BlockCapturer Handling of Contract Creation
The `BlockCapturer` is not always handling contract creation correctly.  I believe this is the sequence where it fails:

- Constract is created
- Contract state is modified in some way

Then, the block capturer adds the account being created to the initial set of `accounts` in the conflation, along with storage values for the locations assigned (where each location is assigned `0x0` because it didn't exist before the conflation).  This then results in BESU thinking the contract already exists at creation time, leading to a failure.

DavePearce:
The block capturer had an issue capturing accounts which were created during the execution of the conflation being captured.  The problematic scenario required an account to exist at the beginning of the conflation, but which was then the subject of a CREATE/CREATE2 contract creation.  This is permitted in certain situations, such as when the account only has a balance (but no storage).

To resolve this case, we have to be a little more careful when creating storage snapshots.  Specifically, when an account only partially exists (i.e. has no storage, but only a balance) then we never create storage snapshots.

Finally, this also adds a debugging mode to ReplayExecutionEnvironment to assist debugging the BlockCapturer.


arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/StorageSnapshot.java
```diff
@@ -19,17 +19,36 @@
...
-    return Optional.ofNullable(world.get(address))
-        .map(
-            account ->
-                new StorageSnapshot(
-                    address.toHexString(),
-                    key.toHexString(),
-                    account.getStorageValue(key).toHexString()));
+    // Lookup account that was touched
+    Account account = world.get(address);
+    // Check account *really* exists from a storage perspective.  This is important to distinguish
```
OlivierBBB, line 30: So in the EYP lingo account existence is equivalent to satisfying at least one of the three conditions below:
- nonzero balance
- nonzero nonce
- nonempty code

Storage is notably absent from that definition.

DavePearce: So, the code used in BESU is:

```java
  private static boolean accountExists(final Account account) {
    // The account exists if it has sent a transaction
    // or already has its code initialized.
    return account.getNonce() != 0 || !account.getCode().isEmpty() || !account.isStorageEmpty();
  }
```

If this returns `true` then account creation fails.

arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/StorageSnapshot.java
```diff
@@ -19,17 +19,36 @@
...
+    // for accounts which are created *during* the conflation.  Such accounts may have technically
+    // existed before the conflation (e.g. they had a non-zero balance) but could still have been
+    // "created" during the conflation.  In such case, this snapshot would be simply assigning 0x0
+    // to the given storage locations.  However, we don't want to create a storage snapshot in such
```
OlivierBBB, line 34: Does taking a storage snapshot insert a storage value ?

OlivierBBB: Also, at least according to the current EYP, storage does not matter when it comes to checking the possibility of doing a CREATE2

image

DavePearce: > Also, at least according to the current EYP

Yeah, but at the end of the day, we are not measuring against the EYP.  We are comparing against execution on MAINNET using BESU.  What I'm looking at is what BESU does.  If that deviates from the EYP ... well, we still have to follow BESU.

However, I agree that it does seem wierd and I can follow up with BESU to see why they do it that way (or even it could be a bug).

DavePearce: > Does taking a storage snapshot insert a storage value ?

Yes, in the replay environment it does.  See `ReplayExecutionEnvironment.initWorld()`.  One option might be to drop `StorageSnapshots` which assign `0x0` at the beginning of the conflation.  It has the same effect.  Maybe you prefer this?

OlivierBBB: But does taking a storage snapshot change the world state ? If so that is a massive problem.

OlivierBBB: In any case we do want to define storage snapshots even if the target account has empty storage or doesn't even exist yet. Simply because it's possible in Ethereum and we will want to test that, see https://github.com/Consensys/linea-tracer/issues/1205, but it should be side-effect free

arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/StorageSnapshot.java
```diff
@@ -19,17 +19,36 @@
...
+          new StorageSnapshot(
+              address.toHexString(),
+              key.toHexString(),
+              account.getStorageValue(key).toHexString()));
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  private static boolean accountExists(final Account account) {
```
OlivierBBB, line 48: AFAIU the EYP defines for an account $a$
- `existence` as existence in the state, i.e. $\sigma[a] \neq \varnothing$
- `EMPTY` as having nontrivial nonce, balance or code
- `DEAD` as either not existing or being empty
- prohibited from being deployed at as existing and either having nontrivial nonce, code or storage


image

image

along with [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610)

So maybe a more accurate name would be `accountMayNotBeDeployedAt` / `deploymentIsProhibitedAt` or so.

OlivierBBB (approved):
LGTM though there were parts where I'm not sure what's going on (e.g. the `executeFrom`)

PR #1296, opened 2024-09-24
perf(StackeSet): less contains check to the full set of all operations

### LINKED ISSUE #1294: optimize StackedSet
Following discussion with on @ahamlat @matkt 

The idea is to :
- change the Set of operations since begiining of conflation to a List
- when adding an op to the StackedSet, only check if already in the set of this tx
- just before lineCount, delete duplicate in the thisTransaction set that are in the List of op since beiig of conflati

letypequividelespoubelles:
The idea is to :
- change the Set of operations since begining of conflation to a List
- when adding an operation to the StackedSet, only check if already in the set of this tx
- just before lineCount, delete duplicate in the thisTransaction set that are also in the List of op since begining of conflation


arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -41,6 +41,7 @@ public class ModuleOperationStackedSet<E extends ModuleOperation> extends Stacke
   private static final Logger log = LoggerFactory.getLogger(ModuleOperationStackedSet.class);
   private final CountOnlyOperation lineCounter = new CountOnlyOperation();
   @Getter private boolean conflationFinished = false;
+  private boolean duplicateAlreadyRemoved = false;
```
letypequividelespoubelles: I added this because I think we don't call `lineCount` when the conflation is already created. So we need to know if we already delete duplicates.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -15,8 +15,8 @@
 
 package net.consensys.linea.zktracer.container.stacked;
 
+import java.util.ArrayList;
```
letypequividelespoubelles, line 20: I have no idea if ArrayList is the best, open to suggestions !

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
```diff
@@ -24,18 +26,19 @@
 @Accessors(fluent = true)
 @Getter
 public class StackedSet<E> {
-  private final Set<E> operationsCommitedToTheConflation;
+  private static final int EXPECTED_PROPORTION_OF_DUPLICATE = 1; // TODO: find me !
```
letypequividelespoubelles: not sure it's really needed, lpus it might depend of the module so ...

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
```diff
@@ -56,9 +60,17 @@ public void pop() {
   }
 
   public boolean add(E e) {
-    if (!operationsCommitedToTheConflation().contains(e)) {
-      return operationsInTransaction().add(e);
+    return operationsInTransaction().add(e);
```
letypequividelespoubelles: This is the real change : we just check if it's a new operation prior to this transaction only

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
```diff
@@ -56,9 +60,17 @@ public void pop() {
...
+  void deleteDuplicate() {
+    final List<E> toRemove =
+        new ArrayList<>(operationsInTransaction().size() * EXPECTED_PROPORTION_OF_DUPLICATE);
+    for (E e : operationsInTransaction()) {
+      if (operationsCommitedToTheConflation().contains(e)) {
+        toRemove.add(e);
+      }
     }
-    return false;
+    toRemove.forEach(operationsInTransaction()::remove);
```
letypequividelespoubelles: maybe a classic loop has better perf ?

OlivierBBB: Question: why not remove the duplicate as you encounter it ? Here it  would seem that the cost is, in the worst case, of the order of the product `cs * ts + ts` where  `cs = operationsCommitedToTheConflation().size` and `ts = operationsInTransaction().size()`

OlivierBBB: Question: why not remove the duplicate as you encounter it ? Here it  would seem that the cost is, in the worst case, of the order of the product `cs * ts + ts` where  `cs = operationsCommitedToTheConflation().size` and `ts = operationsInTransaction().size()`

Also, since I don't know how remove works, what would happen if you had e.g. `operationsInTransaction() = [x, x]` ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
```diff
@@ -56,9 +60,17 @@ public void pop() {
   }
 
   public boolean add(E e) {
-    if (!operationsCommitedToTheConflation().contains(e)) {
-      return operationsInTransaction().add(e);
+    return operationsInTransaction().add(e);
+  }
+
+  void deleteDuplicate() {
```
OlivierBBB: The method name should be `deleteDuplicates`

OlivierBBB (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
OlivierBBB: Why not do e.g.
```java
    for (E e : operationsInTransaction()) {
      if (!operationsCommitedToTheConflation().contains(e)) {
        operationsCommittedToTheConflation().add(e);
      }
    }
```

and not remove duplicates, just add new stuff ?

OlivierBBB: blabla (comment for review)

OlivierBBB (commented):
I have left some comments for review

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
letypequividelespoubelles: > and not remove duplicates, just add new stuff ?

Because we don't want to add them to `operationsCommitedToTheConflation` as we might pop this transaction

> Also, since I don't know how remove works, what would happen if you had e.g. operationsInTransaction() = [x, x] ?

`operationsInTransaction` can't have duplicate : it is a set

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
letypequividelespoubelles: done

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
OlivierBBB: > Because we don't want to add them to operationsCommitedToTheConflation as we might pop this transaction

But you are calling `deleteDuplicates` duing `enter`, just before you `addAll`. Am I missing something ?

letypequividelespoubelles: if you look at `ModuleOperationStackedSet` (which is what is used by all stateless module), we call `deleteDuplicates` at the begining of the lineCounting.
And here, we don't know yet if we'll pop or no the transaction.
image

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -38,15 +39,17 @@
...
   @Getter private boolean conflationFinished = false;
+  private boolean duplicateAlreadyRemoved = false; // TODO: add documentation about this
 
   public ModuleOperationStackedSet() {
     super();
   }
 
-  /** Prefer this constructor as we preallocate more needed memory */
+  /** Prefer this constructor as we preallocate more necessary memory */
   public ModuleOperationStackedSet(
```
lu-pinto, line 54: noticed that this is not used at all even though it's mentioned that it is the preferred

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
lu-pinto: Yes, `ArrayList` is the best smart array you can have. Just watch out for resizes because it resizes automatically so picking a good starting size is always good.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
```diff
@@ -24,18 +25,18 @@
 @Accessors(fluent = true)
 @Getter
 public class StackedSet<E> {
-  private final Set<E> operationsCommitedToTheConflation;
+  private final ArrayList<E> operationsCommitedToTheConflation;
```
lu-pinto, line 28: I don't fully understand the use case. I see below in `add` that you are actually searching for `operations` in the `List` but I would have expected that you would be searching for them in a `Set` as in "have I seen this operation before?" 
Looping through an `ArrayList` is expensive and they are not built for that

lu-pinto: Some operations can also be common across transactions within a block. From what I understand of the code is that the cache is particular to a transaction and it's cleared after. I think an LRU cache that is not controlled by when we enter a transaction or not could be an idea, which the size could be controlled by a parameter to fine tune it during testing.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
letypequividelespoubelles: Yes, we need real block values to know which size we need to inialize.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
letypequividelespoubelles: Do agree, for this we need real life data (see my answer to your comment about the prefered constructor)

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedSet.java
letypequividelespoubelles: >I don't fully understand the use case. I see below in add that you are actually searching for operations in the List but I would have expected that you would be searching for them in a Set as in "have I seen this operation before?"
Looping through an ArrayList is expensive and they are not built for that

The idea is that, at the end of the transaction, we have to delete operation in the cache that are already commited (for the line counting). So when we commit them, the commited operation is a set, we will chck again that we don't have duplicate. If it's a list, we don't. @DavePearce  told me it's not efficient to search in a Set than in a ArrayList. 

>  I think an LRU cache that is not controlled by when we enter a transaction or not could be an idea, 

I don't understand what you mean. The cache is for a transaction, and if the transaction is accepted, we commit it. 

> which the size could be controlled by a parameter to fine tune it during testing.

This is the purpose of the prefered constructor, but we need real life data to know which size is needed for each module

lu-pinto: ah yes. Never mind, I understand now why you need a cache local to a transaction.

I was getting ahead and thinking that a block-wide or runtime-wide cache would be good too. But let's think about it at a later stage.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
ahamlat: Which kind of data are you looking for @letypequividelespoubelles ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -134,11 +146,25 @@ public void clear() {
...
+    if (!duplicateAlreadyRemoved) {
+      lineCount();
+    }
     operationsCommitedToTheConflation().addAll(operationsInTransaction());
     operationsInTransaction().clear();
     lineCounter.enter(); // this is not mandatory but it is more consistent
   }
 
+  void deleteDuplicates() {
+    final List<E> toRemove =
```
ahamlat: I would recommend using un iterator on operationsInTransaction() to be able to remove an element while iterating without triggering ConcurrentModificationException. You won't need to store the list of elements to remove, and it will avoid iterating twice.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -134,11 +146,25 @@ public void clear() {
...
+        new ArrayList<>(operationsInTransaction().size() * EXPECTED_PROPORTION_OF_DUPLICATE);
+    for (E e : operationsInTransaction()) {
+      if (operationsCommitedToTheConflation().contains(e)) {
```
ahamlat: This is what we need to assess in terms of performance. If I understand correctly `deleteDuplicates()` is called each time on transaction `n-1` when we enter a new transaction `n` or when the conflation finishes. The difference from the previous implementation is that with this implementation, we already know that the operation is unique for transaction n-1, where with the previous implementation, we can still check duplicates from transaction n-1 multiple times towards the conflated traces.
One other solution that is very close from current implementation (before this PR) is to check on `add` method if the operation is already present in the current transaction (`operationsInTransaction`), if it is the case, we do nothing, otherwise, we need to check if it is present in `operationsCommitedToTheConflation`. If it is there, we don't need to add it the the current transaction (`operationsInTransaction`), otherwise we add it to `operationsInTransaction`.

arithmetization/src/test/java/net/consensys/linea/zktracer/containers/ModuleOperationStackedSetTests.java
```diff
@@ -62,10 +62,52 @@ public void multiplePushPop() {
...
+    Assertions.assertEquals(2, chunks.size());
+  }
+
+  @Test
+  public void lineCount() {
+    ModuleOperationStackedSet<AddOperation> chunks = new ModuleOperationStackedSet<>();
+    chunks.enter();
+    chunks.add(ONE_PLUS_ONE);
+    chunks.add(ONE_PLUS_ONE);
+    Assertions.assertEquals(2, chunks.lineCount());
```
ahamlat, line 104: Why do we have 2 here and not 1 ?

arithmetization/src/test/java/net/consensys/linea/zktracer/module/ext/TestDuplicatedOperations.java
```diff
@@ -1,50 +0,0 @@
...
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-import net.consensys.linea.testing.BytecodeCompiler;
-import net.consensys.linea.testing.BytecodeRunner;
-import net.consensys.linea.zktracer.opcode.OpCode;
-import org.apache.tuweni.bytes.Bytes;
-import org.junit.jupiter.api.Test;
-
-public class TestDuplicatedOperations {
```
ahamlat, line 26: Why deleting this test ? it looks good for me, I think we should have even more like this one.

arithmetization/src/test/java/net/consensys/linea/zktracer/containers/ModuleOperationStackedSetTests.java
ahamlat, line 104: I guess lineCount() on AddOperation returns 2 for each of these operations, basically ctMax = 1., correct ?

ahamlat (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
letypequividelespoubelles: https://github.com/Consensys/linea-tracer/issues/1394

arithmetization/src/test/java/net/consensys/linea/zktracer/containers/ModuleOperationStackedSetTests.java
letypequividelespoubelles: exaclty.

letypequividelespoubelles: We have a duplicate of an operation that takes two rows, so when calling lineCount it should delete the duplicate, and then return the lineCount of the only operation, which is 2.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/ext/TestDuplicatedOperations.java
letypequividelespoubelles: This test is a duplicate of the `stackContainer` tests, it just verifies that the duplicates are killed, which is the purpose of the tests I added, but are more general.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
letypequividelespoubelles: When adding an operation :
- with previous implem, we check if it's in the alredayCommited, then in current tx
- with this PR : we check if it's in the currentTx only. When calling lineCount or commit this Tx, we kill duplicates that are already commited.
- your proposition : we check if it's in current Tx, then in already commited. 

The issue with third proposition is for operation that are recurrent. Imagine a tx that calls 100 times `1 < 16` in `WCP`, and this operation is already commited. So we would check existence 100 times in the big commited set, instead of doing it only once at the end of the transaction. This is a real use case, as, for exemple, the `MMU` calls `WCP` and `EUC` with always small and similar argument. An other option would be to have an intermediate map (as we did for calls to `WCP` coming from `stackOperation` that filters the check to the full set.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
letypequividelespoubelles: addressed https://github.com/Consensys/linea-tracer/pull/1296/commits/e1236fe685eb7daa02123d9aa7915f4ebe9b0359

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
letypequividelespoubelles: cf https://github.com/Consensys/linea-tracer/issues/1394

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
ahamlat: removeIf on a hashset should use the default implementation and it looks good for me 
```
default boolean removeIf(Predicate<? super E> filter) {
        Objects.requireNonNull(filter);
        boolean removed = false;
        final Iterator<E> each = iterator();
        while (each.hasNext()) {
            if (filter.test(each.next())) {
                each.remove();
                removed = true;
            }
        }
        return removed;
    }

OlivierBBB:
@letypequividelespoubelles Should we close this PR ?

letypequividelespoubelles:
I don't think @ahamlat already did tests on real block. Once it's done and if no better perf, yes we'll close it.

OlivierBBB:
Any update on this @ahamlat / @letypequividelespoubelles ?

letypequividelespoubelles:
As we're going to have the limitless prover "soon" I guess we could close this PR. Thought @ahamlat ? Did you check perf on real blocks ?

ahamlat:
Hi @letypequividelespoubelles ,
Yes, we can close this PR. I didn't have time to check on real blocks.

PR #1314, opened 2024-09-25
Homogenize creation add operation

letypequividelespoubelles:


OlivierBBB (commented):
There is also the `OobOperation.java`'s `callToADD` that uses `add.callADD(arg1, arg2)` where
```java
    final EWord arg1 = EWord.of(arg1Hi, arg1Lo);
    final EWord arg2 = EWord.of(arg2Hi, arg2Lo);
```
I.e. we're feeding it `EWord` rather than `Bytes32`

letypequividelespoubelles:
> There is also the `OobOperation.java`'s `callToADD` that uses `add.callADD(arg1, arg2)` where
> 
> ```java
>     final EWord arg1 = EWord.of(arg1Hi, arg1Lo);
>     final EWord arg2 = EWord.of(arg2Hi, arg2Lo);
> ```
> 
> I.e. we're feeding it `EWord` rather than `Bytes32`

This is the only place where we call `ADD` in fact. But when I try to cast the `EWord` in a `Bytes32` intelliJ tells me it won't do anything, as `EWord` is an implementation of `Bytes32` ...

OlivierBBB (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/OobOperation.java
```diff
@@ -480,8 +481,8 @@ private void callToADD(
     checkArgument(arg1Lo.bitLength() / 8 <= 16);
     checkArgument(arg2Hi.bitLength() / 8 <= 16);
     checkArgument(arg2Lo.bitLength() / 8 <= 16);
-    final EWord arg1 = EWord.of(arg1Hi, arg1Lo);
-    final EWord arg2 = EWord.of(arg2Hi, arg2Lo);
+    final Bytes32 arg1 = EWord.of(arg1Hi, arg1Lo);
+    final Bytes32 arg2 = EWord.of(arg2Hi, arg2Lo);
```
OlivierBBB, line 485: That is very strange ^^ but it seems to not raise an issue

PR #1330, opened 2024-09-30
fix(invalidCodePrefix): make it work

### LINKED ISSUE #1326: expected successfull deployment return is a failure

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -663,14 +662,6 @@ public void traceContextExit(MessageFrame frame) {
...
     exitDeploymentFromDeploymentInfoPov(frame);
 
-    // TODO: why only do this at positive depth ?
-    if (frame.getDepth() > 0) {
-
-      final DeploymentExceptions contextExceptions =
-          DeploymentExceptions.fromFrame(this.currentFrame(), frame);
-      this.currentTraceSection().setContextExceptions(contextExceptions);
-    }
-
```
OlivierBBB, line 673: Yes, I don't know why we kept this in for as long as we did 🤔

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
```diff
@@ -205,12 +205,20 @@ private static boolean isOutOfSStore(MessageFrame frame, OpCode opCode) {
...
     }
-
-    final Bytes deployedCode = frame.getOutputData();
-    return !deployedCode.isEmpty() && (deployedCode.get(0) == (byte) EIP_3541_MARKER);
+    final long size = clampedToLong(frame.getStackItem(1));
+    if (size == 0) {
+      return false;
+    }
+    final long offset = clampedToLong(frame.getStackItem(0));
+    if (offset > frame.memoryByteSize()) {
```
OlivierBBB: I think the condition is wrong. E.g. you have
```rust
memoryByteSize = 1
offset = 0 // you need to check
offset = 1 // you don't need to check
```
The condition should be
```java
if (offset >= frame.memoryByteSize()) {
  return false;
}
```

OlivierBBB (changes requested):
LGTM with small potential bug

PR #1336, opened 2024-09-30
fix(SELFDESTRUCT): deloyment number

### LINKED ISSUE #1335: SELFDESTRUCT : Blockchain reference text 612

OlivierBBB:
There were several problems:
- we were updating the deployment number to a fresh new deployment number twice per SELFDESTRUCT
- ~the ROM_LEX module had no idea about these fresh new deployments~ this isn't an issue and was removed
- we were assuming that DEPLOYMENT_NUMBER_NEW = 1 for any deployment transaction, but this can be at fault in the EVM test suite


arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
```diff
@@ -238,9 +239,23 @@ public void callRomLex(final MessageFrame frame) {
                   }
                 });
       }
+
+      default -> throw new RuntimeException(
+          String.format("%s does not trigger the creation of ROM_LEX", getOpCode(frame)));
     }
   }
 
+  public void callRomLexForSelfdestruct(final Address addressWhichWillSelfdestruct) {
```
letypequividelespoubelles: I don't understand this, ROM / ROM_LEX shouldn't store operations with EMPTY bytecode

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/TransactionStack.java
```diff
@@ -62,20 +62,27 @@ public void resetBlock() {
...
     final TransactionProcessingMetadata newTx =
         new TransactionProcessingMetadata(
-            world, tx, block, relativeTransactionNumber, currentAbsNumber);
+            hub, world, tx, block, relativeTransactionNumber, currentAbsNumber);
 
-    this.txs.add(newTx);
+    transactions.add(newTx);
   }
 
+  // @françois what replaces this for message calls to smart contracts ?
```
letypequividelespoubelles: I don't understand your question. Why do you want to store the CFI of the smart contract recipient of the tx ? We compute it in the recipient AccountFragment. For the moment, the only place where we read the CFI of a deployment transaction is in `TXN_DATA`, this is strange ...

letypequividelespoubelles (commented):
Made some clean up + Idu two things, see comments

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
OlivierBBB: Yes, we fixed this.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/TransactionStack.java
OlivierBBB: > Why do you want to store the CFI of the smart contract recipient of the tx ?

Because the traces require it. The question is when/where does it get computed.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -502,20 +502,27 @@ public void traceStartTransaction(final WorldView world, final Transaction tx) {
...
-      if (txStack.current().requiresPrewarming()) {
+      if (transactionProcessingMetadata.requiresPrewarming()) {
         state.setProcessingPhase(TX_WARM);
         new TxPreWarmingMacroSection(world, this);
       }
       state.setProcessingPhase(TX_INIT);
       new TxInitializationSection(this, world);
     }
 
+    // Note: for deployment transactions the deployment number / status were updated during the
```
letypequividelespoubelles, line 521: :)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/TransactionStack.java
letypequividelespoubelles: we do it at `traceStartTx` in the RomLex so this is ok, already done. We can delete the comment, as this method just set the CFI of the initCode in the TransactionProcessingMetadata. It is read only by `TXN_DATA` for the moment.

letypequividelespoubelles (approved):
LGTM

PR #1337, opened 2024-10-01
fix'selfdestruct): immutable map

### LINKED ISSUE #1332: Try to modify  ImmutableCollection of selfdestruct
reference test 157

letypequividelespoubelles:
This PR :
- (fix) make the List of AttemptedSelfDestruct mutable
- refactors the two records out of TransactionProcessingMetadata
- makes some cleans with removing `this.`


lorenzogentile404 (commented):
It is usually a good practice to keep `List` (interface type) instead of `ArrayList` (actual implementation type) when you declare the type. Beyond that, LGTM.

lorenzogentile404 (approved):

PR #1346, opened 2024-10-02
feat: make the trace deterministic

### LINKED ISSUE #1345: Make stateless modules produce deterministic trace

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/AddOperationComparator.java
```diff
@@ -0,0 +1,35 @@
+/*
```
DavePearce, line 1: It seems odd that you've implemented all the `Comparators` separately from the class itself.  You can have e.g. done `class AddOperation extends ModuleOperation implements Comparable<AddOperation>` then added a `compareTo()` method as required for the interface.  Just saves lines I guess --- not a big deal though.

DavePearce: I suppose the reason for doing it this way is to avoid implemented a `Comparator` for all `Operation` classes?  I think you still could do it though by modifying the signature of `OperationSetModule.sortOperations()`.  I'd have to check it worked though --- so not worth bothering!

DavePearce (commented):
## Notes:

- `ExpOperation` needs a `ExpOperationComparator` as `Exp implements OperationSetModule<ExpOperation>`.

- `StpOperation` needs a `StpOperationComparator` as `Stp implements OperationSetModule<StpOperation>`.

- Does `ExpOperation` need the annotation `@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)` ?

- There is some inconsistency, in that e.g. `BlockHash` does things differently.  But perhaps it has to?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/AddOperationComparator.java
DavePearce: Still, one reason for having `AddOperation` implement `Comparable<AddOperation>` is that you can then modify `ModuleOperationStackedSet` to require that its elements must be sortable like so:

```java
public class ModuleOperationStackedSet<E extends ModuleOperation & Comparable<E>>
    extends StackedSet<E> {
```

That just helps reduce the chance of a problem.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ExpOperation.java
```diff
@@ -50,7 +51,9 @@
 
 @Getter
 public class ExpOperation extends ModuleOperation {
```
DavePearce, line 55: Do we need `@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)` here?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ExpOperationComparator.java
```diff
@@ -0,0 +1,68 @@
...
+public class ExpOperationComparator implements Comparator<ExpOperation> {
+  @Override
+  public int compare(ExpOperation op1, ExpOperation op2) {
+    final int instructionComp =
+        Integer.compare(op1.expCall().expInstruction(), op2.expCall().expInstruction());
+    if (instructionComp != 0) {
+      return instructionComp;
+    }
+
+    if (op1.expCall.expInstruction() == EXP_INST_EXPLOG) {
```
DavePearce: This doesn't seem ideal.  I would have thought it makes more sense to add `compareTo` to the `ExpCall` interface and, hence, its implementations (`ExplogExpCall` and `ModexpLogExpCall`).  Otherwise, this is a bit fragile is someone adds a new implementation of `ExpCall` ... they might forget to update this bit.

DavePearce (dismissed):
Otherwise, LGTM

DavePearce (commented):
I think there is still a problem with `ext` module.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/ext/ExtOperationComparator.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+    // Second sort by Arg1
+    final int arg1Comp = op1.arg1().getBytes32().compareTo(op2.arg1().getBytes32());
+    if (arg1Comp != 0) {
+      return arg1Comp;
+    }
+    // Third, sort by Arg2
+    final int arg2Comp = op1.arg2().getBytes32().compareTo(op2.arg2().getBytes32());
+    if (arg2Comp != 0) {
+      return arg1Comp;
```
DavePearce: Typo --- should be `arg2comp`.

DavePearce (changes requested):
Minor typo in `ext` -- found via testing :)

DavePearce:
Yeah, with that final typo fixed ... I think it looks good.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/AddOperationComparator.java
letypequividelespoubelles: > There is some inconsistency, in that e.g. BlockHash does things differently. But perhaps it has to?

made more consistent. For BlockHash we have to sort before commit time, as we have to trigger WCP.

>Does ExpOperation need the annotation @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false) 

added

arithmetization/src/main/java/net/consensys/linea/zktracer/module/ext/ExtOperationComparator.java
letypequividelespoubelles: indeed. TY

arithmetization/src/main/java/net/consensys/linea/zktracer/module/exp/ExpOperationComparator.java
letypequividelespoubelles: Do agree.

DavePearce (approved):
Nice job.  LGTM.

PR #1348, opened 2024-10-02
fix: remove RpcCliOptions from AbstractLineaSharedOptionsPlugin in order not to make it visible for external plugins

powerslider:


arithmetization/src/main/java/net/consensys/linea/plugins/AbstractLineaPrivateOptionsPlugin.java
```diff
@@ -0,0 +1,46 @@
...
+import lombok.extern.slf4j.Slf4j;
+import net.consensys.linea.plugins.rpc.RpcCliOptions;
+import net.consensys.linea.plugins.rpc.RpcConfiguration;
+
+/**
+ * In this class we put CLI options that are shared with other plugins that are only part of the
+ * tracer
+ */
+@Slf4j
+public abstract class AbstractLineaPrivateOptionsPlugin extends AbstractLineaOptionsPlugin {
```
fab-10:
```suggestion
public abstract class AbstractLineaPrivateOptionsPlugin extends AbstractLineaSharedOptionsPlugin {
```

fab-10 (commented):
Looks good, just extend `AbstractLineaSharedOptionsPlugin` so you also have access to configuration shared with other plugins, like the `LineaL1L2BridgeSharedConfiguration`

fab-10 (approved):

PR #1351, opened 2024-10-02
Fix/force flush+opti

FlorianHuc:
add force flush
performance improvement around hash calculation (parralelism) and Romlex
Add new endpoint getConflatedTracesCountersV2 to count lines for multiple blocks


arithmetization/src/main/java/net/consensys/linea/plugins/rpc/countConflation/ConflatedCountTracesV2.java
```diff
@@ -0,0 +1,104 @@
...
+    Validator.validatePluginRpcRequestParams(rawParams);
+
+    TraceRequestParams params =
+        CONVERTER.fromJson(CONVERTER.toJson(rawParams[0]), TraceRequestParams.class);
+
+    params.validate();
+
+    final long fromBlock = params.startBlockNumber();
+    final long toBlock = params.endBlockNumber();
+    final ZkTracer tracer = new ZkTracer();
```
letypequividelespoubelles, line 88: As we're calling the tracer, we must use the constructor where we provide the network chsinId

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
```diff
@@ -62,6 +60,7 @@ public class RomLex
       new ModuleOperationStackedSet<>();
 
   @Getter private List<RomOperation> sortedOperations;
+  Map<ContractMetadata, List<RomOperation>> lookup = new HashMap<>();
```
letypequividelespoubelles: We have a bijection between contract metadata and one RomOperation, not a list of RomOperations

letypequividelespoubelles: The name lookup is not the right one. It's just a dictionary between CFI <=> RomOperation <=> contract Metadata

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
```diff
@@ -83,19 +82,16 @@ public int getCodeFragmentIndexByMetadata(final ContractMetadata metadata) {
...
+              + "\n\tnot found");
     }
-
-    throw new RuntimeException(
-        "RomChunk with:"
-            + String.format("\n\t\taddress = %s", metadata.address())
-            + String.format("\n\t\tdeployment number = %s", metadata.deploymentNumber())
-            + String.format("\n\t\tdeployment status = %s", metadata.underDeployment())
-            + "\n\tnot found");
+    return romOps.size();
```
letypequividelespoubelles: CFI is not the size (as it's not a list) but the index of the given RomOperation in the sortedRomOperations

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
```diff
@@ -286,6 +282,9 @@ public void determineCodeFragmentIndex() {
     sortedOperations = new ArrayList<>(operations.getAll());
     final RomOperationComparator ROM_CHUNK_COMPARATOR = new RomOperationComparator();
     sortedOperations.sort(ROM_CHUNK_COMPARATOR);
+    for (RomOperation romOperation : sortedOperations) {
```
letypequividelespoubelles, line 285: No, we

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraData.java
```diff
@@ -90,4 +90,22 @@ public void commit(List<MappedByteBuffer> buffers) {
       operation.trace(trace, ++stamp);
     }
   }
+
+  @Override
+  public String toString() {
```
letypequividelespoubelles, line 95: What's the purpose ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraDataOperation.java
```diff
@@ -173,4 +173,30 @@ private void traceResult(Trace trace, final int stamp) {
...
+        + ", inputSize="
+        + inputSize
+        + ", lastNBytes="
+        + lastNBytes
+        + ", indexMaxData="
+        + indexMaxData
+        + '}';
+  }
+
+  public static String bytesToHex(byte[] bytes) {
```
letypequividelespoubelles: Better move this in Conversions.java. Maybe it already existed btw

letypequividelespoubelles (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/shakiradata/ShakiraData.java
FlorianHuc: just to make it readible if when we need to log it. We can delete though as it's not logged atm.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
FlorianHuc: this map is just used within this class to speed up the loop on sortedOperations. There is no business logic related to it. 

Instead of iterating through it for all getCodeFragmentIndexByMetadata, we precompute it once and reuse it. 

And actually we can even make it better by keeping just the size of the list.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/countConflation/ConflatedCountTracesV2.java
FlorianHuc: the other endpoints like GenerateLineCountsV2don't do it. Should we update all class instantiating endpoints the same way?

PR #1364, opened 2024-10-03
feat: add validation of the presence of a valid chainId in ZkTracer

### LINKED ISSUE #1363: Add validation of the presence of a valid chainId in ZkTracer

powerslider:


arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -55,12 +55,14 @@
...
 public class ZkTracer implements ConflationAwareOperationTracer {
+  private static final BigInteger DUMMY_CHAIN_ID = Bytes.fromHexString("c0ffee").toBigInteger();
+
   /** The {@link GasCalculator} used in this version of the arithmetization */
   public static final GasCalculator gasCalculator = new LondonGasCalculator();
 
   private static final Map<String, Integer> spillings;
 
-  public final BigInteger chainId;
+  @Getter private final BigInteger chainId;
```
letypequividelespoubelles, line 65: We don't need to keep the ChainId in the ZkTracer, just to give it to the HUB and store it there.

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -178,7 +178,11 @@ public static void executeTest(final BlockchainReferenceTestCaseSpec spec) {
     final MutableBlockchain blockchain = spec.getBlockchain();
     final ProtocolContext context = spec.getProtocolContext();
 
-    final ZkTracer zkTracer = new ZkTracer(schedule.getChainId().get());
```
letypequividelespoubelles, line 181: Why ? It seems all referenceTest have a chainId in the `schedule` so we can get it for sure

powerslider: Because it is an Optional and we need to do the `isPresent` check before calling `get`

PR #1365, opened 2024-10-03
Fix multi block tests

gauravahuja:
After this commit _[use CliqueProtocolSchedule instead of MainnetProtocolSchedule to extract the correct coinbase value](https://github.com/Consensys/linea-tracer/pull/1222)_ the multi block test fail because unlike Replay tests, they don't have the actual block header with clique information.

This change should fix this exception
```
Invalid Bytes supplied - too short to produce a valid Clique Extra Data object.
java.lang.IllegalArgumentException: Invalid Bytes supplied - too short to produce a valid Clique Extra Data object.
	at org.hyperledger.besu.consensus.clique.CliqueExtraData.decodeRaw(CliqueExtraData.java:115)
	at org.hyperledger.besu.consensus.clique.CliqueBlockHeaderFunctions.
```

### DIFF
--- a/arithmetization/src/test/java/net/consensys/linea/zktracer/ExampleMultiBlockTest.java
+++ b/arithmetization/src/test/java/net/consensys/linea/zktracer/ExampleMultiBlockTest.java
@@ -13,10 +13,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package net.consensys.linea.testing;
+package net.consensys.linea.zktracer;
 
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


DavePearce (approved):
Overall, looks good.  Again it might be useful to add some documentation highlighting what the purpose of the new switch is.  Essentially, you are disabling the clique consensus protocol for manual tests.  You could even rename `useCoinbaseAddressFromBlockHeader` to something more descriptive, like `useCliqueConsensus` (and obviously switch the interpretation of `true` / `false`, etc).  Just a thought.

powerslider (approved):

PR #1369, opened 2024-10-03
feat: update Corset to `v9.7.14`

DavePearce:
This updates the version of `corset` used in the CI pipeline to the latest `v9.7.14`, as this includes several important fixes.

### DIFF
--- a/.github/workflows/gradle-ethereum-tests.yml
+++ b/.github/workflows/gradle-ethereum-tests.yml
@@ -32,7 +32,7 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Install Corset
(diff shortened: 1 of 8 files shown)


gauravahuja (approved):

gauravahuja:
Go corset tests are failing, is it because of the updated version?

OlivierBBB (approved):

DavePearce:
> Go corset tests are failing, is it because of the updated version?

@gauravahuja I'm not exactly sure why.  I'm investigating now.  I think the reason is mostly likely due to memory exhaustion from the look of the `gradle` output.  This would imply that something has changed in the `zkevm.bin` file generate using the new version of `corset`.  Specifically, this particular trace is causing more problems.  I had this directive `GOMEM=196Gib` which worked well for the `ubuntu-latest-128` runners, but I'm guessing the `ubuntu-latest-32` runners have significantly less memory.  I think you can do all the traces in less than `64Gib` ... so I'm tweaking the directive.

PR #1376, opened 2024-10-04
Improve StackOperation memory footprint

ahamlat:
We've noticed a huge amount of StackOperation instances when facing memory issues on Besu Full nodes, 14 millions in the output below. This happened when transaction simulation is enabled on p2p. Even if this feature is now disabled, it showed up that the tracer can generate a lot of StackOperation instances.

```
num     #instances         #bytes  class name (module)
-------------------------------------------------------
   1:       3705022    41516103048  [B (java.base@21.0.4)
   2:      14679740      587189600  net.consensys.linea.zktracer.runtime.stack.StackOperation
   3:       3669937      557830424  net.consensys.linea.zktracer.module.hub.fragment.common.CommonFragmentValues
   4:         34137      509379848  [Ljdk.internal.vm.FillerElement; (java.base@21.0.4)
   5:       3669935      352313760  net.consensys.linea.zktracer.module.hub.fragment.StackFragment
   6:       4631665      221481824  [Ljava.lang.Object; (java.base@21.0.4)
   7:       3151140      201672960  net.consensys.linea.zktracer.module.hub.section.StackOnlySection
   8:       3669935      176156880  [Lnet.consensys.linea.zktracer.runtime.stack.StackOperation;
   9:       3669942      146797680  net.consensys.linea.zktracer.runtime.stack.Stack
  10:       3568659      142746360  org.apache.tuweni.bytes.ArrayWrappingBytes
  11:       4268772      136600704  java.util.ArrayList (java.base@21.0.4)
  12:       3670741       88097784  java.util.Arrays$ArrayList (java.base@21.0.4)
  ```
This PR improves StackOperation memory footprint by using a byte to represent the name of the stack opcode (none, push, pop) instead of using an enum, which is a class under the hood. This will reduce the memory used by the action field from 24 bytes to 1 byte

image

### CODE REVIEW

File: arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/Stack.java [hunk 1]
```diff
@@ -26,7 +26,10 @@
 import org.hyperledger.besu.evm.frame.MessageFrame;
 
 public class Stack {
-  public static final int MAX_STACK_SIZE = 1024;
+  public static final short MAX_STACK_SIZE = 1024;
+  public static final byte NONE = 0;
+  public static final byte PUSH = 1;
+  public static final byte POP = 2;
```
lu-pinto, line 32: Looks good for now. But I would even go a step further and introduce these constants in besu.
ATM it's even hard to reason about what opcode `int` belongs to what. So if we ever need to change something it's easy to make a mistake: https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/evm/src/main/java/org/hyperledger/besu/evm/EVM.java#L231-L333

This can also be passed into operation constructors: https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/evm/src/main/java/org/hyperledger/besu/evm/operation/PopOperation.java#L33

I wouldn't be afraid to give it a static `int` as it will be inlined by the compiler it you make it `final`.

lu-pinto (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/Stack.java
ahamlat: Yes, good catch, it should work for the evm as well. In the scope of this PR, the push opcode is more a general representation than a real opcode, as I guess it covers different push opcodes from 0x5f to 0x7f : https://www.evm.codes/?fork=cancun#5f

ahamlat: In the evm, it is more for readability then performance I guess, maybe we can still replace int with byte. I remember also a discussion with Danno on this specific topic. He didn't want to store the opcodes numbers for some reason, especially when using Besu evm in another network, like Hedera.

lu-pinto: >He didn't want to store the opcodes numbers for some reason, especially when using Besu evm in another network, like Hedera.

I can't see why as well. Maybe because variants of PUSH make it awkward? But then again he is already doing it in a way in PUSH https://github.com/hyperledger/besu/blob/main/evm/src/main/java/org/hyperledger/besu/evm/operation/PushOperation.java#L27-L52.
Well not just for readability, now it would come handy if you want to use them and define them statically once.

PR #1383, opened 2024-10-06
Emtpy stack operation now a singleton

### LINKED ISSUE #1382: Split `StackOperations` into variants
`StackOperation` can and should be split in 3 variants:
- `PushStackOperation`
- `PopStackOperation`
- `EmptyStackOperation`

Where the `EmptyStackOperation` variant should be, as suggested by @lu-pinto, a singleton.

OlivierBBB:
Replaces the "empty stack operation" constructor `public StackOperation() { ... }` with a pointer to a singleton `EmptyStackOperation`.


letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackLine.java
```diff
@@ -65,7 +67,7 @@ public StackLine() {
   public List<StackOperation> asStackOperations() {
     StackOperation[] r =
         new StackOperation[] {
-          new StackOperation(), new StackOperation(), new StackOperation(), new StackOperation()
+          EmptyStackOperation, EmptyStackOperation, EmptyStackOperation, EmptyStackOperation
```
ahamlat: We can make it even more clear by keeping the StackOperation by exposing empty() as a factory method like for push and pop, which will always return the static emptyStackOperation.
```
StackOperation emptyStackOperation = StackOperation.empty();
StackOperation[] r = new StackOperation[] {
        emptyStackOperation  , emptyStackOperation, emptyStackOperation, emptyStackOperation};
```

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackOperation.java
```diff
@@ -50,7 +50,11 @@ public final class StackOperation {
    */
   @Getter private final int stackStamp;
 
-  StackOperation() {
+  /** Singleton ``empty stack operation'' object. */
+  public static final StackOperation EmptyStackOperation = new StackOperation();
```
ahamlat: Make EmptyStackOperation lower camel case, i.e : `emptyStackOperation`

ahamlat (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackOperation.java
lu-pinto: actually convention for static constants is `EMPTY_STACK_OPERATION` AFAIK

ahamlat: I agree, but I used to see  lower camel case when it is the instance of a singleton.

ahamlat: You can check Runtime.JAVA or System.java from the JDK.

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackLine.java
OlivierBBB: Should be addressed now.

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackOperation.java
OlivierBBB: I implemented the factory method (AFAICT) with a Lombok getter. Intellij renamed the empty stack item to `EMPTY_STACK_ITEM`. Also went ahead and renamed `StackOperation` to `StackItem` in line with specification.

lu-pinto (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackItem.java
```diff
@@ -50,30 +50,34 @@ public final class StackOperation {
    */
   @Getter private final int stackStamp;
 
-  StackOperation() {
+  /** Singleton ``empty stack operation'' object. */
+  @Getter private static final StackItem EMPTY_STACK_ITEM = new StackItem();
```
ahamlat: I'm not a fan of adding a getter on this one. I would still make it private and return it only in case we call StackItem.empty() (to implement)
```
public static StackItem empty() {
    return EMPTY_STACK_ITEM ;
  }
  ```

File: arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackLine.java [hunk 4]
```diff
@@ -60,14 +60,13 @@ public StackLine() {
...
+   * @return a consolidated 4-elements array of the {@link StackItem} – or no-ops
    */
-  public List<StackOperation> asStackOperations() {
-    StackOperation[] r =
-        new StackOperation[] {
-          new StackOperation(), new StackOperation(), new StackOperation(), new StackOperation()
-        };
-    for (IndexedStackOperation item : this.items) {
+  public List<StackItem> asStackItems() {
+    StackItem emptyStackItem = StackItem.EMPTY_STACK_ITEM();
```
ahamlat: If the above is implemented, this should be changed to 
```
StackItem emptyStackItem = StackItem.empty();
```

ahamlat (approved):
LGTM, only one last comment to improve readability.

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackOperation.java
ahamlat: Having a a getter on a static field is a bit unusual as it is accessible at the class level. In this case, we cannot access it because the private accessor. I would recommend to use a real factory method, keep the instance private and return it only when calling the factory method.

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackItem.java
OlivierBBB: Done.

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/stack/StackLine.java
OlivierBBB: Done.

ahamlat (approved):

PR #1384, opened 2024-10-07
optim + linea_getConflatedTracesCountersV2 

FlorianHuc:
improve performances of RomLex and Bytecodes + add new endpoint with rpc name linea_getConflatedTracesCountersV2 that allows to get the line count for a batch of blocks.


arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
```diff
@@ -62,6 +60,7 @@ public class RomLex
       new ModuleOperationStackedSet<>();
 
   @Getter private List<RomOperation> sortedOperations;
+  Map<ContractMetadata, Integer> precomputedNbOfOperationsPerMetadata = new HashMap<>();
```
letypequividelespoubelles: could be final + I don't understand at all the name, it should be something like `cfiMetadataCorrespondance`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
```diff
@@ -286,6 +282,9 @@ public void determineCodeFragmentIndex() {
     sortedOperations = new ArrayList<>(operations.getAll());
     final RomOperationComparator ROM_CHUNK_COMPARATOR = new RomOperationComparator();
     sortedOperations.sort(ROM_CHUNK_COMPARATOR);
+    for (RomOperation romOperation : sortedOperations) {
```
letypequividelespoubelles: I don't fully understand what it does, why not just an old style loop 
```
 for (int i = 0; i < sortedOperations.size(); i++) {
      precomputedNbOfOperationsPerMetadata.put(sortedOperations.get(i).metadata(), i+1);
    }

letypequividelespoubelles (approved):
LGTM

PR #1390, opened 2024-10-08
Improve StackItem memory footprint

ahamlat:
Improve StackItem memory footprint by changing height field from int to short.
By analyzing https://github.com/Consensys/linea-tracer/pull/1376 in details, we found that because of the padding mechanism to have multiples of 8, the whole memory gain is actually only 24 * 3 bytes = 72 bytes :sweat_smile:

This is because using a reference to Action or having a primitive byte make StackItem use in both cases 32 bytes. If we change the height from int to short (4 bytes to 2 bytes) this will reduce the shallow size from 32 to 24 bytes.
StackOperation with action as a Action (enum)
```
**Memory layout of StackItem:**
lineatracer.StackItem object internals:
OFF  SZ                            TYPE DESCRIPTION                 VALUE
  0   8                                 (object header: mark)       0x0000000000000005 (biasable; age: 0)
  8   4                                 (object header: class)      0x000e6da0
 12   4                             int StackOperation.height       0
 16   4                             int StackOperation.stackStamp   0
 20   4   org.apache.tuweni.bytes.Bytes StackOperation.value        (object)
 24   4              lineatracer.Action StackOperation.action       (object)
 28   4                                 (object alignment gap)      
Instance size: 32 bytes
Space losses: 0 bytes internal + 4 bytes external = 4 bytes total
```

**StackOperation with action as a primitive byte**
```
Memory layout of StackOperation:
lineatracer.StackItem object internals:
OFF  SZ                            TYPE DESCRIPTION                 VALUE
  0   8                                 (object header: mark)       0x0000000000000005 (biasable; age: 0)
  8   4                                 (object header: class)      0x000e6da0
 12   4                             int StackOperation.height       0
 16   4                             int StackOperation.stackStamp   0
 20   1                            byte StackOperation.action       0
 21   3                                 (alignment/padding gap)     
 24   4   org.apache.tuweni.bytes.Bytes StackOperation.value        (object)
 28   4                                 (object alignment gap)      
Instance size: 32 bytes
Space losses: 3 bytes internal + 4 bytes external = 7 bytes total
```
If we make height a short, so basically going from -32,768 to 32,767, StackItem will consume 24 bytes shallow size instead of 32.
```
lineatracer.StackItem object internals:
OFF  SZ                            TYPE DESCRIPTION                 VALUE
  0   8                                 (object header: mark)       0x0000000000000005 (biasable; age: 0)
  8   4                                 (object header: class)      0x000e6da0
 12   4                             int StackOperation.stackStamp   0
 16   2                           short StackOperation.height       0
 18   1                            byte StackOperation.action       0
 19   1                                 (alignment/padding gap)     
 20   4   org.apache.tuweni.bytes.Bytes StackOperation.value        (object)
Instance size: 24 bytes
Space losses: 1 bytes internal + 0 bytes external = 1 bytes total
```

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
@@ -81,8 +81,8 @@ public class CommonFragmentValues {
     // this.contextNumberNew = hub.contextNumberNew(callFrame);
     this.pc = hubProcessingPhase == TX_EXEC ? hub.currentFrame().pc() : 0;
     this.pcNew = computePcNew(hub, pc, noStackException, hub.state.getProcessingPhase() == TX_EXEC);
-    this.height = (short) callFrame.stack().getHeight();
-    this.heightNew = (short) callFrame.stack().getHeightNew();
+    this.height = callFrame.stack().getHeight();
+    this.heightNew = callFrame.stack().getHeightNew();
 
     // TODO: partial solution, will not work in general
     this.gasExpected = hub.expectedGas();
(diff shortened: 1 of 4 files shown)


OlivierBBB (approved):
LGTM

PR #1391, opened 2024-10-08
use listener feature of junit to write to the file once only.

FlorianHuc:


reference-tests/src/main/java/net/consensys/linea/ReferenceTestOutcomeRecorderTool.java
```diff
@@ -35,174 +31,93 @@
...
-    failedTests.add(testName);
-    return failedTests;
-  }
-
-  private static String getCleanedConstraintName(String constraint) {
-    String cleanedConstraintName = constraint;
-    if (cleanedConstraintName.contains(".")) {
-      cleanedConstraintName = cleanedConstraintName.split("\\.")[1];
+  public static Map<String, Set<String>> extractConstraints(String message) {
+    String cleaned = message.substring(message.indexOf("constraints failed:") + "constraints failed:".length()).replaceAll(("\\[[0-9]+m"), " ").replace(']', ' ').trim();
```
bradbown: Unless something else has changed we can't just parse for for "constraints failed:" we also have to check for "failing constraint" too, there are 2 different log messages.

FlorianHuc: this constraint failed is the summary at the end of the file, i did not see failing constraint at this stage, do you have an example ?

bradbown: It was highlighted to me by @OlivierBBB because it was identifying failing constraints when non were being caught by "constraints failed:". I only have an example of an image of a log unfortunately:
image

.github/workflows/reference-blockchain-tests.yml
```diff
@@ -70,18 +70,23 @@ jobs:
...
+
       - name: Run reference blockchain tests
         run: ./gradlew referenceBlockchainTests -x spotlessCheck
         timeout-minutes: 180
         continue-on-error: true
         env:
-          REFERENCE_TESTS_PARALLELISM: 20
-          JAVA_OPTS: -Dorg.gradle.daemon=false
+          REFERENCE_TESTS_PARALLELISM: 8
+          JAVA_OPTS: -Dorg.gradle.daemon=false -Xmx122G
```
gauravahuja: Xmx is not needed as we already set the max heap space in the gradle task. Also 122G  is way too much and leave only 6G from other processes like corset validation. Instead it would be best to reduce the REFERENCE_TESTS_PARALLELISM to 2.

reference-tests/build.gradle
```diff
@@ -56,11 +56,11 @@ tasks.register("generateGeneralStateReferenceTests", RefTestGenerationTask) {
 
 tasks.register('referenceBlockchainTests', Test) {
   description = 'Runs ETH reference blockchain tests.'
-  dependsOn generateBlockchainReferenceTests
+//  dependsOn generateBlockchainReferenceTests
 
   boolean isCiServer = System.getenv().containsKey("CI")
   minHeapSize = isCiServer ? "8g" :"4g"
-  maxHeapSize = isCiServer ? "50g" : "8g"
+  maxHeapSize = "122g"
```
gauravahuja: 122G is way too much leaving only 6G for other processes.

reference-tests/build.gradle
```diff
@@ -69,7 +69,7 @@ tasks.register('referenceBlockchainTests', Test) {
 
 tasks.register('referenceGeneralStateTests', Test) {
   description = 'Runs ETH reference general state tests.'
-  dependsOn generateGeneralStateReferenceTests
+//  dependsOn generateGeneralStateReferenceTests
```
gauravahuja: Why comment this?

testing/src/main/java/net/consensys/linea/testing/ExecutionEnvironment.java
```diff
@@ -67,8 +68,9 @@ public static void checkTracer(
       logger.ifPresent(log -> log.debug("trace written to {}", finalTraceFilePath));
       CorsetValidator.Result corsetValidationResult = corsetValidator.validate(traceFilePath);
       assertThat(corsetValidationResult.isValid())
-          .withFailMessage("Corset validation result: %s", corsetValidationResult.toString())
+          .withFailMessage(CORSET_VALIDATION_RESULT + "%s", corsetValidationResult.corsetOutput())
           .isTrue();
+      return corsetValidationResult.corsetOutput();
```
gauravahuja: This will only be returned when the above assertion is true. When it is true returning this output adds no value.

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -222,9 +240,10 @@ public static void executeTest(final BlockchainReferenceTestCaseSpec spec) {
 
     zkTracer.traceEndConflation(worldState);
 
-    ExecutionEnvironment.checkTracer(zkTracer, CORSET_VALIDATOR, Optional.of(log));
-
-    assertThat(blockchain.getChainHeadHash()).isEqualTo(spec.getLastBlockHash());
+    String result = ExecutionEnvironment.checkTracer(zkTracer, CORSET_VALIDATOR, Optional.of(log));
```
gauravahuja: If corset validation fails the checkTracer method will throw an assertion error and will never return the result of invalid trace validation, so this might not do what you inted.

reference-tests/build.gradle
```diff
@@ -121,4 +121,5 @@ dependencies {
   implementation project(":arithmetization")
   implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml'
   testImplementation project(path: ':testing')
+  implementation 'org.junit.platform:junit-platform-launcher:1.11.2'
```
gauravahuja, line 124: Where is this dependency being used?

reference-tests/src/main/java/net/consensys/linea/BlockchainReferenceTestOutcome.java
```diff
@@ -14,10 +14,17 @@
...
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 
 @JsonPropertyOrder({"failedCounter", "successCounter", "modules"})
 public record BlockchainReferenceTestOutcome(
-    int failedCounter, int successCounter, List<ModuleToConstraints> modulesToConstraints) {}
+    AtomicInteger failedCounter,
+    AtomicInteger successCounter,
+    AtomicInteger disabledCounter,
+    AtomicInteger abortedCounter,
+    ConcurrentMap<String, ConcurrentMap<String, ConcurrentSkipListSet<String>>>
```
gauravahuja: Java records provide a way to model immutable data. You can of course mutate the data inside a record like update a map or increment a counter but that is an anti-pattern.

reference-tests/src/main/java/net/consensys/linea/ReferenceTestOutcomeRecorderTool.java
```diff
@@ -14,195 +14,200 @@
...
+  public static final String JSON_INPUT_FILENAME = "failedBlockchainReferenceTests-input.json";
+  public static final String JSON_OUTPUT_FILENAME = "failedBlockchainReferenceTests.json";
+  public static JsonConverter jsonConverter = JsonConverter.builder().build();
+  public static volatile BlockchainReferenceTestOutcome testOutcomes =
+          new BlockchainReferenceTestOutcome(
+                  new AtomicInteger(0),
+                  new AtomicInteger(0),
+                  new AtomicInteger(0),
+                  new AtomicInteger(0),
+                  new ConcurrentHashMap<>());
```
gauravahuja: Mutating a record is an antipattern. It would be best to have individual counters and a concurrent map. At the end when you want to write test outcome create a BlockchainReferenceTestOutcome with values from counter and the map.

reference-tests/build.gradle
```diff
@@ -56,11 +56,11 @@ tasks.register("generateGeneralStateReferenceTests", RefTestGenerationTask) {
 
 tasks.register('referenceBlockchainTests', Test) {
   description = 'Runs ETH reference blockchain tests.'
-  dependsOn generateBlockchainReferenceTests
+//  dependsOn generateBlockchainReferenceTests
```
gauravahuja: Why comment this line?

gauravahuja (changes requested):

reference-tests/src/main/java/net/consensys/linea/ReferenceTestOutcomeRecorderTool.java
FlorianHuc: added

reference-tests/build.gradle
FlorianHuc: this is for the LauncherSessionListener

gauravahuja (approved):
I think you just need to run spotless as the CI job is failing

PR #1393, opened 2024-10-08
Fixing issues with column sizes

### LINKED ISSUE #1392: Fixing issues with column sizes

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
```diff
@@ -77,6 +77,14 @@ public boolean isHalt() {
     return this.instructionFamily == InstructionFamily.HALT;
   }
 
+  public boolean isCall() {
```
letypequividelespoubelles, line 88: We already have those method somewhere else (based on the opcode, not the opcode family)

OlivierBBB: Yes. But it makes more sense to have them defined on the instruction family level.

OlivierBBB: I moved these methods to the `OpCodeData` level.

letypequividelespoubelles (approved):

PR #1398, opened 2024-10-09
Debug blockchain reference test 455

gauravahuja:
Steps to reproduce the OOM error:
1. Checkout this branch
2. `./gradlew generateBlockchainReferenceTests`
3. In `BlockchainReferenceTest_455.java` set `private static final String[] TEST_CONFIG_FILE_DIR_PATH = { "BlockchainTests/GeneralStateTests/stTimeConsuming/static_Call50000_sha256.json" };`
4. Run BlockchainReferenceTest_455 in the IDE or in terminal `./gradlew reference-tests:referenceBlockchainTests --tests "BlockchainReferenceTest_455"` 

The test case `static_Call50000_sha256_*` in `BlockchainReferenceTest_455` causes the JVM to run out of heap space. Increasing the heap to 24G worked for me.

In this test `ZkTracer.traceContextEnter` is invoked 50000 times (Don't know why), and inside this method  `this.hub.traceContextEnter(frame)` consumes up all the heap.

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
@@ -267,12 +267,15 @@ public class ZkTracer implements ConflationAwareOperationTracer {
       }
     }
   }
+  static int counter = 0;
 
   @Override
   public void traceContextEnter(MessageFrame frame) {
     // We only want to trigger on creation of new contexts, not on re-entry in
     // existing contexts
+    counter += 1;
     if (frame.getState() == MessageFrame.State.NOT_STARTED) {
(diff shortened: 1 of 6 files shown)


ahamlat:
Most of the objects in memory are related to the MMU module, which is coherent with SHA256 test, but I see also that Optional takes non negligible amount of memory (almost 5 GiB), and we should get rid of them. 
Here are some fields that we need change and remove Optional (From the heap dump)

In MmuCall.java

```
  private Optional<Bytes> sourceRamBytes = Optional.empty();
  private Optional<Bytes> targetRamBytes = Optional.empty();
  private Optional<Bytes> exoBytes = Optional.empty();
```
In CallSection.java

```
  public Optional<Address> precompileAddress;
```
in AccountFragment.java
```
  private final Optional<Bytes> addressToTrim;
```

image

PR #1401, opened 2024-10-09
Fixing `underflowException` in `KeccakSection` and same fix for `blockhash` module

### LINKED ISSUE #1399: `UnderflowException` in `KeccakSection`
This line in `hub.section.KeccakSection` is causing an `UnderflowException` in the blockchain tests (e.g. test 99):

```
 public void resolvePostExecution(
      Hub hub, MessageFrame frame, Operation.OperationResult operationResult) {
    final Bytes32 hashResult = Bytes32.leftPad(frame.getStackItem(0)); // <---------
   ...
```

The problem is that at this point, there are no items on the stack.

OlivierBBB:
We also address the `BLOCKHASH` case. Maybe that's undesirable. This "fix" could be
- incorrect given the BLOCKHASH module implementation
- unnecessary or too late: the `BLOCKHASH` module should only be triggered if the underlying instruction is
unexceptional (no `SOX` and no `OOGX`)

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
@@ -31,7 +31,6 @@ import net.consensys.linea.zktracer.container.module.OperationSetModule;
 import net.consensys.linea.zktracer.container.stacked.ModuleOperationStackedSet;
 import net.consensys.linea.zktracer.module.hub.Hub;
 import net.consensys.linea.zktracer.module.hub.defer.PostOpcodeDefer;
-import net.consensys.linea.zktracer.module.hub.signals.Exceptions;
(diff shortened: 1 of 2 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
letypequividelespoubelles, line 1: It's not usefull for BLOCKHASH as we trigger it at preOpCode ony not exceptional : 
image

letypequividelespoubelles (approved):
no need for the modif for BLOCKHASH

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
OlivierBBB: Thanks. I undid the "fix".

PR #1402, opened 2024-10-09
Optimize reference tests run and report

gauravahuja:


.github/workflows/reference-blockchain-tests.yml
```diff
@@ -75,7 +75,6 @@ jobs:
         timeout-minutes: 180
         continue-on-error: true
         env:
-          REFERENCE_TESTS_PARALLELISM: 20
```
gauravahuja, line 78: Some tests consume a lot of heap space so it is not feasible to have tests run in parallel as the sum of heap usage of all concurrent tests will exceed the max heap size.

reference-tests/src/test/resources/templates/BlockchainReferenceTest.java.template
```diff
@@ -49,15 +50,14 @@ public class %%TESTS_NAME%% {
...
-        Arguments.of(params[0], params[1], params[2])
-        );
-    } else {
-      return generateTestParametersForConfig(TEST_CONFIG_FILE_DIR_PATH, FAILED_MODULE, FAILED_CONSTRAINT).stream().map(params ->
-          Arguments.of(params[0], params[1], params[2])
-      );
+Collection<Object[]> testParameters = FAILED_MODULE.isEmpty()
+        ? generateTestParametersForConfig(TEST_CONFIG_FILE_DIR_PATH)
+        : generateTestParametersForConfig(TEST_CONFIG_FILE_DIR_PATH, FAILED_MODULE, FAILED_CONSTRAINT);
+    Collection<Object[]> filteredTestParameters = testParameters.stream().filter(params -> (Boolean) params[2]).toList();
```
gauravahuja, line 56: This ensures that the test configs that were going to be skipped will be filtered here itself resulting in a cleaner test report without 10s of thousands of skipped tests. I still need to ensure if all test configs are filtered there is atleast one dummy test config returned. This is because junit expects atleast 1 set of parameters for parameterized test.

reference-tests/src/test/resources/templates/GeneralStateReferenceTest.java.template
```diff
@@ -39,9 +40,13 @@ public class %%TESTS_NAME%% {
   private static final String[] TEST_CONFIG_FILE_DIR_PATH = new String[] {%%TESTS_FILE%%};
 
   public static Stream<Arguments> getTestParametersForConfig() {
-    return generateTestParametersForConfig(TEST_CONFIG_FILE_DIR_PATH).stream().map(params ->
-        Arguments.of(params[0], params[1], params[2])
-        );
+    Collection<Object[]> testParameters = generateTestParametersForConfig(TEST_CONFIG_FILE_DIR_PATH)
+        .stream().filter(params -> (Boolean) params[2]).toList();
```
gauravahuja, line 44: This ensures that the test configs that were going to be skipped will be filtered here itself resulting in a cleaner test report without 10s of thousands of skipped tests. I still need to ensure if all test configs are filtered there is atleast one dummy test config returned. This is because junit expects atleast 1 set of parameters for parameterized test.

PR #1421, opened 2024-10-16
fix: support regeneration of `GlobalConstants.java` (#952)

### LINKED ISSUE #952: Enable regenerate `GlobalConstant.java`
Since its creation in https://github.com/Consensys/linea-tracer/pull/714 it's not possible to easily regenerate GlobalConstant.java

DavePearce:
This enables regeneration of the `GlobalConstants.java` via the command `gradle constants`.  This requires Corset `v9.7.15` (which at this time has not yet been released).  The solution is to allow for an override to the class name being produced in Corset.  So, it doesn't always have to generate a file `Trace.java`.  Rather you can specify the class name to generate (in this case, `GlobalConstants`).  The relevant gradle option for the `constants` command is updated to do this.

### DIFF
--- a/buildSrc/src/main/groovy/TraceFilesTask.groovy
+++ b/buildSrc/src/main/groovy/TraceFilesTask.groovy
@@ -7,6 +7,10 @@ import org.gradle.api.tasks.Optional
 abstract class TraceFilesTask extends Exec {
 
   @Input
+  @Optional
+  abstract Property<String> getClassName()
+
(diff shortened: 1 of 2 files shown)


DavePearce:
I updated the release `v9.7.15` of Corset to fix the generation issue.  I think it must be caching the tag or something.  I will rerun the job shortly.

gauravahuja (approved):

PR #1422, opened 2024-10-16
Add nonce to ToyTransactionBuilder and Txn result validator for multi block tests

gauravahuja:
This PR:

- Fixes the cases where tests using `ToyExecutionEnvironmentV2` or `MultiBlockExecutionEnvironment` with multiple transactions from the same sender account would fail. These tests fail because because the sender account nonce is updated after each transaction in test framework but the nonce of the following transactions from the same sender are not, which would result in an error like `transaction nonce 5 below sender account nonce 6`.
- Adds ability to use a `TransactionProcessingResultValidator` for `MultiBlockExecutionEnvironment` and `ReplayExecutionEnvironment`.
  - By default the `ReplayExecutionEnvironment` uses an empty `TransactionProcessingResultValidator` that does no validation.
  - By default  `MultiBlockExecutionEnvironment` and `ToyExecutionEnvironmentV2` uses a default `TransactionProcessingResultValidator` that validates if the transaction processing was successful.
  - The TransactionProcessingResultValidator can be overridden as is the case in some tests.

### DIFF
--- a/arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobCallTest.java
+++ b/arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobCallTest.java
@@ -25,6 +25,7 @@ import net.consensys.linea.testing.BytecodeRunner;
 import net.consensys.linea.testing.ToyAccount;
 import net.consensys.linea.testing.ToyExecutionEnvironmentV2;
 import net.consensys.linea.testing.ToyTransaction;
+import net.consensys.linea.testing.TransactionProcessingResultValidator;
 import net.consensys.linea.zktracer.module.hub.Hub;
 import net.consensys.linea.zktracer.module.hub.signals.Exceptions;
 import net.consensys.linea.zktracer.types.EWord;
@@ -199,7 +200,8 @@ public class OobCallTest {
         ToyExecutionEnvironmentV2.builder()
             .accounts(List.of(userAccount, contractCallerAccount, contractCalleeAccount))
             .transaction(tx)
-            .testValidator(x -> {})
+            .transactionProcessingResultValidator(
+                TransactionProcessingResultValidator.EMPTY_VALIDATOR)
             .build();
 
     toyExecutionEnvironmentV2.run();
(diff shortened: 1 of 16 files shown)


bogdanbear (approved):
It looks good to me. I tested it on my storage snippets and did not find any issue.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobCallTest.java
```diff
@@ -251,7 +253,8 @@ private void testRecursiveCalls(EWord iterations, int numberOfOnesInOobEvent1) {
         ToyExecutionEnvironmentV2.builder()
             .accounts(List.of(userAccount, contractCallerAccount))
             .transaction(tx)
-            .testValidator(x -> {})
+            .transactionProcessingResultValidator(
```
DavePearce, line 256: Can't you use a default instead of adding all these invocations?  Or is that you want the default not to be the `EMPTY_VALIDATOR` generally, but in all these cases (for some reason) you don't?

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobCallTest.java
```diff
@@ -199,7 +200,8 @@ private void testCallSendValue(
         ToyExecutionEnvironmentV2.builder()
             .accounts(List.of(userAccount, contractCallerAccount, contractCalleeAccount))
             .transaction(tx)
-            .testValidator(x -> {})
+            .transactionProcessingResultValidator(
+                TransactionProcessingResultValidator.EMPTY_VALIDATOR)
```
DavePearce, line 204: So, is this transaction supposed to fail --- or we just don't care?  If its supposed to fail then be good to make sure it does.

DavePearce (approved):
I think it looks good.  I left a few comments on things I didn't understand.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobCallTest.java
gauravahuja: I just preserve the same behaviour as before. Earlier it was `.testValidator(x -> {})` which is essentially same as empty validator.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobCallTest.java
gauravahuja: I just preserve the same behaviour as before. Earlier it was `.testValidator(x -> {})` which is essentially same as empty validator. 

I am not sure what the test is intended for

PR #1423, opened 2024-10-16
filter tests with invalid transactions.

FlorianHuc:
use ref test info to skip test not meaningfull


reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -77,6 +77,14 @@ public class BlockchainReferenceTestTools {
     if (NETWORKS_TO_RUN.isEmpty()) {
       PARAMS.ignoreAll();
     }
+    // ignore tests that are failing because there is an account with nonce 0 and
+    // non empty code which can't happen in Linea since we are post LONDON only.
+    PARAMS.ignore("InitCollision_d0g0v0_London[London]");
```
DavePearce, line 82: Yes, the three tests you have here are "impossible".  For the DafnyEVM, for example, we also had to mark them as ignored because they cannot arise in any meaningful way.  I don't actually understand what they are for.

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -194,6 +202,9 @@ public static void executeTest(final BlockchainReferenceTestCaseSpec spec) {
     zkTracer.traceStartConflation(spec.getCandidateBlocks().length);
 
     for (var candidateBlock : spec.getCandidateBlocks()) {
+      Assumptions.assumeTrue(
+          candidateBlock.areAllTransactionsValid(),
```
DavePearce, line 206: I don't fully understand this.  Does `areAllTransactionsValid()` indicate only that transactions must be well formed?  Presumably a transaction which reverts because of an out-of-gas condition would not be considered "invalid"?

DavePearce (approved):
There's quite a lot going on here, so its hard for me to fully understand what you are doing.  But, overall, I didn't see anything concerning ... so LGTM!!!

letypequividelespoubelles (approved):
LGTM should modify the `.getAccount`method everywhere we use it (Rolex, sstoreSection, etc ...)

PR #1426, opened 2024-10-16
Fix `GAS_COST` and `GAS_NEXT` in all cases except `CALL` and `CREATE`

### LINKED ISSUE #1425: most instructions

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -733,51 +735,84 @@ public void tracePostExecution(MessageFrame frame, Operation.OperationResult ope
...
-    if ((!memoryExpansionException & outOfGasException) || unexceptional) {
-      currentSection.commonValues.gasCost(gasCost);
-      currentSection.commonValues.gasNext(
-          unexceptional ? currentSection.commonValues.gasActual - gasCost : 0);
-    } else {
-      currentSection.commonValues.gasCost(
-          0); // TODO: fill with correct values --- make sure this works in all cases
-      currentSection.commonValues.gasNext(0);
-    }
-
```
OlivierBBB, line 773: Main point: computing the gas cost and gas next _after_ instruction processing by Besu leads to wrong results: the stack items were already popped, memory was expanded, addresses may or may not have been warmed at that stage ... the pricing will be off.

We do these computations (that belong to the common values) in the `CommonValuesFragment` now. It is therefore triggered with every `TraceSection`.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/ExtCodeCopy.java
```diff
@@ -62,6 +60,6 @@ public long accountAccess() {
 
   @Override
   public long linearPerWord() {
-    return linearCost(GlobalConstants.GAS_CONST_G_COPY, this.bitSize, 32);
+    return linearCost(GlobalConstants.GAS_CONST_G_COPY, this.size, 32);
```
OlivierBBB, line 63: Legitimate bug fix.

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/SimpleGasCalculatorTest.java
OlivierBBB, line 1: The purpose of this test was purely to be able to simply compare gas costs in their old placement (`tracePostExecution`) with the gas cost of Besu.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -186,4 +194,56 @@ static int computePcNew(
...
+      return currentFrame.lastValidGasNext();
+    }
+
+    return currentFrame.frame().getRemainingGas();
+  }
+
+  private long computeGasCost() {
+    return Hub.GAS_PROJECTOR.of(hub.messageFrame(), hub.opCode()).upfrontGasCost();
+  }
+
```
lorenzogentile404, line 219: rename to `computeUpfrontGasCost`?

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
```diff
@@ -116,14 +116,14 @@ public long rawStipend() {
       return 0L;
     } else {
       final long remaining = frame.getRemainingGas() - cost;
-      final long weird = remaining - remaining / 64;
+      final long sixtyThreeSixtyFourths = remaining - remaining / 64;
```
lorenzogentile404: what is the purpose of this? do we have a self-explanatory name for it?

lorenzogentile404 (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
OlivierBBB: It's a little strange, I agree. But we have an issue here with naming:
- `GAS_COST` in the spec typically contains the `upfrontGasCost`
- the "real gas cost" (e.g. for CALL's and CREATE's) is deduced from that and other elements
- things get even more annoying for aborted CALL's, EOA CALL's, precompile CALL's

I don't think we have a good means to solve this. I would suggest keeping `GAS_COST` as a name or renaming it to `GAS_UPFRONT`. We can change this later.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
OlivierBBB: `weird` isn't a good name, is all. And what we are computing here is the `63/64-ths` of the `remaingGas` minus the `upfrontGasCost`.

OlivierBBB: We should, if anything, rename `cost` to `upfrontGasCost`. Will do.

lorenzogentile404 (approved):
LGTM

Issue #1395, opened 2024-10-09
Implement modulesToCount selection from module limits TOML config


powerslider:

gauravahuja:
We realised that we do not need a dynamic list of modulesToCount and instead what to trace and count all modules however have a dynamic list of modules to prove. However, after discussion with the prover team we decided to not have a dynamic list of modules to prove either as It is difficult for the prover would need to recreate the public input, update the L1 verifier SMC, etc. So we do as before, ie manually update the list of module to trace/constraints in the makefile when we want to release on sepolia / mainnet.

PR #1443, opened 2024-10-22
feat: implement modulesToCount selection from module limits TOML config

gauravahuja:
#1395

Copied over from Tsvetan's [PR#1396](https://github.com/Consensys/linea-tracer/pull/1396)

This PR enable the dynamic inclusion/exclusion of modules for the line count RPC response:

- Introduces the plugin-linea-line-counts-modules-to-count-config-file-path plugin CLI parameter to set the path of a module limits TOML file from which in can read all the included module keys and calculate line counts only for those modules.
- When plugin-linea-line-counts-modules-to-count-config-file-path is not provided it returns line counts for all modules that are intended for counting


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -310,50 +302,50 @@ private List<Module> precompileLimitModules() {
...
+    mmio = new Mmio(mmu);
+
+    refTableModules = initRefTableModules();
+    precompileLimitModules = initPrecompileLimitModules();
+    standardModules = initStandardModules();
+    standardModulesToCount = initStandardModulesToCount();
+    standardModulesToTrace = initStandardModulesToTrace();
+    modulesToCountMap = initModulesToCountMap();
+
+    modules = Stream.concat(standardModules.stream(), precompileLimitModules.stream()).toList();
```
letypequividelespoubelles: is modules the modules triggered during execution of the evm? If yes, we don't need the precompileLimitModule in, they are now manually triggered by other modules when needed

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -362,95 +354,160 @@ public List<Module> getModulesToTrace() {
...
+        shf,
+        stp,
+        trm,
+        txnData,
+        wcp,
+        l2Block,
+        l2L1Logs);
+  }
+
+  private List<Module> initStandardModulesToTrace() {
```
letypequividelespoubelles, line 400: maybe, the standardModuleToTrace (and to Count) could be initStandardModule + some few others (just take care if some order are needed, like for mum/mmio at tracing time)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -905,7 +962,7 @@ public final MessageFrame messageFrame() {
   private void handleStack(MessageFrame frame) {
     this.currentFrame()
         .stack()
-        .processInstruction(this, frame, MULTIPLIER___STACK_HEIGHT * (stamp() + 1));
+        .processInstruction(this, frame, MULTIPLIER___STACK_HEIGHT * stamp());
```
letypequividelespoubelles: ?

fab-10 (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
gauravahuja: I am not sure, this was in Tsventan's PR and could have been a merge issue.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
gauravahuja: I was trying to do that but I noticed one issue. `StandardModulesToTrace` does not have `gas` but both `StandardModules` and `StandardModulesToCount` have it. I am not sure is this is correct but it this is the expected case then we are removing a module from `StandardModules` to get `StandardModulesToTrace`.

Another issue that can arise is it would be hard to enforce an order if I do list operations because it will be very difficult to position the extra modules in an already existing list of `StandardModules`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
gauravahuja: I am not sure what you mean in this comment, but adding the `precomileLimitModules` to the list of modules just preserves the [current behaviour ](https://github.com/Consensys/linea-tracer/blob/arith-dev/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java#L452)

letypequividelespoubelles (approved):

letypequividelespoubelles:
so I guess we can close this @gauravahuja @FlorianHuc ?

PR #1448, opened 2024-10-22
test(CALL): zero size, huge return at offset, huge call data offset, huge return data offset

### LINKED ISSUE #1444: Failing Tests for `testMxpRandomAdvanced()`
The test `testMxpRandomAdvanced()` describes a large number of potential tests, of which one is run as determined by a fixed random seed.  Since the random seed is shared between tests in the same file, it means that the order in which this test is run will determine exactly which parameters are used for it.  For example, if a new test is added before `testMxpRandomAdvanced` then this will change the parameters used to generate its test.

It turns out, for some parameters, that `testMxpRandomAdvanced` fails.  Some examples (with their parameters) are given below.

## Example 1

- Random Seed: 166
- `instructionCount` = 2
- `INSTRUCTION_COUNT_INIT` = 3

## Example 2

- Random Seed: 44
- `instructionCount` = 3
- `INSTRUCTION_COUNT_INIT` = 3

## Example 3

- Random Seed: 244
- `instructionCount = 5`
- `INSTRUCTION_COUNT_INIT = 5`

lorenzogentile404:


OlivierBBB:
This, or something related, should be blowing up in the first of the tests, the one about `zeroSizeHugeReturnAtOffset`

https://github.com/Consensys/linea-tracer/blob/cab035d5eef415af7cb0c007bdd569e92aa4ffc5/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/transients/OperationAncillaries.java#L164-L179

⏫ @letypequividelespoubelles @lorenzogentile404 ⏫

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/transients/OperationAncillaries.java
```diff
@@ -164,14 +165,36 @@ public static Bytes initCode(final MessageFrame frame) {
...
-        return MemorySpan.fromStartLength(offset, length);
+        long callDataOffset = Words.clampedToLong(frame.getStackItem(2));
+        long callDataSize = Words.clampedToLong(frame.getStackItem(3));
+        long returnAtOffset = Words.clampedToLong(frame.getStackItem(4));
+        long returnAtCapacity = Words.clampedToLong(frame.getStackItem(5));
+
+        System.out.println("callDataOffset: " + callDataOffset);
+        System.out.println("callDataSize: " + callDataSize);
+        System.out.println("returnAtOffset: " + returnAtOffset);
+        System.out.println("returnDataSize: " + returnAtCapacity);
```
OlivierBBB: Same

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/transients/OperationAncillaries.java
```diff
@@ -164,14 +165,36 @@ public static Bytes initCode(final MessageFrame frame) {
...
+        long returnAtOffset = Words.clampedToLong(frame.getStackItem(5));
+        long returnAtCapacity = Words.clampedToLong(frame.getStackItem(6));
+
+        Preconditions.checkArgument(!(callDataOffset >= Math.pow(2, 32) && callDataSize == 0));
+        Preconditions.checkArgument(!(returnAtOffset >= Math.pow(2, 32) && returnAtCapacity == 0));
+
+        System.out.println("callDataOffset: " + callDataOffset);
+        System.out.println("callDataSize: " + callDataSize);
+        System.out.println("returnAtOffset: " + returnAtOffset);
+        System.out.println("returnAtCapacity: " + returnAtCapacity);
```
OlivierBBB: This doesn't explain the HUB tracing not blowing up 🤔.

OlivierBBB:
@lorenzogentile404  We also need simple unit tests that trigger this behaviour.

lorenzogentile404:
At this point, the purpose of this PR is just adding some tests as described in the title. Fixed to MXP tests have been already merged into arith-dev by @DavePearce.

PR #1450, opened 2024-10-23
Mega "simple exceptions testing" issue

### LINKED ISSUE #1435: Mega "simple exceptions testing" issue
# Main point

We should create tests that trigger, for every single opcode, every single possible exception (including aborting conditions, failure conditions for `CREATE`'s, issues with precompiles.)

# Details

- make tables with one row per opcode and one column per exception and mark a cell iff that opcode can trigger that exception
- write tests that trigger the relevant exception for every such cross

## List of ZKEVM exceptions

```rust
SUX
SOX
MXPX
OOGX
RDCX
JUMPX
STATICX
SSTOREX
ICPX
MAXCSX
OPCX
```

## Accessing `delta` and `alpha`

```java
  public void bla(OpCode opCode) {
    opCode.getData().stackSettings.delta(); // number of items that we pop, ∈ {0, 1, ..., 16 or 17}
    opCode.getData().stackSettings.alpha(); // number of items that we add to the stack; the only thing that matter is whether $\alpha > \delta$ ? Also $\alpha - \delta \leq 1$
  }
```

## `stackUnderflowException`

If an instruction removes $\delta$ many items from the stack there should be $\delta$ many tests where you initially do $0, 1, \dots, \delta - 1$ initial pushes before triggering the opcode.
```
PUSH1 01
PUSH1 02
PUSH1 03
PUSH1 04
SOME_OPCODE
```
You can access $\delta$ through the `OpCodeData`. As suggested by @lorenzo.gentile you can batch these considering the set of all instructions with $\delta = \delta_0$ and go through all **positive** values of $\delta_0$. We can also include the final one (which won't raise an exception.)

```
ArgumentStream1 streamOfDeltas  = List.of(0, 1, ...);
ArgumentStream1 streamOfOpCodes = [just walk through all the opcodes and keep those opcodes that have this delta];
```

## `stackOverflowException`

You just have to do something like
```java
for (i = 0; i < 1024; i++) {
	program.push(i);
}
program.op(OpCode.BLA);
```

## Almost unique instructions

Those two should essentially be already done (though checking it won't hurt):
- RDCX:    can only be triggered by `RETURNDATACOPY` (OOB)
- JUMPX:   can only be triggered by `JUMP` and `JUMPI` (OOB)

Those would be new:
- STATICX: can only be triggered by `CALL`, `SSTORE`, `SELFDESTRUCT`, `LOG's`, `CREATE(2)`, 
- SSTOREX: can only be triggered by `SSTORE`
- ICPX:    can only be triggered by `RETURN` during a deployment
- MAXCSX:  can only be triggered by `RETURN` during a deployment (we may already have stuff in place for this in OOB)
- OPCX:    can only be triggered by `INVALID` and any byte that isn't an opcode.

# Deployment specific exceptions (`MAXCSX` and  `ICPX`)

Wrt `MAXCSX` and  `ICPX` we should test
- small deployment which starts with `0xEF`  (i.e. only `ICPX`)
- large deployment which starts with something else (i.e. only `MAXCSX`)
- large deployment which starts with `0xEF` (i.e. both conditions at once.)

OlivierBBB:


arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/StaticExceptionTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    BytecodeRunner bytecodeRunner = BytecodeRunner.of(program.compile());
+
+    bytecodeRunner.run(List.of(calleeAccount));
+
+    if (value != 0) {
+      assertEquals(
+          STATIC_FAULT,
+          bytecodeRunner.getHub().previousTraceSection(2).commonValues.tracedException());
+    }
```
OlivierBBB, line 75: `CALLCODE` can't trigger the `staticContextViolationException`, only `CALL` may do so.

OlivierBBB: `CALL` is the only `CALL`-type instruction that may raise the `staticContextViolationException`.

lorenzogentile404:
Note that OOGX tests in the "warm" case and some OpCode in the "cold" case are not covered yet and will be covered in a new PR. All other exceptions and OpCodes are covered.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -939,10 +939,18 @@ public OpCode opCode() {
...
-  TraceSection currentTraceSection() {
+  public TraceSection currentTraceSection() {
     return state.currentTxTrace().currentSection();
   }
 
+  public TraceSection previousTraceSection() {
+    return state.currentTxTrace().previousSection();
+  }
+
+  public TraceSection previousTraceSection(int n) {
```
OlivierBBB, line 950: The other analoguous `previousTraceSection` method includes  tests on the size of the trace. Is this not needed here ?

OlivierBBB (commented):
LGTM

lorenzogentile404 (approved):

PR #1456, opened 2024-10-28
Smart contract compilation update and utility class

gauravahuja:
- Disable optimizer for yul compilation
- Add method to SmartContractUtils to get the compile bytecode for solidity and yul


bogdanbear (approved):
I checked the modifications on my tests, and there are no issues. 
Thanks, Gaurav!

PR #1486, opened 2024-11-07
feat: support Variable Report Width for Corset

### LINKED ISSUE #1485: Increase report width for `corset`
The number of rows showed either side for a failing constraint is too little in many cases to facilitate debugging.  Therefore, the goal here is to increase the number to something more reasonable.  Specifically, we want to allow this to be configured via the `CORSET_FLAGS` feature so that we can run different widths in the CI pipeline.

DavePearce:
This updates the `CorsetValidator` to support changing the number of trace rows reported for a failing constraint.  The default is firstly changed to `8` (compared with `2` previously).  Secondly, its possible to override the default via the `CORSET_FLAGS` environment variable. This means we can set a small width for the CI pipeline, where horizontal real estate is more limited.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -84,13 +84,13 @@ jobs:
         run: ./gradlew :arithmetization:test --stacktrace
         env:
           JAVA_OPTS: -Dorg.gradle.daemon=false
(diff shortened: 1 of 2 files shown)


letypequividelespoubelles (approved):
LGTM

PR #1492, opened 2024-11-09
`HUB` debugging continued

### LINKED ISSUE #1447: Calling `TRM` on deployment address during unexceptional, unaborted `CREATE` instructions
This is the new _status quo_
- [spec PR](https://github.com/Consensys/linea-constraints/pull/466)
- [constraints PR](https://github.com/Consensys/linea-constraints/pull/466).

### LINKED ISSUE #1490: HUB debugging continued

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/gas/Gas.java
```diff
@@ -74,7 +74,7 @@ public void commit(List<MappedByteBuffer> buffers) {
   public void resolvePostExecution(
       Hub hub, MessageFrame frame, Operation.OperationResult operationResult) {
     gasParameters.gasActual(BigInteger.valueOf(commonValues.gasActual));
-    gasParameters.gasCost(BigInteger.valueOf(commonValues.gasCost()));
+    gasParameters.gasCost(BigInteger.valueOf(commonValues.gasCostToTrace()));
```
OlivierBBB, line 77: We don't always compute the gas cost (e.g. if an exception occurs that has greater priority, e.g. `SUX`/`SOX`/`STATICX`/`MXPX`/... but we always fill the GAS_COST column. We fill it with _some value_ which we call `gasCostToTrace`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/AccountSnapshot.java
```diff
@@ -67,6 +67,14 @@ public static AccountSnapshot canonical(Hub hub, Address address) {
...
   }
 
+  public static AccountSnapshot canonical(Hub hub, WorldView world, Address address) {
+    return fromArguments(
+        world,
+        address,
+        hub.transients.conflation().deploymentInfo(),
+        isAddressWarm(hub.messageFrame(), address));
+  }
+
```
OlivierBBB, line 77: At times the trace method we are using provides us with a `WorldView`. When that is the case it is often more accurate to use this `world` rather than the maybe deprecated `hub.messageFrame().getWorldUpdater()` that we use in the other `canonical` method. This method is then preferred.

This issue arose e.g. in `traceEndTransaction` and was the reason for why some accounts appeared with invalid balances.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -634,7 +634,7 @@ public void traceContextEnter(MessageFrame frame) {
 
       final long callDataContextNumber = callStack.currentCallFrame().contextNumber();
 
-      currentFrame().rememberGasNextBeforePausing();
+      currentFrame().rememberGasNextBeforePausing(this);
```
OlivierBBB, line 637: We need to get the `gasNext` from the `hub`, likely from the `commonValuesFragment` of the `currentTraceSection`;

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -690,16 +690,17 @@ public void traceContextExit(MessageFrame frame) {
...
+        this.state.setProcessingPhase(TX_FINL);
+        new TxFinalizationSection(this, frame.getWorldUpdater(), false);
       }
     }
 
     defers.resolveUponContextExit(this, this.currentFrame());
     // TODO: verify me please @Olivier
     if (this.currentFrame().opCode() == REVERT || Exceptions.any(pch.exceptions())) {
-      defers.resolvePostRollback(this, frame, this.currentFrame());
+      defers.resolveUponRollback(this, frame, this.currentFrame());
```
OlivierBBB, line 703: Renaming to be consistent with the other methods.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -690,16 +690,17 @@ public void traceContextExit(MessageFrame frame) {
               coinbaseIsWarm,
               txStack.getAccumulativeGasUsedInBlockBeforeTxStart());
 
-      if (state.getProcessingPhase() != TX_SKIP) {
-        state.setProcessingPhase(TX_FINL);
-        new TxFinalizationSection(this, frame.getWorldUpdater());
+      if (state.getProcessingPhase() != TX_SKIP
+          && frame.getState() == MessageFrame.State.COMPLETED_SUCCESS) {
+        this.state.setProcessingPhase(TX_FINL);
+        new TxFinalizationSection(this, frame.getWorldUpdater(), false);
```
OlivierBBB, line 696: There are now 2 places that can trigger the creation of a `TxFinalizationSection`:
- `traceContextExit`
- `tracePostExecution`

the reason being: the root context behaves differently in Besu. But looking at Besu I don't remember why that is. And maybe we want to change that and only trace it during `traceContextExit`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -740,7 +741,7 @@ public void tracePostExecution(MessageFrame frame, Operation.OperationResult ope
      */
     if (isExceptional()) {
       this.currentTraceSection()
-          .addFragments(ContextFragment.executionProvidesEmptyReturnData(this));
+          .exceptionalContextFragment(ContextFragment.executionProvidesEmptyReturnData(this));
```
OlivierBBB, line 744: Subtle change: previously we were adding the exceptional context row during `tracePostExecution`. The issue was with sections that defer their tracing to a later point in time. We did find ourselves adding "undoing" account or storage fragments at e.g. `tracePostRollback` _after_ having added the exceptional context row.

The new approach is thus to simply _log_ that exceptional context row and add it later on when everything has been done in terms of adding rows. Currently that is: at line counting time.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/TransactionStack.java
```diff
@@ -31,6 +33,7 @@ public class TransactionStack implements StackedContainer {
       new ArrayList<>(200); // TODO: write the allocated memory from .toml file
   private int currentAbsNumber;
   private int relativeTransactionNumber;
+  @Setter @Getter public TxInitializationSection initializationSection;
```
OlivierBBB, line 36: The initialization section is required for undoing the initial value transfer during the finalization phase. As such it is only required when the transaction fails.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/TxTrace.java
```diff
@@ -112,6 +111,7 @@ public void commit(Trace hubTrace) {
   public int lineCount() {
     int lineCount = 0;
     for (TraceSection s : trace) {
+      if (s.exceptionalContextFragment != null) s.fragments().add(s.exceptionalContextFragment);
```
OlivierBBB, line 114: We add the `exceptionalContextFragment` at line counting time.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/ContextFragment.java
```diff
@@ -93,7 +93,7 @@ public static ContextFragment executionProvidesEmptyReturnData(final Hub hub) {
 
   public static ContextFragment executionProvidesEmptyReturnData(final Hub hub, int contextNumber) {
     CallStack callStack = hub.callStack();
-    int parentId = callStack.getByContextNumber(contextNumber).callerId();
+    int parentId = callStack.getByContextNumber(contextNumber).parentId();
```
OlivierBBB, line 96: `callerId -> parentId` since this `caller` isn't the right term in case of a deployment and also we constantly use the term "child" so `parent` is appropriate.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragment.java
```diff
@@ -125,24 +125,6 @@ public Trace trace(Trace trace) {
         .counterNsr((short) nonStackRowsCounter);
   }
 
-  private Bytes gasCostToTrace() {
```
OlivierBBB, line 128: This was moved to `CommonFragmentValues`. The computation takes place using only values from there.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -60,7 +61,7 @@ public class CommonFragmentValues {
   public final long gasExpected;
   public final long gasActual;
   @Getter final long gasCost;
-  final long gasNext;
+  @Getter @Setter long gasNext;
```
OlivierBBB, line 64: CALL's and CREATE's require us setting the `gasNext` in steps.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -230,29 +231,78 @@ private long computeGasCostExcludingDeploymentCost() {
...
+      case CREATE -> gasAfterDeductingCost;
+        // TODO: this is only part of the story because of
+        //  1. nonempty init code CREATE's where gas is paid out of pocket
+      case CALL -> gasAfterDeductingCost;
+        // TODO: this is only part of the story because of
+        //  1. aborts with value transfers (immediately reapStipend)
+        //  2. EOA calls with value transfer (immediately reapStipend)
+        //  3. SMC calls: gas paid out of pocket
+        //  4. PRC calls: gas paid out of pocket + special PRC cost + returned gas
+
```
OlivierBBB, line 274: This method will have to be refactored later on. Currently it returns the same value in all cases. What changes is how we later on amend it for CALL's and CREATE's.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/ImcFragment.java
```diff
@@ -20,18 +20,20 @@
...
 import net.consensys.linea.zktracer.module.hub.fragment.imc.oob.OobCall;
+import net.consensys.linea.zktracer.runtime.callstack.CallFrame;
 import net.consensys.linea.zktracer.types.TransactionProcessingMetadata;
 
 /**
  * IMCFragments embed data required for Inter-Module Communication, i.e. data that are required to
  * correctly trigger other modules from the Hub.
  */
-public class ImcFragment implements TraceFragment {
+public class ImcFragment implements TraceFragment, ContextReEntryDefer {
```
OlivierBBB, line 36: We need `ContextReEntryDefer` in order to be able to fill in the following fields:
```rust
CCSR = CHILD_CONTEXT_SELF_REVERTS
CCRS = CHILD_CONTEXT_REVERT_STAMP
```
These values are used by CALL and CREATE to set future undoing row `DOM`/`SUB` stamps related to failed internal transactions.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/ImcFragment.java
```diff
@@ -43,6 +45,8 @@ public class ImcFragment implements TraceFragment {
   private boolean mmuIsSet = false;
   private boolean stpIsSet = false;
 
+  private CallFrame childFrame = null;
```
OlivierBBB, line 48: This pointer to the child latest child context simplifies us extracting the relevant data later on. The main point being: when the child terminates `it knows` whether it will `SELF_REVERT` but it does not necessarily know its `REVERT_STAMP`.

Currently the arithmetization just copies the `REVERT_STAMP` rather than `SELF_REVERT x REVERT_STAMP` which would make more logical sense since we only care about the `REVERT_STAMP` of the child in case it self reverts.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/ImcFragment.java
```diff
@@ -136,6 +140,30 @@ public Trace trace(Trace trace) {
       subFragment.trace(trace, hub.state.stamps());
     }
 
+    if (childFrame != null) {
+      trace.pMiscCcrsStamp(childFrame.revertStamp()).pMiscCcsrFlag(childFrame.selfReverts());
```
OlivierBBB, line 144: Tracing of `CCRS` and `CCSR`. The child context revert information is accessed directly through the child call frame.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/ImcFragment.java
```diff
@@ -136,6 +140,30 @@ public Trace trace(Trace trace) {
...
+    }
+
     return trace;
   }
+
+  // TODO: The most natural thing would be to implement resolveAtContextEntry instead.
+  @Override
+  public void resolveAtContextReEntry(Hub hub, CallFrame frame) {
+    childFrame = hub.callStack().getById(frame.childFramesId().getLast());
```
OlivierBBB, line 153: We log the child call frame at context re-entry. This could alternatively have been done at `immediateContextEntry`. The only reason for doing it at re-entry is that we are fairly certain that at that point the child context will have been properly defined.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/AccountSection.java
```diff
@@ -37,6 +37,8 @@ public class AccountSection extends TraceSection implements PostRollbackDefer {
 
   Bytes rawTargetAddress;
   Address targetAddress;
+  AccountSnapshot accountSnapshotBefore;
+  AccountSnapshot accountSnapshotAfter;
```
OlivierBBB, line 41: We remember these snapshots

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/AccountSection.java
```diff
@@ -94,25 +96,15 @@ public AccountSection(Hub hub) {
...
   }
 
-  public void resolvePostRollback(Hub hub, MessageFrame messageFrame, CallFrame callFrame) {
+  public void resolveUponRollback(Hub hub, MessageFrame messageFrame, CallFrame callFrame) {
 
+    final AccountSnapshot preRollBackAccountSnapshot =
+        accountSnapshotAfter.deepCopy().setDeploymentInfo(hub);
     final AccountSnapshot postRollBackAccountSnapshot =
-        AccountSnapshot.canonical(hub, targetAddress);
-
```
OlivierBBB, line 101: Taking canonical snapshots after the rollback is wrong. That snapshot has _everything_ reverted. We only want to revert the specific action performed in this section.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -195,30 +195,33 @@ public CreateSection(Hub hub) {
           new ShakiraDataOperation(hub.stamp(), create2InitCode);
       hub.shakiraData().call(shakiraDataOperation);
 
-      triggerHashInfo(shakiraDataOperation.result());
+      writeHashInfoResult(shakiraDataOperation.result());
```
OlivierBBB, line 198: Here the distinction is: we may want to _write_ a hash result without triggering HASH_INFO. This is for instance the case when doing
- a KECCAK with zero size or
- an (unexceptional, unaborted) CREATE2 with zero init code size

I change the method to **not** set `triggerHashInfo` anymore but to simply write the HASH to the relevant cells. Thus the name change.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -195,30 +195,33 @@ public CreateSection(Hub hub) {
...
-      if (emptyInitCode) {
-        scenarioFragment.setScenario(CREATE_EMPTY_INIT_CODE_WONT_REVERT);
-        hub.transients().conflation().deploymentInfo().newDeploymentSansExecutionAt(createeAddress);
-        return;
-      }
+    if (emptyInitCode) {
+      finalContextFragment = ContextFragment.nonExecutionProvidesEmptyReturnData(hub);
+      scenarioFragment.setScenario(CREATE_EMPTY_INIT_CODE_WONT_REVERT);
+      hub.transients().conflation().deploymentInfo().newDeploymentSansExecutionAt(createeAddress);
+      return;
```
OlivierBBB, line 214: ras: just splitting the logic.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -195,30 +195,33 @@ public CreateSection(Hub hub) {
...
     }
 
     // Finally, non-exceptional, non-aborting, non-failing, non-emptyInitCode create
+    ////////////////////////////////////////////////////////////////////////////////
+
+    // we capture revert information about the child context: CCSR and CCRS
+    hub.defers().scheduleForContextReEntry(imcFragment, hub.currentFrame());
```
OlivierBBB, line 221: important.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -195,30 +195,33 @@ public CreateSection(Hub hub) {
...
+
+    // The current execution context pays (63/64)ths of it current gas to the child context
+    commonValues.payGasPaidOutOfPocket(hub);
```
OlivierBBB, line 224: We amend the `gasCostToTrace`. See previous discussion.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TraceSection.java
```diff
@@ -190,10 +190,9 @@ private List<TraceFragment> makeStackFragments(final Hub hub, CallFrame currentF
     return stackFragments;
   }
 
-  public void triggerHashInfo(Bytes hash) {
+  public void writeHashInfoResult(Bytes hash) {
     for (TraceFragment fragment : this.fragments()) {
       if (fragment instanceof StackFragment) {
-        ((StackFragment) fragment).hashInfoFlag = true;
```
OlivierBBB, line 196: This was a mistake: there are times where we get the HASH without triggering the HASH_INFO lookup. E.g. whenever we have to compute `KECCAK(( ))`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxFinalizationSection.java
```diff
@@ -75,39 +70,39 @@ public void resolvePostTransaction(
...
     senderSnapshotAfterTxFinalization.turnOnWarmth(); // purely constraints based
 
     final Address recipientAddress = recipientSnapshotBeforeFinalization.address();
-    recipientSnapshotAfterTxFinalization = AccountSnapshot.canonical(hub, recipientAddress);
+    recipientSnapshotAfterTxFinalization = AccountSnapshot.canonical(hub, world, recipientAddress);
     recipientSnapshotAfterTxFinalization.turnOnWarmth(); // purely constraints based
 
     final Address coinbaseAddress = coinbaseSnapshotBeforeTxFinalization.address();
-    coinbaseSnapshotAfterFinalization = AccountSnapshot.canonical(hub, coinbaseAddress);
+    coinbaseSnapshotAfterFinalization = AccountSnapshot.canonical(hub, world, coinbaseAddress);
```
OlivierBBB, line 81: Using the correct world to make snapshots. This world is provided to us in `traceContextExit` (in the unexceptional case.)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxPreWarmingMacroSection.java
```diff
@@ -80,7 +80,13 @@ public TxPreWarmingMacroSection(WorldView world, Hub hub) {
 
                   final DomSubStampsSubFragment domSubStampsSubFragment =
                       new DomSubStampsSubFragment(
-                          DomSubStampsSubFragment.DomSubType.STANDARD, hub.stamp(), 0, 0, 0, 0, 0);
+                          DomSubStampsSubFragment.DomSubType.STANDARD,
+                          hub.stamp() + 1,
```
OlivierBBB, line 84: We have to pre-emptively raise the `hub.stamp()` as this only happens later.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxSkipSection.java
```diff
@@ -124,27 +124,30 @@ public void resolvePostTransaction(Hub hub, WorldView state, Transaction tx, boo
...
                 senderAccountSnapshotAfter,
+                senderAddress,
                 DomSubStampsSubFragment.standardDomSubStamps(hub.stamp(), 0));
 
     // recipient account fragment
     final AccountFragment recipientAccountFragment =
         hub.factories()
             .accountFragment()
-            .make(
+            .makeWithTrm(
```
OlivierBBB, line 137: This is required by the arithmetization, specifically: in the permuted domain the first appearance of an account should be with some trimming.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
```diff
@@ -303,8 +332,8 @@ public void resolveUponContextEntry(Hub hub) {
           // is decremented by the value transferred. This becomes the initial state
           // of the callee, which is then credited by that value. This can happen
           // only for the SMC case.
-          postOpcodeCallerSnapshot.decrementBalanceBy(value);
-          preOpcodeCalleeSnapshot.decrementBalanceBy(value);
+          preOpcodeCalleeSnapshot = postOpcodeCallerSnapshot;
+          postOpcodeCalleeSnapshot = preOpcodeCallerSnapshot;
```
OlivierBBB, line 336: Switcheroo

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
```diff
@@ -385,35 +406,42 @@ public void resolveAtContextReEntry(Hub hub, CallFrame frame) {
...
         scenarioFragment.setScenario(CALL_SMC_FAILURE_WONT_REVERT);
 
         if (selfCallWithNonzeroValueTransfer) {
           childContextExitCallerSnapshot.decrementBalanceBy(value);
           reEntryCalleeSnapshot.decrementBalanceBy(value);
         }
 
+        int childId = hub.currentFrame().childFramesId().getLast();
+        CallFrame childFrame = hub.callStack().getById(childId);
+        int childContextRevertStamp = childFrame.revertStamp();
```
OlivierBBB, line 426: **Note.** This revertStamp is that of a "self-induced" revert. Not the actual revert stamp of the child context. The child call / deployment may be successful (in which case we don't enter here) and be reverted later. Here we only care about children that are failures.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/AccountSnapshot.java
```diff
@@ -234,13 +242,17 @@ public AccountSnapshot setWarmthTo(boolean newWarmth) {
...
-    nonce(nonce + 1);
+    this.nonce(nonce + 1);
     return this;
   }
 
+  public AccountSnapshot setDeploymentInfo(Hub hub) {
+    return this.setDeploymentInfo(hub.transients.conflation().deploymentInfo());
+  }
+
   public AccountSnapshot setDeploymentInfo(DeploymentInfo deploymentInfo) {
```
letypequividelespoubelles, line 253: Why not having a `@Setter` ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
letypequividelespoubelles: you mean we don't traceContextExit the root context in case of exception ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -750,6 +751,11 @@ public void tracePostExecution(MessageFrame frame, Operation.OperationResult ope
     if (!this.currentFrame().opCode().isCall() && !this.currentFrame().opCode().isCreate()) {
       this.unlatchStack(frame, currentSection);
     }
+
+    if (frame.getDepth() == 0 && (isExceptional() || opCode() == REVERT)) {
```
letypequividelespoubelles, line 755: That's strange that we don't traceContextExit in case of exception in the root context ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -753,7 +753,7 @@ public void tracePostExecution(MessageFrame frame, Operation.OperationResult ope
     }
 
     if (frame.getDepth() == 0
-        && (frame.getState() == MessageFrame.State.EXCEPTIONAL_HALT || opCode() == REVERT)) {
+        && (isExceptional() || opCode() == REVERT)) {
```
OlivierBBB: It's unclear exactly when Besu's `MessageFrame` enters the `EXCEPTIONAL_HALT` state. This is particularly true wrt deployment related exceptions (`invalidCodePrefixException` and `maxCodeSizeException`) for whatever reason.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
```diff
@@ -75,12 +75,14 @@ private StackFragment(
     this.exceptions = exceptions;
     this.opCode = stack.getCurrentOpcodeData().mnemonic();
     this.hashInfoFlag =
-        switch (this.opCode) {
-          case SHA3 -> Exceptions.none(exceptions) && gp.messageSize() > 0;
-          case RETURN -> Exceptions.none(exceptions) && gp.messageSize() > 0 && isDeploying;
-          case CREATE2 -> Exceptions.none(exceptions) && aborts.none() && gp.messageSize() > 0;
-          default -> false;
-        };
+        Exceptions.none(exceptions)
```
letypequividelespoubelles: not sure it's an optim, we'll add two more operation (`Exceptions.none` and `gp.messageSize > 0` ) for every stack row ... and it doesn't make the code much easier to read

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
letypequividelespoubelles: can't we get the gasNext from besu (ie from the next section, in `seal()` method ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxPreWarmingMacroSection.java
letypequividelespoubelles: yes true, we raise h only when creating the section ! :) nice

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxSkipSection.java
letypequividelespoubelles: there is a issue for this, did you link it ? 

edit: I did

testing/src/main/java/net/consensys/linea/testing/BytecodeRunner.java
```diff
@@ -107,7 +107,7 @@ public void run(Wei senderBalance, Long gasLimit, List<ToyAccount> additionalAcc
         ToyTransaction.builder()
             .sender(senderAccount)
             .to(receiverAccount)
-            .value(Wei.ONE)
+            .value(Wei.of(69))
```
letypequividelespoubelles, line 110: :)

testing/src/main/java/net/consensys/linea/testing/BytecodeRunner.java
```diff
@@ -41,7 +41,7 @@
  */
 @Accessors(fluent = true)
 public final class BytecodeRunner {
-  public static final long DEFAULT_GAS_LIMIT = 25_000_000L;
+  public static final long DEFAULT_GAS_LIMIT = 61_000_000L;
```
letypequividelespoubelles, line 44: could use the constant from GlabalConsta 'LINEA_BLOCK_GAS_LIMIT`or whatever

letypequividelespoubelles (approved):
LGTM, some minor comments

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
OlivierBBB: I probably phrased that wrong. We certainly don't `traceContextReEnter` as there is nothing to re-enter. TBH I'm not sure it makes sense to have two entry points for `TxFinalizationSection`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
OlivierBBB: I exchanged positions so that these checks are done after the `switch`. I believe that Java is smart enough to ignore conditions in a chain of `&&`'s if a previous condition was false.

testing/src/main/java/net/consensys/linea/testing/BytecodeRunner.java
OlivierBBB: Mostly to disambiguate with other 1's in the trace 🙂.

OlivierBBB:
https://github.com/Consensys/linea-tracer/pull/1492#discussion_r1844924070

No sure. I don't think this would work at the boundaries of execution contexts for instance.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
letypequividelespoubelles: I'd prefer to have only one entry point, yes ...

PR #1510, opened 2024-11-18
fix(txSkip): address collisions

### LINKED ISSUE #1018: `TxSkippedSection` doesn't take address collision shenanigans into account
It seems to me this doesn't take into account the
```java
if (sender == recipient && recipient == coinbase) {
  // full collision
  // ...
  } else if (sender == recipient) {
  // single collision
  // ...
  }  else if (recipient == coinbase) {
  // single collision
  // ...
  } else if (coinbase == sender) {
  // single collision
  // ...
  } else {
  // no collisions
  // ...
}
```
shenanigans.

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxFinalizationSection.java
letypequividelespoubelles, line 1: no real change in this file, just added `final`, moved some function to `TransactionProcessingMetadata`and factorize the last transaction fragment.

testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironmentV2.java
letypequividelespoubelles, line 1: just added a method run(Address coinbase) where we can modify the coinbase address for the test

testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironmentV2.java
```diff
@@ -68,7 +68,20 @@ public class ToyExecutionEnvironmentV2 {
...
+
+    GeneralStateReferenceTestTools.executeTest(
+        generalStateTestCaseEipSpec,
+        protocolSpec,
+        tracer,
+        transactionProcessingResultValidator,
+        zkTracerValidator);
+  }
+
+  public void run(Address coinbase) {
```
gauravahuja: Instead of creating a new run method with the coinbase parameter, you can add coinbase as a variable to ToyExecutionEnvironmentV2 similar to transactions and accounts. By default the coinbase can be the default value or null which can be overriden while building ToyExecutionEnvironmentV2. Something like

```
        .accounts(List.of(senderAccount, receiverAccount, emptyCodeAccount))
        .transaction(tx)
        **.coinbase(<address>)**
        .build()
        .run()

```

letypequividelespoubelles: done

arithmetization/src/test/java/net/consensys/linea/zktracer/module/hub/TxSkipTests.java
```diff
@@ -200,4 +199,120 @@ void test() {
...
+        .accounts(List.of(senderAccount))
+        .transaction(tx)
+        .zkTracerValidator(zkTracer -> {})
+        .coinbase(senderAddress)
+        .build()
+        .run();
+  }
+
+  @Test
+  void senderIsCoinbaseIsReceiver() {
```
OlivierBBB, line 294: One test to rule them all 👑

> “It is a strange fate that we should suffer so much fear and doubt over so small a thing … such a little thing.”

arithmetization/src/test/java/net/consensys/linea/zktracer/module/hub/TxSkipTests.java
```diff
@@ -200,4 +199,120 @@ void test() {
...
+            .value(Wei.of(123))
+            .build();
+
+    ToyExecutionEnvironmentV2.builder()
+        .accounts(List.of(coinbaseAccount, senderAccount))
+        .transaction(tx)
+        .zkTracerValidator(zkTracer -> {})
+        .build()
+        .run();
+  }
```
OlivierBBB, line 233: Does this test succeed ? We had a regression in Besu with @ahamlat / @Gabriel-Trintinalia / @lu-pinto's fix being undone in a recent commit, which may make this test fail now.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxSkipSection.java
```diff
@@ -84,80 +87,119 @@ public TxSkipSection(
...
+          hub.factories()
+              .accountFragment()
+              .makeWithTrm(
+                  thirdAccountFragmentSnapshotBefore,
+                  coinbaseAccountSnapshotAfter,
+                  coinbaseAccountSnapshotBefore.address(),
+                  DomSubStampsSubFragment.standardDomSubStamps(hub.stamp(), 2));
+
+      this.addFragments(firstAccountFragment, secondAccountFragment, thirdAccountFragment);
+    }
```
OlivierBBB, line 197: We will really need account consistency constraints to test this properly 😅.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxSkipSection.java
```diff
@@ -84,80 +87,119 @@ public TxSkipSection(
...
+    if (txMetadata.noAddressCollisions()) {
+
+      final AccountSnapshot senderAccountSnapshotAfter =
+          AccountSnapshot.canonical(hub, world, senderAccountSnapshotBefore.address());
+
+      final AccountSnapshot recipientAccountSnapshotAfter =
+          AccountSnapshot.canonical(hub, world, recipientAccountSnapshotBefore.address());
+
+      final AccountSnapshot coinbaseAccountSnapshotAfter =
+          AccountSnapshot.canonical(hub, world, coinbaseAccountSnapshotBefore.address());
```
OlivierBBB, line 103: I'm noticing that in the old version we were copying over the warmths. I don't have the spec in mind, and I doubt that it fundamentally matters. But the question remains: is the fact that there is no mention of warmths an oversight ?

The spec says that one should keep warmths for `TX_SKIP` as they are. I wonder what that would mean e.g. if
- sender is a precompile (can we even test this ? Or would we need to produce a signature for the (unknown) private key of a precompile) ?
- coinbase is a precompile (those are warm by default)

**N.B.** Recipients may not be precompiles.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxFinalizationSection.java
```diff
@@ -104,105 +104,94 @@ public void resolvePostTransaction(
...
+                      .deepCopy()
+                      .incrementBalanceBy(txMetadata.getGasRefundInWei()),
+                  DomSubStampsSubFragment.standardDomSubStamps(hub.stamp(), 0));
+
+      final AccountFragment coinbaseAccountFragment =
+          hub.factories()
+              .accountFragment()
+              .make(
+                  senderSnapshotBeforeFinalization,
+                  coinbaseSnapshotAfterFinalization,
```
OlivierBBB, line 143: This _should_ indeed work

OlivierBBB (approved):
We need some more testing. In particular the case(s?) I laid out, where the coinbase is a precompile. We may find issues wrt its warmth.

OlivierBBB:
Also if we don't have them yet we should add tests for TX_SKIP with trivial deployments. We can play the same games of
- sender is coinbase
- (deployment address) is coinbase
- coinbase is a precompile

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/TxSkipSection.java
letypequividelespoubelles: I don't think IRL it's possible to have sender / coinbase that is a precompile ...

In the spec, we only constrain the warmth to remains constant, thus it's better to be hot -> hot as it could be warmed (being on the accessList , or so). So when taking teh snapshot, better to turn on the warmth all the time.

PR #1513, opened 2024-11-19
Refactoring to use test watcher for unit tests as well.

### LINKED ISSUE #1563: skip reference test with too big values

FlorianHuc:


reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -105,6 +119,16 @@ public class BlockchainReferenceTestTools {
     // Absurd amount of gas, doesn't run in parallel.
```
OlivierBBB, line 119: I imagine that the CALL50000 could prove problematic with the MemoryRange implementation we currently have @letypequividelespoubelles ^^ If we have to keep 50k snapshots of memory ^^ I mean, it's not an issue. But I wonder whether the tests that @amkCha was able to run again on that 64GB machine would crash with our current tracer.

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -105,6 +119,16 @@ public class BlockchainReferenceTestTools {
...
+    // Balance is more than 128 bits
+    PARAMS.ignore(("Call1024PreCalls_d0g0v0_London[London]"));
+    PARAMS.ignore(("Call1024PreCalls_d0g1v0_London[London]"));
+    PARAMS.ignore(("OverflowGasRequire_London[London]"));
+    PARAMS.ignore("StrangeContractCreation_London[London]");
+    PARAMS.ignore("SuicideIssue_London[London]");
+    PARAMS.ignore("DelegateCallSpam_London[London]");
+    PARAMS.ignore("OverflowGasRequire2_d0g0v0_London[London]");
+    PARAMS.ignore("HighGasLimit_d0g0v0_London[London]");
+
```
OlivierBBB, line 131: OK

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/prc/ecrecover/GasStipendTests.java
OlivierBBB, line 1: Was that file deletion intentional @FlorianHuc ?

testing/src/main/java/net/consensys/linea/reporting/TestOutcomeWriterTool.java
```diff
@@ -0,0 +1,107 @@
...
+  public static void addSuccess() {
+    successCounter.incrementAndGet();
+  }
+
+  public static void addAborted() {
+    abortedCounter.incrementAndGet();
+  }
+
+  public static void addSkipped() {
+    disabledCounter.incrementAndGet();
```
OlivierBBB, line 65: It's strange that we talk about `disabled` and `skipped` rather than having one word for both. Presumably `disabled` makes the most sense if it in any way involves the `@Disabled` annotation (which I have no idea if it does or not)

OlivierBBB (approved):
Modulo the deletion of a test file which I don't know if it was intentional and my limited understanding of other parts, it looks good to me.

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -105,6 +119,16 @@ public class BlockchainReferenceTestTools {
     // Absurd amount of gas, doesn't run in parallel.
     PARAMS.ignore("randomStatetest94_\\w+");
 
+    // Balance is more than 128 bits
+    PARAMS.ignore(("Call1024PreCalls_d0g0v0_London[London]"));
```
letypequividelespoubelles: is it normal to have double bracket @FlorianHuc ?

testing/src/main/java/net/consensys/linea/reporting/TestOutcomeWriterTool.java
FlorianHuc: it comes from the interface i implement. So I guess it's standard.

PR #1515, opened 2024-11-19
test(oogx): CALL different scenarios and gas accumulator

### LINKED ISSUE #1514: `CALL` exceptions testing
- call transfers zero value vs. nonzero value
- target address exists vs. does not exist (in the state)
- target is warm vs. is cold

lorenzogentile404:
Addressing this too here https://github.com/Consensys/linea-tracer/issues/1663.


OlivierBBB:
WRT
https://github.com/Consensys/linea-tracer/blob/41a0c0bbafe98f128d1c2dbc7a9366381e73d1fc/arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java#L162
The method can definitely be simplified (e.g. there is no need for nested `if .. else ..`.) Also it is missing memory expansion costs. Maybe we could use the static `Hub.GAS_PROJECTOR`'s methods ?

lorenzogentile404:
@OlivierBBB using `GasProjector` requires passing the frame, so executing the program. In this case, we need to know the cost before execution. That is the reason why I used that approach. For memory expansion, during execution we end up using a method of `FrontierGasCalculator.class`:

```
  public long memoryExpansionGasCost(MessageFrame frame, long offset, long length) {
    long pre = memoryCost((long)frame.memoryWordSize());
    long post = memoryCost(frame.calculateMemoryExpansion(offset, length));
    return post == Long.MAX_VALUE ? Long.MAX_VALUE : post - pre;
  }
```
that also requires the frame. Let's have a chat about it.

amkCha:
seen with @lorenzogentile404 Extract and merge tests for calls and leave the rest for a separate PR for gas cost in case of memory expansion

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -61,22 +109,150 @@ void outOfGasExceptionColdTest(
...
+        + (isWarm ? GAS_CONST_G_WARM_ACCESS : GAS_CONST_G_COLD_ACCOUNT_ACCESS);
+  }
+
+  /**
+   * We provide a non-zero value in storage so to disambiguate between writing the value in storage
+   * to the stack and writing 0 to the stack.
+   */
+  @ParameterizedTest
+  @ValueSource(ints = {-1, 0, 1})
+  void outOfGasExceptionSStore(int cornerCase) {
```
amkCha: Can be renamed `outOfGasExceptionSLoad`  :ok_hand:

amkCha (approved):

amkCha (approved):

PR #1524, opened 2024-11-19
fix: incorrect generation of `Trace.java` files

### LINKED ISSUE #1523: Incorrect Generation of `Trace.java` for Instruction Decoder
This affects other modules as well.  But, for example, this command:

```
gradle instdecoder
```

Generates an invalid `Trace.java` file.  The problem is that the `package` is incorrect.  Specifically, we have this:

```
package net.consensys.linea.zktracer.module.instructionDecoder;
```
When we should have this:
```
package net.consensys.linea.zktracer.module.tables.instructionDecoder;
```

This looks like it affects any module which uses the `moduleDir` attribute.

@OlivierBBB FYI

DavePearce:
There was an issue with the groovy task `TraceFilesTask.groovy` which was not accounting correctly for modules located in subdirectories.  Specifically, it wasn't generating the correct Java package name.

### DIFF
--- a/buildSrc/src/main/groovy/TraceFilesTask.groovy
+++ b/buildSrc/src/main/groovy/TraceFilesTask.groovy
@@ -26,7 +26,7 @@ abstract class TraceFilesTask extends Exec {
                      "-P", "${moduleDir.getOrElse(module.get()).replaceAll('/','.')}",
                      "-o", "${project.projectDir}/src/main/java/net/consensys/linea/zktracer/module/${moduleDir.getOrElse(module.get())}"
     ]
-    if(className) {
+    if(project.hasProperty("className")) {
       arguments.add("-c")
       arguments.add("${className.get()}")
     }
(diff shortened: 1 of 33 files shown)


letypequividelespoubelles (approved):
LGTM

buildSrc/src/main/groovy/TraceFilesTask.groovy
```diff
@@ -23,10 +23,10 @@ abstract class TraceFilesTask extends Exec {
   @Override
   protected void exec() {
     def arguments = ["besu",
-                     "-P", module.get(),
+                     "-P", "${moduleDir.getOrElse(module.get()).replaceAll('/','.')}",
                      "-o", "${project.projectDir}/src/main/java/net/consensys/linea/zktracer/module/${moduleDir.getOrElse(module.get())}"
     ]
-    if(className) {
+    if(project.hasProperty("className")) {
```
gauravahuja, line 29: I wonder if this would be better
```suggestion
    if(className.isPresent()) {
```

gauravahuja (approved):
lgtm, just the one comment

PR #1531, opened 2024-11-21
fix(Identity): tgt offset for partial return data copy

### LINKED ISSUE #1528: EcMul memory consistency fails
ref test 532

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/mmu/MmuCall.java
```diff
@@ -440,7 +441,7 @@ public static MmuCall forIdentityReturnData(
         .targetId(hub.currentFrame().contextNumber())
         .targetRamBytes(Optional.of(precompileSubsection.callerMemorySnapshot))
         .sourceOffset(EWord.ZERO)
-        .targetOffset(EWord.of(precompileSubsection.parentReturnDataTarget.offset()))
+        .referenceOffset(precompileSubsection.parentReturnDataTarget.offset())
```
letypequividelespoubelles, line 444: The only fix, the rest is using constant instead of values

OlivierBBB: I guess the `targetOffset` at 0 is implicit ? Do we want to add it like for the `sourceOffset` ?

OlivierBBB (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/mmu/MmuCall.java
letypequividelespoubelles: It's not in the hub->MMU interface so it's unspecified, so IMHO it's better to leave it as it is

PR #1542, opened 2024-11-22
fix: many things

### LINKED ISSUE #1534: `HUB`/`MXP` bug: zero size, enormous offsets testing
We recently solved a bug where zero size MXP operations could break stuff in the HUB. We need tests to verify that this worked. The relevant operations that have to be tested are
- [x] CALL's (which produce call data in the callee, and "return at ..." data in the caller.)
  - [x] done for all CALL-type instructions and all [zero/nonzero] size combinations for CALLs [here](https://github.com/Consensys/linea-tracer/pull/1508/files#diff-1e3d6ac035e7875d6476c5ddc273ba764d57e09839e778bd376d3d5c1f0c7282)
  - [x] some partial testing [here](https://github.com/Consensys/linea-tracer/pull/1508/files#diff-eadfce27f5a6859885a75ec49b86394f9b0982aab339485f0b2ca8146c3900e0)
- [x] #1544, solved in PR https://github.com/Consensys/linea-tracer/pull/1545
  - [x] REVERT's both message call context and deployment contexts alike
  - [x] RETURN from message call
  - [x] RETURN from deployment context just to be sure

**Note.** Tests for RETURN / REVERT should distinguish between happening in the root context (as we discard the output data of the root context unless it is deploying byte code) and happening in a non root context (where we don't discard the output data.) There are therefore
```
8 = 2 * 2 * 2 = [RETURN/REVERT] * [message call/deployment] * [root/non root]
```
tests to write.

I don't believe there are anymore though one would have to check.

OlivierBBB:
- the `enter` method of `CallFrame` now uses a `CallDataInfo` object
- `CallSection`'s now produce the relevant `CallDataInfo` object, which the `enter(...)` method of the `CallFrame` re-uses
- a `CallSection`'s `callDataInfo` and `returnAtMemorySpan` now implement the logic whereby if the size is zero we discard the (meaningless) offset and set the span to `EMPTY`
- copy paste of the test in @lorenzogentile404 PR + some fixes to it (it likely be replaced by the analoguous parametric tests on the `HUB` issue)

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
@@ -107,6 +107,7 @@ import net.consensys.linea.zktracer.module.wcp.Wcp;
 import net.consensys.linea.zktracer.opcode.OpCode;
 import net.consensys.linea.zktracer.opcode.OpCodeData;
 import net.consensys.linea.zktracer.opcode.gas.projector.GasProjector;
+import net.consensys.linea.zktracer.runtime.callstack.CallDataInfo;
 import net.consensys.linea.zktracer.runtime.callstack.CallFrame;
 import net.consensys.linea.zktracer.runtime.callstack.CallFrameType;
 import net.consensys.linea.zktracer.runtime.callstack.CallStack;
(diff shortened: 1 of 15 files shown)


letypequividelespoubelles (approved):
I like it. We'll have to refactor a lot of `parentFrame.contextNumber()` we have in lot of place to make it cleaner.

OlivierBBB:
This was merged into the big HUB branch. I'm not sure what to do about this PR now 🤔

OlivierBBB:
I'm closing it as this was already merged into the current HUB branch.

PR #1543, opened 2024-11-22
fix(RAM): fix nightly tests

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/precompiles/ModexpMetadata.java
```diff
@@ -37,14 +35,14 @@ public class ModexpMetadata {
...
 
-  public ModexpMetadata(final Hub hub, final Bytes callData) {
+  public ModexpMetadata(final Bytes callData) {
     this.callData = callData;
-    int exponentOffsetInCallData = BASE_MIN_OFFSET + bbsInt();
-    this.rawLeadingWord =
-        EWord.of(hub.messageFrame().shadowReadMemory(exponentOffsetInCallData, WORD_SIZE));
+  }
+
+  public EWord rawLeadingWord() {
```
letypequividelespoubelles, line 44: the only fix, we were using the offset inside the callData to search in the callerMemory instead ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/Util.java
```diff
@@ -174,7 +174,7 @@ public static short getTxTypeAsInt(TransactionType txType) {
    * @param data, right-padded with 0's if needed, starting from
    * @param positionStart
    */
-  public static Bytes slice(Bytes data, int positionStart, int size) {
+  public static Bytes rightPadSlice(Bytes data, int positionStart, int size) {
```
OlivierBBB, line 177: I think the logic of what follows is off. First of all `positionStart` should be renamed to `offset`. In any case the logic should be

```java
    if (offset < dataSize) {  // this is where the change is 
      if (offset + size <= dataSize) {
        output = data.slice(offset, size);
      } else {
        final int nbPresentBytes = dataSize - offset;
        output = rightPadTo(data.slice(offset, nbPresentBytes), size);
      }
    }
```

The main point is that `offset = dataSize` is already out of bounds of `data`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/precompiles/ModexpMetadata.java
OlivierBBB: The HUB was not being used anywhere ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/precompiles/ModexpMetadata.java
```diff
@@ -37,14 +35,14 @@ public class ModexpMetadata {
...
+    return EWord.of(rightPadSlice(callData, BASE_MIN_OFFSET + bbsInt(), WORD_SIZE));
```
OlivierBBB, line 45: LGTM

OlivierBBB: Why remove the check that was present ?
```java
checkArgument(loadRawLeadingWord());
```
Did it ever blow up ?

OlivierBBB (changes requested):
Please modify `rightPadSlice`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/precompiles/ModexpMetadata.java
letypequividelespoubelles: the function was never used.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/precompiles/ModexpMetadata.java
letypequividelespoubelles: nope. We needed the hub to read the callerMemory, but useless as we pass the callData in the constructor

letypequividelespoubelles:
changes has been merged in https://github.com/Consensys/linea-tracer/pull/1548 to avoid conflicts

PR #1548, opened 2024-11-24
Unifying ranges

### LINKED ISSUE #1547: Unifying ranges
We have currently
- `MemorySpan` (which is also a misnomer) which ought to be called `Range`
- `CallDataInfo`
- and we don't use a singular structure for either "return data" nor for "return at ..."

I propose to refactor this to have
- `Range`'s that are just an `offset` and a `size`
- `MemoryRange`'s that are a `Range`, a `long contextNumber` it and a `Bytes data`, maybe as an abstract class
- `CallDataRange` `ReturnDataRange`, `OutputDataRange` and `ReturnAtRange` that extend it

Or something like that.

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/Util.java
```diff
@@ -168,25 +168,45 @@ public static short getTxTypeAsInt(TransactionType txType) {
...
+
+    checkArgument(offset >= 0, "Offset must be non-negative");
+    checkArgument(size >= 0, "Size must be non-negative");
 
     final int dataSize = data.size();
-    Bytes output = Bytes.repeat((byte) 0x0, size);
-
-    if (dataSize >= positionStart) {
-      if (dataSize >= (positionStart + size)) {
-        output = data.slice(positionStart, size);
```
OlivierBBB, line 184: The condition `dataSize >= positionStart` is problematic as `Bytes slice(int i, int length);` throws an exception if `i >= length`. Also some simple reformatting.

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
OlivierBBB, line 1: The main change in this PR. The goal is to replace all ad-hoc ranges with a `MemoryRange`.

arithmetization/src/main/java/net/consensys/linea/zktracer/types/Range.java
OlivierBBB, line 1: ras: renaming from `MemorySpan -> Range`. Two reasons:
- a **range** is a minimum value and a maximum value while a **span** is the difference between the two
- I want to reserve the word `MemoryRange` to something that actually points into _someone's_ memory

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
```diff
@@ -12,53 +12,103 @@
...
+    return contextNumber;
   }
 
-  public boolean isUint64() {
-    return length().isUInt64();
+  public Bytes extract() {
+    return range.isEmpty()
+        ? Bytes.EMPTY
+        : slice(rawData, safeLongToInt(range.offset()), safeLongToInt(range.size()));
   }
```
OlivierBBB, line 109: This method e.g. extracts the relevant data from the `rawData` e.g.
- `callData` from the `callerMemorySnapshot`
- `returnData` from the `callerMemorySnapshot`

arithmetization/src/main/java/net/consensys/linea/zktracer/types/Range.java
OlivierBBB: image

arithmetization/src/main/java/net/consensys/linea/zktracer/module/Util.java
```diff
@@ -168,25 +168,45 @@ public static short getTxTypeAsInt(TransactionType txType) {
...
+  }
+
+  /**
+   * This method expects a "small-ish" long value and returns the corresponding int value.
+   *
+   * @param value
+   * @return
+   * @throws ArithmeticException
+   */
+  public static int safeLongToInt(long value) throws ArithmeticException {
```
letypequividelespoubelles: Move this fct in conversions.java

arithmetization/src/main/java/net/consensys/linea/zktracer/module/Util.java
```diff
@@ -168,25 +168,45 @@ public static short getTxTypeAsInt(TransactionType txType) {
...
+   * Extracts {@code size} many bytes from {@code data} starting at {@code offset}. If {@code data}
+   * ``runs out'' it substitutes the missing bytes with 0's.
    *
    * @param size Bytes of
    * @param data, right-padded with 0's if needed, starting from
-   * @param positionStart
+   * @param offset
    */
-  public static Bytes slice(Bytes data, int positionStart, int size) {
+  public static Bytes slice(Bytes data, int offset, int size) {
```
letypequividelespoubelles: I was planning to do it in my PR. I'll have to resolve conflicts now 😅

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -1033,7 +1033,7 @@ void traceOpcode(MessageFrame frame) {
         }
 
         final boolean returnFromDeployment =
-            (this.opCode() == RETURN || this.currentFrame().isDeployment());
```
OlivierBBB: No idea what happened here 😆

PR #1554, opened 2024-11-24
fix: `go-corset` running `test` in CI

### LINKED ISSUE #1553: Fix `go-corset` in `fastReplayTests`1
This is again failing for reasons unknown (specifically, test `testMxpRandom()`).  It doesn't seem to me that it can be related to memory, since the amount of memory used for the failing test is quite small.

DavePearce:
This disables the use of parallel forks in gradle unless this is
explicitly requested.  The issue is that parallel forks interferes with
go-corset which is highly concurrent.  Hence, running go-corset tasks in
parallel can (in some cases) lead to out-of-resource issues and workflow
tasks being cancelled unexpectedly.

However, since using parallel forks does improve the runtime for Rust
corset, this option is retained when specifically specified via the
UNIT_TESTS_PARALLELISM environment variable.

### DIFF
--- a/gradle/tests.gradle
+++ b/gradle/tests.gradle
@@ -118,13 +118,11 @@ tasks.withType(Test).configureEach {
 }
 
 tasks.test.configure {
-  if (System.getenv().containsKey("REPLAY_TESTS_PARALLELISM") ||
-      System.getenv().containsKey("WEEKLY_TESTS_PARALLELISM") ||
-      System.getenv().containsKey("REFERENCE_TESTS_PARALLELISM")) {
+  if (System.getenv().containsKey("UNIT_TESTS_PARALLELISM")) {
     // Only set this if parallelism is explicitly requested.
     // Specifically, because this option is not compatible with
     // go-corset.
-    maxParallelForks = Runtime.getRuntime().availableProcessors()
+    maxParallelForks = System.getenv().get("UNIT_TESTS_PARALLELISM").toInteger()
     println "Enabled parallel forks."
   }
   useJUnitPlatform {
--- a/gradle/tests.gradle
+++ b/gradle/tests.gradle
@@ -118,7 +118,15 @@ tasks.withType(Test).configureEach {
 }
 
 tasks.test.configure {
-  maxParallelForks = Runtime.getRuntime().availableProcessors()
+  if (System.getenv().containsKey("REPLAY_TESTS_PARALLELISM") ||
+      System.getenv().containsKey("WEEKLY_TESTS_PARALLELISM") ||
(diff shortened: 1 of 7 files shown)


gradle/tests.gradle
```diff
@@ -118,7 +118,15 @@ tasks.withType(Test).configureEach {
 }
 
 tasks.test.configure {
-  maxParallelForks = Runtime.getRuntime().availableProcessors()
```
gauravahuja, line 121: - The gradle task `test`, `./gradlew test`is only used for unit tests. Thus `maxParallelForks` is only used in unit tests.
- Replay, Weekly and Reference tests have their own gradle tasks that set the parallelism. You can set the parallelism for go corset workflows by either removing the `*_TESTS_PARALLELISM` env varaible or setting it to 1.
- You can introduce another variable for UNIT_TESTS_PARALLELISM that is disabled for go corset workflows

gauravahuja (changes requested):

gradle/tests.gradle
DavePearce: So (surprisingly) the unit tests are the ones causing the problem (they are getting very expensive now).  We cannot have `maxParallelForks` being defined as above when `go-corset` is being used.  That's basically what I'm doing here.  We could have `UNIT_TESTS_PARALLELISM` and, when that is defined, set `maxParallelForks`.  That would work fine I guess, and is pretty comparable to what we've got here.

DavePearce: But I do understand you are saying this is redundant:

```
   if ( ...
      System.getenv().containsKey("WEEKLY_TESTS_PARALLELISM") ||
      System.getenv().containsKey("REFERENCE_TESTS_PARALLELISM")) {
```

Because these keys are never used for the `test` task.

DavePearce: Ok, I added `UNIT_TESTS_PARALLELISM`

gradle/tests.gradle
```diff
@@ -118,7 +118,13 @@ tasks.withType(Test).configureEach {
...
+  if (System.getenv().containsKey("UNIT_TESTS_PARALLELISM")) {
+    // Only set this if parallelism is explicitly requested.
+    // Specifically, because this option is not compatible with
+    // go-corset.
+    maxParallelForks = System.getenv().get("UNIT_TESTS_PARALLELISM").toInteger()
+    println "Enabled parallel forks."
+  }
```
gauravahuja, line 127:
```suggestion
       System.getenv().getOrDefault("UNIT_TESTS_PARALLELISM", "1").toInteger())
```

gauravahuja (approved):

PR #1557, opened 2024-11-25
revert due to performance regression

FlorianHuc:


arithmetization/src/main/java/net/consensys/linea/zktracer/types/Bytecode.java
```diff
@@ -58,7 +57,7 @@ public Bytecode(Bytes bytes) {
    */
   public Bytecode(Code code) {
     this.bytecode = code.getBytes();
-    this.hash = CompletableFuture.completedFuture(code.getCodeHash());
+    this.hash = code.getCodeHash();
```
ahamlat, line 60: Could you double check with the previous version ? I think we don't initialize the hash at build time. This actually what represents the performance regression,

ahamlat (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/types/Bytecode.java
FlorianHuc: It is as it was before the PR. I believe the other constructor is used mainly.

ahamlat: Ah, you're right, I was referring to the other one,

ahamlat (approved):

PR #1558, opened 2024-11-25
change error thrown when too many requests.

FlorianHuc:
change error to keep 200 http error message to ease error handling in coordinator.


arithmetization/src/main/java/net/consensys/linea/plugins/rpc/RequestLimiter.java
```diff
@@ -36,7 +38,7 @@ public RequestLimiter(int concurrentRequestsCount) {
   public <T extends PluginRpcRequest, R> R execute(T request, Function<T, R> processingFunc) {
     if (!semaphore.tryAcquire()) {
       throw new PluginRpcEndpointException(
-          RpcErrorType.INVALID_REQUEST,
+          UNAVAILABLE_EXCEPTION,
```
gauravahuja, line 41: How about this
```suggestion
          RpcErrorType.UNKNOWN,
```

gauravahuja (commented):
Just left one comment on why can't you reuse the existing UNKNOWN error type.

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/RequestLimiter.java
FlorianHuc: Using a dedicated error allows to pass an error code that the coordinator can use to customize its error handling.

gauravahuja (approved):

PR #1566, opened 2024-11-25
Reenable TX Outcome Checks for Replay Tests

### LINKED ISSUE #1521: Reenable Tx Outcome Checks
The goal here is to undo the temporary fix in #1520.  The problem is that a change in the BESU version (which fixed a bug with traceEndTransaction) has meant that all recorded tx outcomes are potentially invalid with respect to e.g. the sender.  The temporary fix in #1520 simply disabled tx outcome checks.  The goal now is to update the recorded tx outcome information so that the tests will pass with outcome checking enabled.

DavePearce:
This updates the record TX outcomes in the replay test files to correctly follow mainnet. This was done by playing the replays back through the ReplayExecutionEnvironment and the recording the outcomes using the BlockCapturer.

**NOTE**: this also updates linea-besu to 660a354 (as this includes the finalised fix for traceEndTransaction()).


fab-10 (commented):
the "Update BesuPlugin implementation" commit LGTM

gauravahuja (approved):

PR #1571, opened 2024-11-26
1568 some last mmio issues second part

### LINKED ISSUE #1568: some last MMIO issues
ref tests 120, 150, 201

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
```diff
@@ -86,8 +86,7 @@ public MemoryRange(
   public MemoryRange(final long contextNumber, final Range range, final MessageFrame frame) {
     this.contextNumber = contextNumber;
     this.range = range;
-    this.rawData =
-        (range.isEmpty()) ? Bytes.EMPTY : frame.shadowReadMemory(0, frame.memoryByteSize());
+    this.rawData = frame.shadowReadMemory(0, frame.memoryByteSize());
```
letypequividelespoubelles: The only fix: as we pass the memory snapshot to the MMU, we need to take a snapshot of the memory even if callData is empty. If not, we don't remember the caller memory when doing a (empty) sha2 for exempel

OlivierBBB: I don't understand. In case of a SHA2-256, and unless both `callDataRange` and `returnAtRange` are both empty, we will remember the caller memory one time or two times. And if both are empty we don't need the caller memory. What am I missing ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/precompileSubsection/PrecompileSubsection.java
```diff
@@ -209,7 +209,9 @@ public long returnDataSize() {
   }
 
   public Bytes rawCallerMemory() {
-    return callSection.getCallDataRange().getRawData();
```
letypequividelespoubelles, line 212: This is the fix after @OlivierBBB 's comment: to get the rawCallerMemory, if the callDataRange is empty, then we check in the returnAtRange

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
```diff
@@ -73,21 +73,21 @@ public MemoryRange(long contextNumber) {
   public MemoryRange(final long contextNumber, final Range range, final Bytes rawData) {
     this.contextNumber = contextNumber;
     this.range = range;
-    this.rawData = rawData;
+    this.rawData = range.isEmpty() ? Bytes.EMPTY : rawData;
```
letypequividelespoubelles, line 76: this (and following) are just to make the constructor consistent: if the range isEmpty, then don't store the rawData

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
```diff
@@ -113,6 +113,6 @@ public boolean isEmpty() {
   }
 
   public MemoryRange snapshot() {
-    return new MemoryRange(this.contextNumber, this.range.snapshot(), this.rawData);
+    return new MemoryRange(contextNumber, range.snapshot(), rawData.copy());
```
letypequividelespoubelles, line 116: not sure it's usefull as rawData is final ... better to just store a pointer instead of a copy, wdyt, do we have any chance we may require a copy @OlivierBBB  ?

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
letypequividelespoubelles: Yep, you right. I modified `rawCallerMemory()` to retrieve the memory from the `returnAtRange`if the `callDataRange` is empty.

Next point, as we almost always have a nin empty callData and a returnAt, wouldn't it be better to have a pointer in each, pointing to the full memory snapshot ? In almost all cases, we'll store double data for nothing.

OlivierBBB: I think it's worth changing down the road. The current solution is clearly sub optimal in terms of memory usage, but it's simple and I don't expect that it will have a performance impact. I would leave things as they are for now.

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
OlivierBBB: The `rawData` being final only means that the reference is set in stone. The contents not so much. I'm for this change. We will in any case see if this creates issues. Let's do the safe thing for now.

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
```diff
@@ -73,21 +73,21 @@ public MemoryRange(long contextNumber) {
...
-    this.rawData = rawData;
+    this.rawData = range.isEmpty() ? Bytes.EMPTY : rawData;
   }
 
   public MemoryRange(
       final long contextNumber, final long offset, final long size, final Bytes rawData) {
     this.contextNumber = contextNumber;
     this.range = Range.fromOffsetAndSize(offset, size);
-    this.rawData = rawData;
+    this.rawData = range.isEmpty() ? Bytes.EMPTY : rawData;
```
OlivierBBB: Same `range` and `isEmpty()` remark

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
```diff
@@ -73,21 +73,21 @@ public MemoryRange(long contextNumber) {
...
   }
 
   public MemoryRange(final long contextNumber, final Range range, final MessageFrame frame) {
     this.contextNumber = contextNumber;
     this.range = range;
     this.rawData =
-        (range.isEmpty()) ? Bytes.EMPTY : frame.shadowReadMemory(0, frame.memoryByteSize());
+        range.isEmpty() ? Bytes.EMPTY : frame.shadowReadMemory(0, frame.memoryByteSize());
```
OlivierBBB, line 90: You could directly write `isEmpty()` like below i.e. not mention the `range`

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
OlivierBBB: Yes

arithmetization/src/main/java/net/consensys/linea/zktracer/types/MemoryRange.java
OlivierBBB: Although this is a constructor ... I'm not sure if you can "already" call the `isEmpty()` method here 🤔

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
```diff
@@ -361,8 +354,17 @@ private void smcProcessing(Hub hub, MessageFrame frame) {
...
+        ADDRESS_TO_PRECOMPILE.get(preOpcodeCalleeSnapshot.address()).apply(hub, this);
     hub.defers().scheduleForContextEntry(this);
     hub.defers().scheduleForContextReEntry(this, hub.currentFrame());
+    // In case of arguments too large for MODEXP, transaction will be popped anyway, and resolving
+    // some defers will create NPE
+    if (precompileSubsection instanceof ModexpSubsection
+        && ((ModexpSubsection) precompileSubsection).transactionWillBePopped) {
+      hub.defers().unscheduleForContextReEntry(this, hub.currentFrame());
+      hub.defers().unscheduleForPostTransaction(this);
+    }
```
OlivierBBB, line 367: Wouldn't it make more sense to
- schedule all the `PrecompileSubsection`'s that aren't `instanceof ModexpSubsection`
- only schedule a `ModexpSubsection` if `!transactionWillBePopped` ?

You are scheduling and unscheduling but you already have all requisite information to decide whether you will need to `schedule` at all.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
```diff
@@ -313,21 +309,18 @@ private void abortingCall(Hub hub) {
...
+  private void refineUndefinedScenario(Hub hub, MessageFrame frame) {
 
     final boolean aborts = hub.pch().abortingConditions().any();
     if (aborts) {
       scenarioFragment.setScenario(CALL_ABORT_WONT_REVERT);
       return;
     }
 
-    final WorldUpdater world = hub.currentFrame().frame().getWorldUpdater();
+    final WorldUpdater world = frame.getWorldUpdater();
```
OlivierBBB, line 320: yes

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
```diff
@@ -224,13 +220,13 @@ public CallSection(Hub hub, MessageFrame frame) {
...
-    CallScenarioFragment.CallScenario scenario = scenarioFragment.getScenario();
+    refineUndefinedScenario(hub, frame);
+    final CallScenarioFragment.CallScenario scenario = scenarioFragment.getScenario();
     switch (scenario) {
-      case CALL_ABORT_WONT_REVERT -> abortingCall(hub);
       case CALL_EOA_UNDEFINED -> eoaProcessing(hub);
-      case CALL_PRC_UNDEFINED -> prcProcessing(hub);
       case CALL_SMC_UNDEFINED -> smcProcessing(hub, frame);
+      case CALL_PRC_UNDEFINED -> prcProcessing(hub);
+      case CALL_ABORT_WONT_REVERT -> abortingCall(hub);
```
OlivierBBB, line 229: Is this reorg an optimization à la "most CALL's will be EOA / SMC and aborts will be rare" ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
```diff
@@ -211,7 +207,7 @@ public CallSection(Hub hub, MessageFrame frame) {
 
     value =
         opCode.callHasValueArgument()
-            ? Wei.of(currentFrame.frame().getStackItem(2).toUnsignedBigInteger())
+            ? Wei.of(frame.getStackItem(2).toUnsignedBigInteger())
```
OlivierBBB, line 210: yes

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/defer/DeferRegistry.java
```diff
@@ -224,4 +224,8 @@ public void resolveUponContextReEntry(Hub hub, CallFrame callFrame) {
   public void unscheduleForContextReEntry(ContextReEntryDefer defer, CallFrame callFrame) {
     contextReEntryDefers.get(callFrame).remove(defer);
   }
+
+  public void unscheduleForPostTransaction(PostTransactionDefer defer) {
+    postTransactionDefers.remove(defer);
+  }
```
OlivierBBB, line 230: As explained below I'm not convinced we need this.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/precompileSubsection/ModexpSubsection.java
```diff
@@ -75,6 +76,7 @@ public ModexpSubsection(final Hub hub, final CallSection callSection) {
             >= 0) {
```
OlivierBBB: I don't remember the precise semantics of `BigInteger#compareTo` but it looks like the logic here is saying that "if either `bbs`, `ebs`, `mbs` is `≥ 512` then the `MODEXP` operation is rejected by Linea".

But the real condition is "if either `bbs`, `ebs`, `mbs` is `> 512` then the `MODEXP` operation is rejected by Linea".

> Is my interpretation correct ? If so we must replace all `>=` by `>`.

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/callstack/CallFrame.java
```diff
@@ -71,7 +71,6 @@ public boolean isMessageCall() {
   @Getter private int byteCodeDeploymentNumber;
   private EWord eCodeAddress = null; // memoization
   @Getter private Bytecode code = Bytecode.EMPTY;
-  @Getter private int codeFragmentIndex = -1;
```
OlivierBBB, line 74: Isn't this desirable ? A sort of sanity check ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
letypequividelespoubelles: yes, only.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/precompileSubsection/ModexpSubsection.java
letypequividelespoubelles: I think you're absolutely right. I'm making the change. Very good catch.

arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/callstack/CallFrame.java
letypequividelespoubelles: it wasn't used at all, as we're using a function called `getCodeFragmentIndex()`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
letypequividelespoubelles: I do agree, but I find it way more readable to do it in any case, and undo it in some (very rare) case.

letypequividelespoubelles: plus when we schedule for postTransaction, we still don't have the information as the precompile subsection is created after. So in any case we'll have to unschedule it.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/defer/DeferRegistry.java
letypequividelespoubelles: see other comment. We schedule for postTx at the creation of the callSection, when the (modexp) subsection is still not created, so we don't have the info at the time of the schedulling.

OlivierBBB: So wouldn't we have to unscheduled the postTransaction stuff, too ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/call/CallSection.java
OlivierBBB: I agree that both should be equivalent. Your call

OlivierBBB (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/defer/DeferRegistry.java
letypequividelespoubelles: this is the postTx we unschedule !

letypequividelespoubelles: The one we don't unSchedule (because it doesn't create NPE) is contextEntry

PR #1572, opened 2024-11-27
`CREATE` tests

### LINKED ISSUE #1599: Deployment tests: nontrivial deployments
# Definitions

Nontrivial CREATE's are those CREATE's that are
- unexceptional
- don't abort
- don't raise the failure condition **F**
- but deploy with **nonempty** initialization code

# Parameters

In terms of the parameters we can play with we find a lot, so many that breaking them up into orthogonal subsets is mandatory

The first distinction is wrt its **initiation**
- [ ] deployment transaction
- [ ] CREATE-induced deployment
- [ ] CREATE2-induced deployment

And we must explore the option where said deployment gets reverted later or not
- [ ] WILL_REVERT
- [ ] WONT_REVERT

Then **during deployment** we can play around with
- [ ] modifying its own storage during deployment phase
- [ ] (using CREATE2) calling its own creator to attempt a second CREATE2 deployment within the deployment at the current deployment address

In terms of how we **terminate deployment** we must explore those terminations that **revoke it altogether**
- [ ] exception terminates deployment
  - [ ] standard exception
  - [ ] maxCodeSizeException
  - [ ] invalidCodePrefixException
- [ ] REVERT terminates deployment, revoking it altogether

In terms of how we **terminate deployment** we must explore those terminations that **lead to (temporary) deployment**
- [ ] STOP terminates deployment (leading to empty deployment)
- [ ] SELFDESTRUCT terminates deployment (leading to empty deployment)
- [ ] REVERT terminates deployment
  - [ ] empty return data
  - [ ] nonempty return data
- [ ] RETURN terminates deployment
  - [ ] empty deployment
  - [ ] nonempty deployment

Once we are in the realm of successful deployments we must explore **deployed byte code**, e.g. byte code
- [ ] allows for modifying its own storage
- [ ] allows for SELFDESTRUCT through its own code
- [ ] ability to DELEGATECALL into other accounts

# Specific scenarios that we need to test

- [ ] deployment transaction
  - [x] COINBASE is the deployment address [yes / no]
  - [ ] deployment address already has funds [yes / no]
  - [ ] deployment success [yes / no]
  - [ ] 0th transaction CALL's future deployment address with value [yes / no]
  - [ ] 2nd transaction CALL's failed deployment address with value [yes / no]
  - [ ] 2nd transaction EXTCODEHASH's failed deployment address [yes / no]
  - [ ] 2nd transaction SELFDESTRUCT's and targets failed deployment address with value [yes / no]
- [ ] same thing with CREATE-induced deployment (though this should be working fine as is)

The rationale for the last points: the recent change that we made to TX_INIT / TX_FINL solved an issue whereby
after a failed deployment transaction the deployment address would still have a nonzero nonce (≥ 1) and thus would
EXIST in the state σ after transaction end. This would have down stream effects such as
- EXTCODEHASH would not behave the same
- CALL with value may cost 25k less gas
- sending funds via SELFDESTRUCT would cost 25k less gas

### LINKED ISSUE #1600: Deployment tests: trivial deployments
# Definitions

Trivial CREATE's are those CREATE's that are
- unexceptional
- don't abort
- don't raise the failure condition **F**
- but deploy with empty initialization code

# Parameters

In terms of the parameters we can play with we find:
- [ ] level
  - [x] root level
  - [ ] higher level
- [x] CREATE vs CREATE2
- [x] zero value vs nonzero value
- [x] WILL_REVERT vs WONT_REVERT
- [ ] zero size
  - [ ] 'zero size, huge offset' shenanigans
- [ ] nonzero size

We _could_ also consider collisions via CREATE2 or through multi transactions and SELFDESTRUCT's.
This will be the focus of other tests. We should also measure the interaction that these trivial deployments have with EXTCODEHASH.
Furthermore, we should see how these deployments squash any return data (if present.) And then there is obviously the question of deployment numbers, deployment statuses.

- EXTCODEHASH before and after
- squashing of return data and RETURNDATASIZE
- warmth post deployment (EXTCODEXXX + BALANCE + CALL's + SELFDESTRUCT heir + ...)

### LINKED ISSUE #1601: Deployment tets: failure
# Definitions

**Failure Condition F** CREATE's are those CREATE's that
- are unexceptional
- don't abort
- but raise **Failure Condition F**

Recall that this condition is triggered when the deployment address has one or more of the following properties
- nonzero nonce
- nonempty code

**Note.** Nonzero balance isn't an issue when scanning for **Failure Condition F**..

## Sequential **Failure Condition F** via `CREATE2`

Raising **Failure Condition F** via CREATE2 is trivial. When testing this we should look at
- init code
  - [ ] is empty
  - [ ] is nonempty
- deployment 1
  - [ ] success (only if init code not empty)
    - [ ] yes
      - [ ] finishes on STOP (empty deployment)
      - [ ] finishes on SELFDESTRUCT (empty deployment)
      - [ ] finishes on RETURN (pot. nonempty deployment)
    - [ ] no
      - [ ] finishes on exception
      - [ ] finishes on REVERT
  - [ ] gets reverted before deployment 2
    - [ ] yes
    - [ ] no
- deployment 2
  - same questions
  - raises or not the 

## Nested **Failure Condition F** via `CREATE2`

- contract A does a CREATE2 with init code that calls A again to do a CREATE2 with the same init code.

## Failure Condition F via `CREATE`

This is classical by now:
- contract A does a CREATE2 deploying contract B
  - contract B should be able to do a SELFDESTRUCT / CREATE
- contract B does a CREATE deploying contract C while it has nonce n (e.g. n = 1)
- contract B selfdestructs
- contract A re-deploys contract B using CREATE2 again
- contract B attempts a CREATE at the same nonce n as previously, effectively attempting to re-deploy at contract C's address

I guess the simplest way to achieve this would be to have B have the following code

```asm
// contract B deployed code
PUSH1 0
PUSH1 0
PUSH1 0
PUSH1 0
PUSH1 0
CALLDATALOAD  // expecting to find an address in call data
GAS
DELEGATECALL
```

This byte code measures 13 bytes. The associated init code would be
```asm
PUSH13 ...
PUSH1 0
MSTORE
PUSH1 13
PUSH1 0
RETURN
```

which is 22 bytes. It should be deployed in a CREATE2 with 1 Wei of value.

We then need two contracts:
- one that simply selfdestructs
- another one that CREATEs.

E.g.

```asm
// basic CREATE-or contract (deploys 0x)
PUSH1 0
PUSH1 0
PUSH1 0
CREATE
```
or
```asm
// alternative basic CREATE-or contract (deploys 0x00 .. 00)
PUSH1 32
PUSH1 0
PUSH1 0
CREATE
```

```asm
// basic SELFDESTRUCT-or
ORIGIN
SELFDESTRUCT
```

The simplest scenario would be
- transaction 1:
  - to contract A
  - deploy contract B
  - get back to A
  - CALL contract B with call data the address of the basic CREATE-or
  - CALL contract B with call data the address of the basic SELFDESTRUCT-or
- transaction 2:
  - to contract A
  - deploy contract B (which is clean)
  - CALL contract B with call data the address of the basic CREATE-or
    - raise **Failure Condition F**

### LINKED ISSUE #1602: Deployment tests: abort
# Definitions

Aborted CREATE's are those CREATE's that are
- unexceptional
- attempt to provide more balance than is avaiable or happen at CSD = 1024

# Parameters

- [ ] fail because of BALANCE
- [ ] fail because of CALL_STACK_DEPTH

We can then
- [ ] do an EXTCODEHASH on the deployment address to test
  - warmth (shouldn't be warm)
  - existence (should exist or not depending on balance alone unless collision)
- [ ] two deployment tests
  - [ ] first deployment aborts with CREATE(2), a second deployment at this place doesn't abort
  - [ ] first deployment succeeds with CREATE2, a second attempt aborts before FailureCondition **F** can be raised.

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -206,6 +206,7 @@ public CreateSection(Hub hub, MessageFrame frame) {
     if (failedCreate) {
       finalContextFragment = ContextFragment.nonExecutionProvidesEmptyReturnData(hub);
       scenarioFragment.setScenario(CREATE_FAILURE_CONDITION_WONT_REVERT);
+      commonValues.payGasPaidOutOfPocket(hub);
```
OlivierBBB, line 185: omission

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/CreateSection.java
```diff
@@ -361,8 +364,8 @@ private void finishAbort(final Hub hub) {
         hub.factories().accountFragment();
     final AccountFragment creatorAccountFragment =
         accountFragmentFactory.make(
-            preOpcodeCreateeSnapshot,
-            preOpcodeCreateeSnapshot,
+            preOpcodeCreatorSnapshot,
+            preOpcodeCreatorSnapshot,
```
OlivierBBB, line 409: typo

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpaddr/RlpAddr.java
```diff
@@ -109,7 +109,13 @@ public void callRlpAddrCreate2(MessageFrame frame, Bytes32 salt, Bytes32 hash) {
...
     final Bytes32 rawCreate2Address = getCreate2RawAddress(currentAddress, salt, hash);
     final RlpAddrOperation operation =
-        new RlpAddrOperation(rawCreate2Address, OpCode.CREATE2, currentAddress, salt, hash);
+        new RlpAddrOperation(
+            rawCreate2Address,
+            OpCode.CREATE2,
+            currentAddress,
+            salt,
+            hash,
+            longToUnsignedBigInteger(frame.getWorldUpdater().get(currentAddress).getNonce()));
```
OlivierBBB, line 118: Even though it's unnecessary, the nonce is part of the HUB -> RLPADDR lookup for CREATE2.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpaddr/RlpAddr.java
```diff
@@ -134,6 +142,7 @@ private void traceCreate2(int stamp, RlpAddrOperation chunk, Trace trace) {
           .saltLo(chunk.salt().orElseThrow().slice(LLARGE, LLARGE))
           .kecHi(chunk.keccak().orElseThrow().slice(0, LLARGE))
           .kecLo(chunk.keccak().orElseThrow().slice(LLARGE, LLARGE))
+          .nonce(nonceBytes)
```
OlivierBBB, line 145: We trace the nonce in all cases.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpaddr/RlpAddrOperation.java
```diff
@@ -49,8 +49,13 @@ public RlpAddrOperation(Bytes32 rawDepAddress, OpCode opCode, BigInteger nonce,
...
   // CREATE2 operation
   public RlpAddrOperation(
-      Bytes32 rawHash, OpCode opCode, Address address, Bytes32 salt, Bytes32 kec) {
-    this(rawHash, opCode, Optional.empty(), address, Optional.of(salt), Optional.of(kec));
+      Bytes32 rawHash,
+      OpCode opCode,
+      Address address,
+      Bytes32 salt,
+      Bytes32 kec,
+      BigInteger nonce) {
```
OlivierBBB, line 56: Adding the nonce to the interface

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Create.java
```diff
@@ -52,12 +52,13 @@ public long largestOffset() {
...
 
-    if (gasCost > currentGas) {
+    if (upfrontGasCost > currentGas) {
       return 0;
-    } else {
-      return currentGas - currentGas / 64;
     }
+
+    long gasMinusUpfront = currentGas - upfrontGasCost;
+    return gasMinusUpfront - gasMinusUpfront / 64;
```
OlivierBBB, line 62: This was a genuine mistake. the 63/64-ths business happens after paying the upfront gas cost.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Create2.java
```diff
@@ -57,13 +57,15 @@ public long linearPerWord() {
...
-    if (gasCost > currentGas) {
+    if (upfrontGasCost > currentGas) {
       return 0;
-    } else {
-      return currentGas - currentGas / 64;
     }
+
+    final long gasMinusUpfrontGasCost = currentGas - upfrontGasCost;
+
+    return gasMinusUpfrontGasCost - gasMinusUpfrontGasCost / 64;
```
OlivierBBB, line 68: Same mistake for CREATE2

testing/src/main/java/net/consensys/linea/testing/ToyAccount.java
```diff
@@ -202,4 +202,9 @@ public ReferenceTestWorldState.AccountMock toAccountMock() {
     return new ReferenceTestWorldState.AccountMock(
         Long.toHexString(nonce), balance.toHexString(), accountMockStorage, code.toHexString());
   }
+
+  public ToyAccount raiseNonceBy(long k) {
+    long updatedNonce = this.getNonce() + k;
+    return new ToyAccount(this.parent, this.getAddress(), updatedNonce, this.balance, this.balance);
+  }
```
OlivierBBB, line 209: Bandaid

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/createTests/failure/Create.java
```diff
@@ -0,0 +1,236 @@
...
+                  + (1 + 1)
+                  + (1 + 1)
+                  + (1 + 1)
+                  + 1
+                  + (1 + 1)
+                  + 1
+                  + 1 //
+              )
+          .op(JUMPI)
+          //
```
lorenzogentile404: Could you add some comment about the purpose of the programs you are defining in this test class, especially for this long one above with this sum of 1s.

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/createTests/trivial/RootLevel.java
```diff
@@ -0,0 +1,193 @@
...
+
+    switch (sizeParameter) {
+      case ZERO -> program.push(0);
+      case TWELVE -> program.push(12);
+      case THIRTEEN -> program.push(13);
+      case FOURTEEN -> program.push(14);
+      case THIRTY_TWO -> program.push(0x20);
+      case MSIZE -> program.op(MSIZE);
+      case MAX -> program.push("ff".repeat(32));
+    }
```
lorenzogentile404, line 142: Here you may consider to do the following.

Modify the enum to:

```
public enum SizeParameter {
  ZERO(0),
  TWELVE(12), // - 3 - 1
  THIRTEEN(13), // - 3 + 0
  FOURTEEN(14), // - 3 + 1
  THIRTY_TWO(32),
  MSIZE,
  MAX;

 public int getValue() {
        if (this == MSIZE || this == MAX) {
            throw new UnsupportedOperationException("...");
        }
        return this.value;
    }

  public boolean isAnyOf(SizeParameter... sizeParameters) {
    for (SizeParameter sizeParameter : sizeParameters) {
      if (this == sizeParameter) {
        return true;
      }
    }
    return false;
  }

  public boolean willRaiseException() {
    return this.isAnyOf(MAX);
  }
}
```
Then, the code above would become:

```
 switch (sizeParameter) {
      case MSIZE -> program.op(MSIZE);
      case MAX -> program.push("ff".repeat(32));
      deafult -> program.push(sizeParameter.getValue();
    }
```
However, it is not a big deal to leave it as it is, but this could be a good approach for the future.

lorenzogentile404: Another approach may be using `Either`, where you basically either store an `int` or another type, that could be an `enum`, but it is probably not worth facing that complexity for this simple scenario.

lorenzogentile404: also, what translates to an integer may have a more meaningful name rather than `ZERO, TWELVE` etc.

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/createTests/failure/Create.java
OlivierBBB: Done.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -625,7 +625,7 @@ public void traceContextEnter(MessageFrame frame) {
           frameType,
           newChildContextNumber(),
           this.deploymentStatusOf(frame.getContractAddress()),
-          frame.getValue(),
+          frame.getApparentValue(),
```
OlivierBBB, line 628: Important bug fix, required e.g. for DELEGATECALL to provide the CallFrame with the correct `value` i.e. `callValue`

PR #1573, opened 2024-11-27
update equals on wcpoperation due to Bytes32 equal method inconsistency.

### LINKED ISSUE #1564: lineCounting issue in WCP
ref test 506: pointMulAdd_d4g2v0 : we have a final row of 0s at the end of wcp module

FlorianHuc:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/WcpOperation.java
```diff
@@ -249,4 +248,33 @@ protected int computeLineCount() {
...
+    if (o == null || getClass() != o.getClass()) return false;
+    WcpOperation that = (WcpOperation) o;
+    return wcpInst == that.wcpInst
+        && compareBytes(arg1, ((WcpOperation) o).arg1())
+        && compareBytes(arg2, ((WcpOperation) o).arg2());
+  }
+
+  private boolean compareBytes(Bytes32 current, Bytes32 other) {
+    for (int i = 0; i < 32; i++) {
+      if (current.get(i) != other.get(i)) {
```
letypequividelespoubelles: Better perf to check from last bytes as we'll have often a lot of leading 0s

ahamlat (approved):
LGTM, maybe add few more tests to check that different Eword returns false when calling equals

PR #1585, opened 2024-11-28
clean: no need to store two times same hub in each section

### LINKED ISSUE #1533: In a section we have two pointers to the `HUB`, one in `commonFragment` one in the `TraceSection` itself
We should keep only one of the two

letypequividelespoubelles:


OlivierBBB (approved):
LGTM

PR #1588, opened 2024-11-28
fix: Daily Blockchain Reference Test Workflow Failures

### LINKED ISSUE #1577: Fix Daily Blockchain Reference Tests
These are currently, but actually reporting a pass in Github.

DavePearce:
Currently, workflow failures are being recorded as passing.  The reason for this is simply a flag within the yaml file that was presumably copied over from the original gradle tests.  This should now resolve the issus, and make sure that when the workflow fails --- we see that.


FlorianHuc (approved):
LG

PR #1593, opened 2024-12-01
Split Unit and Replay Tests

### LINKED ISSUE #1592: Split Unit Tests from Replay Tests
Since the Unit Tests are now taking quite a while, its worth splitting up these two things and potentially doing them concurrently

DavePearce:
The goal here is to split out the unit and replay tests into distinct workflows.  This will allow them to be potentially executed in parallel, leading to reduced time to PR.  It will also give slightly more feedback in terms of which parts are failing.


.github/workflows/gradle-fast-replay-tests.yml
```diff
@@ -122,13 +115,6 @@ jobs:
       - name: Setup Go Corset
         uses: ./.github/actions/setup-go-corset
 
-      - name: Run unit tests
```
gauravahuja: You can run unit tests and replay tests in parallel and still keep all the job definitions in the same yaml file. IMO the benefit of having one yaml file is that all the jobs that are run on every commit are defined in one single place. 

As of now there are 3 major job definitions in this file:

1. build
2. tests
3. go-corset-tests

Both tests and go-corset-tests depend on the build job and run in parallel after the build job completes.

As part of the tests and go-corset-tests we run unit tests and replay tests sequentially. One reason for doing this was to cut on unnecessary runs for replay tests in case the unit tests failed.

You can remove the tests (anb go-corset-tests) job(s) and introduce `unit_tests` and `replay_tests` jobs which can run in parallel after the build job is complete.

DavePearce: Yeah, I understand what you are saying ... but having everything in one file can be crowded.  However, factoring out the common parts would make a big difference.  In fact, @FlorianHuc is looking into that separately as it happens.

DavePearce: I will try to refactor it though, as I understand you're saying there's duplication which is not needed here.

DavePearce: Also, are you familiar with this:

```
     - name: Setup upterm session
        if: ${{ inputs.tests-with-ssh }}
        uses: lhotari/action-upterm@v1
```

Do we ever actually use this?

DavePearce: Ok, I have now put through a fairly heavy refactoring.  Hopefully that's better!

gauravahuja (approved):
I think this a great upgrade to the workflows. 

Also, I see linea-constraints updated as part of this PR, was that intended?

DavePearce:
@gauravahuja I will check `linea-constraints` now.  Its probably just that this branch is out of date.  I'll update and check.

DavePearce:
@gauravahuja Yes, that has resolved it!!

PR #1607, opened 2024-12-02
MmioPatterns optimisation

ahamlat:
Following an issue we faced recently on sepolia on linea_generateConflatedTracesToFileV2, where tracing the batch of blocks take more than 30 seconds we noticed that MmioPatterns.updateTemporaryTargetRam method takes more than 2.5 seconds to execute (see profiling below).

This PR improves this method by using Java native calls (System.arraycopy) to copy byte arrays instead of using Tuweni library.

**Before this PR**

image

image

**With this PR**
image

image

Currently the test takes around 25 seconds instead of 30 seconds
```
[ec2-user@ip-10-40-43-134 besu]$ time curl --location --request POST 'localhost:8545' --header 'Content-Type: application/json' --data-raw '{"jsonrpc": "2.0", "method": "linea_generateConflatedTracesToFileV2", "params": [{"startBlockNumber":6488125,"endBlockNumber":6488189, "expectedTracesEngineVersion": "0.8.0-rc6"}], "id": 1}'
{"jsonrpc":"2.0","id":1,"result":{"tracesEngineVersion":"0.8.0-rc6","conflatedTracesFileName":"/data/besu/conflated/6488125-6488189.conflated.0.8.0-rc6.lt"}}
real	0m24.473s
user	0m0.005s
sys	0m0.000s
```

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
@@ -182,18 +182,6 @@ public class MmioPatterns {
 
   public static void updateTemporaryTargetRam(
       MmuData mmuData, final long targetLimbOffsetToUpdate, final Bytes16 newLimb) {
-
-    int limbStart = (int) (LLARGE * targetLimbOffsetToUpdate);
-    int limbEnd = limbStart + LLARGE;
-    byte[] originalRam = mmuData.targetRamBytes().toArray();
-    int sliceSize = Math.max(0, originalRam.length - limbEnd);
-
-    int size = limbStart + newLimb.size() + sliceSize;
-    byte[] updatedRam = new byte[size];
-    System.arraycopy(newLimb.toArray(), 0, updatedRam, limbStart, newLimb.size());
-    if (sliceSize > 0)
-      System.arraycopy(originalRam, limbEnd, updatedRam, limbStart + newLimb.size(), sliceSize);
-
-    mmuData.targetRamBytes(Bytes.wrap(updatedRam));
+    mmuData.targetRamBytes().set((int) targetLimbOffsetToUpdate, newLimb);
   }
 }
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
@@ -181,7 +181,7 @@ public class MmioPatterns {
   }
 
   public static void updateTemporaryTargetRam(
-          MmuData mmuData, final long targetLimbOffsetToUpdate, final Bytes16 newLimb) {
+      MmuData mmuData, final long targetLimbOffsetToUpdate, final Bytes16 newLimb) {
     byte[] originalRam = mmuData.targetRamBytes().toArray();
 
     int limbStart = (int) (LLARGE * targetLimbOffsetToUpdate);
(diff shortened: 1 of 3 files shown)


linea-constraints
FlorianHuc, line 1: this should not be updated.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
```diff
@@ -182,6 +182,23 @@ public static Bytes16 excision(
 
   public static void updateTemporaryTargetRam(
       MmuData mmuData, final long targetLimbOffsetToUpdate, final Bytes16 newLimb) {
+
+    int limbStart = (int) (LLARGE * targetLimbOffsetToUpdate);
```
FlorianHuc: Maybe we could add a simple unit test to validate the method?

linea-constraints
ahamlat: Not sure why it was updated, it changed when I checkout the project. I thought it was expected.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
ahamlat: Agreed, @letypequividelespoubelles, I will need some help on creating a unit test as we don't have one before. I don't visualize yet the real use case.

ahamlat: Or maybe @OlivierBBB, as Francois is OOO currently.

letypequividelespoubelles: This is needed when a (target) limb is updated in two times by a MMU instructions. It's the case for example for MMU instructions generating *TwoTarget MMIO instructions. 
We already have some memory tests that triggers those MMIO inst. 

Maybe this is the failing unit tests (not much network at the hospital, can't open the test output), but I guess it's MMIO.ram_excision or MMIO.*_to_ram_two_targets constraints that are failing.

linea-constraints
letypequividelespoubelles: Just point to latest master of constraint and it'll be ok.  The tracer is currently pointing to a commit that disables the MMIO (needed for the release)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
```diff
@@ -182,17 +182,18 @@ public static Bytes16 excision(
...
-
-    mmuData.targetRamBytes(Bytes.concatenate(bytesPreLimb, newLimb, bytesPostLimb));
+
+    int limbStart = (int) (LLARGE * targetLimbOffsetToUpdate);
+    int limbEnd = limbStart + LLARGE;
+    byte[] originalRam = mmuData.targetRamBytes().toArray();
+    int sliceSize = Math.max(0, originalRam.length - limbEnd);
+
+    int size = limbStart + newLimb.size() + sliceSize;
+    byte[] updatedRam = new byte[size];
```
letypequividelespoubelles: Here the updated ram size should be the same as the original one ram

letypequividelespoubelles: What's this method you're modifying is doing : you just want to modify a 16 bytes limb the midldle of your memory so :
- you wipe everything before the limb (you could keep it, but as you don't need it anymore, tiu can put 0s instead)
- you modify your 16 bytes limb
- you keep unchanged everything that's after the unmodified limb


Not sure of what you're doing, I'm on mobile with low network

letypequividelespoubelles (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/MmioPatterns.java
ahamlat: > Here the updated ram size should be the same as the original one ram

In my current implementation, it is the case, but any way, we need to follow this logic. The total size is the sum of the parts' sizes that we're putting together.

ahamlat: > Not sure of what you're doing, I'm on mobile with low network

I'm doing the same work, but this implementation is faster as you can see in the flamegraphs

ahamlat:
Using the new approach we discussed today during the meeting is even better. So the idea is to use an implementation of MutableByte on targetRamBytes and set the new limb at the right offset. 
The method doesn't appear in the profiling any more

image

ahamlat:
I don't understand why linea-constraints are modified, I haven't modify those files myself, maybe an encoding issue as I'm working on Mac.

PR #1613, opened 2024-12-03
Build and upload plugin archive artifact on release

### LINKED ISSUE #1612: Build and upload plugin archive artifact on release
Add the `distPlugin` task to the release workflow, so that artifacts that is produced is added to the release, this way we can start migrating deployment procedure to use the plugin archive instead of the shaded jar

fab-10:
fixes #1612

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,8 +25,8 @@ jobs:
         id: get_version
         run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
 
-      - name: Build without tests
-        run: ./gradlew build -PreleaseVersion=${{ steps.get_version.outputs.VERSION }} -x test -x spotlessCheck
+      - name: Build artifacts
+        run: ./gradlew artifacts -PreleaseVersion=${{ steps.get_version.outputs.VERSION }}
         env:
           JAVA_OPTS: -Xmx2g -Dorg.gradle.daemon=false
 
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -25,8 +25,8 @@ jobs:
         id: get_version
         run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
 
-      - name: Build without tests
-        run: ./gradlew build -PreleaseVersion=${{ steps.get_version.outputs.VERSION }} -x test -x spotlessCheck
+      - name: Build artifacts
+        run: ./gradlew artifacts -PreleaseVersion=${{ steps.get_version.outputs.VERSION }}
```
DavePearce, line 29: Ok, I understand.  On a release workflow, you want to run the tests to ensure the release is safe.  The only issue I see here is that this takes quite a while to run through.  In fact, I don't know if its even possible on a local machine ?

DavePearce: Hmmm, so I guess that doesn't actually matter!!  Its going to be running the CI pipeline.  But, still I'm not clear exactly which of the various test suits it will run now.  I'll check this for my own sanity.

macfarla (approved):
LGTM

fab-10:
No tests are run by this, it only build the jar and the zip

Il mer 4 dic 2024, 01:44 David Pearce ***@***.***> ha scritto:

.github/workflows/release.yml
fab-10: No tests are run by this, `artifacts` is only an alias to build the jar and the zip

PR #1615, opened 2024-12-03
Feat/multi transaction with create tests

### LINKED ISSUE #1590: Improve `ToyAccount` management in `BytecodeRunner` and `ToyExecutionEnvironmentV2`
Currently, when we write tests, we explicitly create accounts in the test itself and then pass them to `BytecodeRunner`:

```
public void run(Wei senderBalance, Long gasLimit, List<ToyAccount> additionalAccounts) {
    ...

    List<ToyAccount> accounts = new ArrayList<>();
    accounts.add(senderAccount);
    accounts.add(receiverAccount);
    accounts.addAll(additionalAccounts);

    toyExecutionEnvironmentV2 =
            .transactionProcessingResultValidator(
                TransactionProcessingResultValidator.EMPTY_VALIDATOR)
            .accounts(accounts)
            .zkTracerValidator(zkTracerValidator)
            .transaction(tx)
            .build();
    toyExecutionEnvironmentV2.run();
  }
```
In this way, the `ToyExecutionEnvironment` receives the accounts we want to be available and potentially containing some bytecode associated to them.

However, this may be not convenient when the same account is used to send more than one transaction within the same test, as the `nonce` has to be increased manually in test itself.
Ideally, we want the `ToyExecutionEnvironment` to take care of it for us. For this reason, we want the test to have the possibility to handle addresses (and potentially bytecode associated to them) only and let the `ToyExecutionEnvironment` do the rest of the work.

amkCha:
Closes #1590 

- have the world deal with everything in between (nonce updates, deployments, balance changes, ...)
==> change worldStateUpdater.commit() place in `testing/src/main/java/net/consensys/linea/testing/GeneralStateReferenceTestTools.java` 

- when defining the transactions only use Addresses to refer to accounts
The nonce is a private parameter in a Transaction so once the Transaction is built, it cannot be changed. What we can feed ToyMultiTransaction with is a list of ToyTransactionBuilders i/o ToyTransactions and then go
`final List<Transaction> transactions =
      ToyMultiTransaction.builder().build(toyTransactionBuilders, userAccount);`
We can use the Address as well vs the Account. It requires a search in the HashMap addressToAccount though and I'm not sure I see it adds more value - TBD


DavePearce:
Note that you are currently changing the commit for the constraints in this PR.  Check that's what you intended.  It could be that after updating to HEAD of arith-dev this will be resolved.

amkCha:
> Note that you are currently changing the commit for the constraints in this PR. Check that's what you intended. It could be that after updating to HEAD of arith-dev this will be resolved.

@DavePearce I'm checking, I'm not sure where I can find the commit change ? Also I'm a fork of `CREATE-tests` branch as well, not from `arith-dev`, don't know if it's relevant for this

OlivierBBB:
This PR is apparently building off of some PR of mine I guess `CREATE` related ?

testing/src/main/java/net/consensys/linea/testing/GeneralStateReferenceTestTools.java
```diff
@@ -126,6 +126,7 @@ public static void executeTest(
 
       transactionProcessingResultValidator.accept(transaction, result);
       zkTracerValidator.accept(tracer);
+      worldStateUpdater.commit();
```
OlivierBBB, line 129: This is I guess what allows you to extract the right `nonce` in the next transaction ?

testing/src/main/java/net/consensys/linea/testing/ToyMultiTransaction.java
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @return an instance of {@link List<Transaction>}
+     */
+    public List<Transaction> build(
+        ToyTransactionBuilder[] toyTxBuilders, ToyAccount senderAccount) {
+      long senderAccountNonce = senderAccount.getNonce();
+      List<Transaction> results = new ArrayList<>();
+      for (ToyTransactionBuilder toyTxBuilder : toyTxBuilders) {
+        results.add(toyTxBuilder.nonce(senderAccountNonce).build());
+        senderAccountNonce++;
```
OlivierBBB, line 42: Maybe stupid question but why do we raise the nonce manually ? There is the commit stuff that, IIUC, updates the state with every transaction. AFAICT the above will only work for multi transaction tests where the sender is the same in all transactions. (Which, to be fair, is currently our only use case.)

testing/src/main/java/net/consensys/linea/testing/GeneralStateReferenceTestTools.java
OlivierBBB: Now I'm confused: we do the `commit` at the end of `executeTest`. IIUC this would mean that for multi block tests we only `commit` once (after all blocks were processed.) Why do that instead of after every block as was previously the case ? Am I misunderstanding ?

OlivierBBB: I think I misunderstood: you are doing this inside of a for loop over all transactions in the `blockBody`.

OlivierBBB (approved):
LGTM

testing/src/main/java/net/consensys/linea/testing/ToyMultiTransaction.java
amkCha: Okay, so the way I understand it is : we raise the nonce manually because we craft all the transactions prior to broadcasting them. So they have to be crafted with the correct nonce as they are signed. Then the commit stuff, it updates the worldstate, but it doesn't update transaction - it's only going to update the nonce of the userAccount

In terms of use-case, was there another one than crafting transactions from the same account ? For instance you'd like to be able to craft N transactions with N different accounts easily ? I can add code for that

testing/src/main/java/net/consensys/linea/testing/GeneralStateReferenceTestTools.java
amkCha: Yes, we commit after every transaction so that the userAccount nonce is updated and can accept the following transaction properly and has the updated balance for instance

testing/src/main/java/net/consensys/linea/testing/ToyMultiTransaction.java
OlivierBBB: Not N / N, no. It think it's fine the way it is, we can elaborate on this if the need every arises.

PR #1637, opened 2024-12-10
Move plugins into a separate project

### LINKED ISSUE #1626: Separate plugins from base shared classes
organize the tracer to create an artifacts that only exposes the things that are needed by the sequencer ( ZKTracer and some config classes) and another one that includes the tracer plugins, so these will not be duplicated in case both the sequencer and the tracer artifacts are loaded

fab-10:
Move plugins to a separated project, so the sequencer can use the arthmetization lib without including the plugins, and avoid duplication.

Some rework was required, since the `pluginImplementation` configuration was not working with multiple project, so I reverted to use plain java plugin configurations.

Also the detection of libs already provided by Besu has been improved

fixes #1626


gradle.properties
```diff
@@ -1,4 +1,4 @@
-releaseVersion=0.8.0-rc8
+releaseVersion=0.8.0-rc8-local
```
Filter94: Is this a leftover or why is it called local now?

fab-10: a leftover, removing it

Filter94 (approved):

PR #1658, opened 2024-12-14
`PUSHVALUE` fix for exceptional `PUSH`es

OlivierBBB:
Ugly hack. Can be optimized to
- use `getPushValue` if and only if there is an exception
- use the old approach if there is no exception


DavePearce (commented):
LGTM

PR #1661, opened 2024-12-15
Various `SELFDESTRUCT` related fixes

OlivierBBB:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/halt/SelfdestructSection.java
```diff
@@ -168,37 +169,44 @@ public SelfdestructSection(Hub hub, MessageFrame frame) {
...
 
-    selfdestructorAccountAfter = selfdestructorAccountBefore.deepCopy().setBalanceToZero();
+    selfdestructorNew = selfdestructor.deepCopy().setBalanceToZero();
+
+    final boolean isDeployment = frame.getType() == MessageFrame.Type.CONTRACT_CREATION;
+    checkState(isDeployment == selfdestructor.deploymentStatus());
+    if (isDeployment) {
+      selfdestructorNew = selfdestructorNew.deploymentStatus(false);
+      selfdestructorNew.code(Bytecode.EMPTY);
+    }
```
OlivierBBB, line 177: Unexceptional `SELFDESTRUCT`'s in deployment contexts trigger an empty deployment. Thus we reset the byte code (from init code to `Bytes.EMPTY`.)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/account/AccountFragment.java
```diff
@@ -173,7 +174,8 @@ public void resolvePostTransaction(
         new EphemeralAccount(oldState.address(), oldState.deploymentNumber());
     if (effectiveSelfDestructMap.containsKey(ephemeralAccount)) {
       final int selfDestructTime = effectiveSelfDestructMap.get(ephemeralAccount);
-      markedForSelfDestruct = hubStamp > selfDestructTime;
+      markedForSelfDestruct =
+          domSubStampsSubFragment.domStamp() > MULTIPLIER___DOM_SUB_STAMPS * selfDestructTime;
```
OlivierBBB, line 178: This is a subtle point with respect to self-targeting `SELFDESTRUCT`'s. For the two "doing" account-rows the `HUB_STAMP` remains the same. We therefore **cannot** use the following definition

```java
markedForSelfDestruct    = hubStamp >  selfDestructTime; // won't work on the 2nd row
markedForSelfDestructNew = hubStamp >= selfDestructTime;
```

since the second time we inspect the account, this time as the "recipient", the flag ought to be `true` yet the parameters of this equation haven't changed. Instead we now differentiate according to the `DOM_STAMP`.

```java
markedForSelfDestruct    = domSubStampsSubFragment.domStamp() > MULTIPLIER___DOM_SUB_STAMPS * selfDestructTime;
markedForSelfDestructNew = hubStamp >= selfDestructTime;
```

This achieves the following result, for 

```
// self-targeting SELFDESTRUCT's
- row 1: hubStamp * MULTIPLIER___DOM_SUB_STAMPS + 0 => markedForSelfDestruct ≡ false
- row 2: hubStamp * MULTIPLIER___DOM_SUB_STAMPS + 1 => markedForSelfDestruct ≡ true
```



This is what we fix here

PR #1680, opened 2025-01-06
fix: update links get-started.md

youyyytrok:
Hi! I noticed that some of the links in the get-started.md file were broken, so I updated them to point to the correct resources. Now all links should lead to the appropriate pages, improving navigation and documentation usability.


letypequividelespoubelles:
Hi, 
Guidelines for Non-Code and other Trivial Contributions
Please keep in mind that we do not accept non-code contributions like fixing comments, typos or some other trivial fixes. Although we appreciate the extra help, managing lots of these small contributions is unfeasible, and puts extra pressure in our continuous delivery systems (running all tests, etc). Feel free to open an issue pointing to any of those errors, and we will batch them into a single change.

PR #1689, opened 2025-01-13
feat: go-corset front end integration

### LINKED ISSUE #1688: feat: `go-corset` front-end integration
The goal here is to replace the Rust `corset` tool entirely on the `go-corset` testing path.

DavePearce:


.github/workflows/gradle-ethereum-tests.yml
```diff
@@ -72,14 +72,15 @@ jobs:
...
-          CORSET_FLAGS: fields,expand,expand,expand
+          ZKEVM_BIN: "zkevm_for_reference_tests.go.bin"
 
       - name: Run General State Reference Tests
         run: GOMEMLIMIT=64GiB ./gradlew -Dblockchain=Ethereum referenceGeneralStateTests
         env:
           JAVA_OPTS: -Dorg.gradle.daemon=false
           CORSET_FLAGS: disable
-          GOCORSET_FLAGS: -wd --ansi-escapes=false --report --air
+          GOCORSET_FLAGS: -vw --ansi-escapes=false --report --air
```
FlorianHuc, line 82: is the option change wd to vw intentional?

.github/workflows/reference-blockchain-tests.yml
```diff
@@ -39,20 +39,17 @@ jobs:
...
+        uses: ./.github/actions/setup-environment
+        with:
+          go-corset: true
+        
       - name: Generate block chain reference tests
         run: ./gradlew :reference-tests:generateBlockchainReferenceTests -Dorg.gradle.caching=true
         env:
           JAVA_OPTS: -Dorg.gradle.daemon=false
-          CORSET_FLAGS: fields,expand,expand,expand
+          ZKEVM_BIN: "zkevm_for_reference_tests.go.bin"
```
FlorianHuc: i would expect that to generate the test we don't need this, do we ?

gradle/corset.gradle
```diff
@@ -19,38 +19,54 @@ tasks.register('corsetExists') {
...
+    commandLine 'make', ZKEVM_BIN        
     // store the output instead of printing to the console
     standardOutput = new ByteArrayOutputStream()
     ext.output = {
         return standardOutput.toString()
     }
 }
 
+// Allowing overriding the default zkevm.bin file.
+def ZKEVM_REFTESTS_BIN = System.getenv("ZKEVM_BIN") ?: "zkevm_for_reference_tests.bin"
```
FlorianHuc, line 54: if we want to allow to override ZKEVM_REFTESTS_BIN, why don't directly set the value of ZKEVM_REFTESTS_BIN (instead of setting ZKEVM_BIN)? I feel we could use ZKEVM_BIN only and remove ZKEVM_REFTESTS_BIN. no?

reference-tests/build.gradle
```diff
@@ -88,9 +89,10 @@ tasks.register('referenceGeneralStateTests', Test) {
   tasks.named("referenceBlockchainTests"),
 ].forEach {
   it.configure {
-    dependsOn buildReferenceTestsZkevmBin
-    environment.put("ZKEVM_BIN", "zkevm_for_reference_tests.bin")
-
+    def ZKEVM_REFTESTS_BIN = System.getenv("ZKEVM_BIN") ?: "zkevm_for_reference_tests.bin"
+    // Record zkevm.bin for CorsetValidator
+    environment.put("ZKEVM_BIN", ZKEVM_REFTESTS_BIN)
```
FlorianHuc: What about 
environment.put("ZKEVM_BIN", System.getenv("ZKEVM_BIN") ?: "zkevm_for_reference_tests.bin")

FlorianHuc (approved):

.github/workflows/gradle-ethereum-tests.yml
DavePearce: Yes, it is.  Its a minor change in the CLI for `go-corset`.

.github/workflows/reference-blockchain-tests.yml
DavePearce: I can double check that ... but for some reason I think I ended up needing it.

reference-tests/build.gradle
DavePearce: Ok, I can do that.  I did have it like that originally, it just felt a bit neater this way.

gradle/corset.gradle
DavePearce: Well, its a choice and I did consider this option.  The problem is that it will mean running the reference tests from the command-line and/or via intelliJ will require the `ZKEVM_BIN` environment variable being set.  I think that will cause some problems when others need to debug failing reference tests.

.github/workflows/reference-blockchain-tests.yml
DavePearce: Yeah, so I agree it can be dropped --- done.

PR #1690, opened 2025-01-13
Feat/blockdata gas limit name refacto

amkCha:
Regenerate trace files on tests launch to dynamically account for a different Min and Max gas limit between Ethereum and Linea

Paired with `BLOCKHASH-redesign` /linea-constraints branch

TBC
With Ethereum Gas limits
- StateRefTests
- BlockchainRefTests

With Linea Gas limits
- fastReplayTests


OlivierBBB:
In the "With Linea Gas limits" section we also need to include the "unit tests" that we wrote.

PR #1691, opened 2025-01-13
CHANGED - use github actions concurrency to reduce outdated runs

waskow-consensys:
Following best practice 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs

for example if someone commits to a branch 3 times, 3 actions are run. This would cancel previous actions that are now stale. Saving Time and Money :)


DavePearce (approved):
LGTM

PR #1699, opened 2025-01-14
Feat/blockdata unique gas limit constant

amkCha:
Regenerate trace files on tests launch to dynamically account for a different Min and Max gas limit between Ethereum and Linea

With Ethereum Gas limits
    StateRefTests
    BlockchainRefTests

With Linea Gas limits
    FastReplayTests
    Unit tests


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/BlockdataOperation.java
```diff
@@ -178,11 +175,11 @@ private void handleGasLimit() {
 
     // row i
     // comparison to minimum
-    wcpCallToGEQ(0, data, EWord.of(LINEA_GAS_LIMIT_MINIMUM));
+    wcpCallToGEQ(0, data, EWord.of(GAS_LIMIT_MINIMUM));
 
     // row i + 1
     // comparison to maximum
-    wcpCallToLEQ(1, data, EWord.of(Bytes.ofUnsignedLong(LINEA_GAS_LIMIT_MAXIMUM)));
+    wcpCallToLEQ(1, data, EWord.of(Bytes.ofUnsignedLong(GAS_LIMIT_MAXIMUM)));
```
OlivierBBB, line 182: 👍

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/BlockdataOperation.java
```diff
@@ -60,7 +57,7 @@ public class BlockdataOperation extends ModuleOperation {
 
   private final boolean firstBlockInConflation;
   private final int ctMax;
-  @EqualsAndHashCode.Include @Getter private final OpCode opCode;
```
DavePearce, line 63: This change looks odd, and potentially unexpected.  Is there a specific reason we wanted to remove this?

DavePearce (commented):
Basically, I think this looks good.  Just unsure about the `EqualsAndHashCode.Include` for `BlockdataOperation`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/BlockdataOperation.java
OlivierBBB: I don't see that on my local branch

image

But it's strange that there should be an `@EqualsAndHashCode.Include` in the first place. We don't need to compare `BlockdataOperation`. And if we did we would have to include some extra identifier (e.g. the `blockHeader`)

DavePearce: Ok, well if you believe that's safe ... then I'm happy to merge.

DavePearce (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/BlockdataOperation.java
amkCha: Yes, it's a copy paste to get the OpCode type that I got from another file. I forgot to remove the annotations, I don't see the need there

PR #1706, opened 2025-01-15
feat: update `release` action to install `corset`

### LINKED ISSUE #1705: fix: release action requires `corset`
The `release.yml` action did not previously require `corset` to be installed.  However, since we now use `corset` to rebuild the `Trace.java` files, then `corset` has to be available to do that part.  Therefore, we need to install `corset` as part of the release process.

DavePearce:
This updates the release action so that it installs `corset`.  This is required now, since we use `corset` to rebuild the `Trace.java` files from scratch.


OlivierBBB (approved):
LGTM

PR #1716, opened 2025-01-16
Fix Java jars publication and remove redundant upload to Cloudsmith

### LINKED ISSUE #1715: Fix Java jars publication
Correctly publish all the built jars to Cloudsmith, and remove the upload of the distribution to Cloudsmith since it is already uploaded on Github

fab-10:
fixes #1715


DavePearce (approved):
LGTM

PR #1711, opened 2025-01-15
fix: use both traceStartBlock calls where appropriate

lu-pinto:
I rolled back usage of `traceStartBlock(BlockHeader, BlockBody, Address)` where appropriate because I feel we shouldn't be changing `TraceServiceImpl` on the Besu side just to fix the linea tracer. This avoids an inconsistency between `traceStartBlock` and `traceEndBlock` and also opening a can of worms going through yet another round of consensus as to why it was done like that in the first place.

I see that your tracer implementations were implementing both cases at the same time, so it's hard to see which is definitely being used, but given that we have to modify `TraceServiceImpl` to make things work I would say `traceStartBlock(BlockHeader, BlockBody)` was definitely being used.

Also with this you don't need another release from linea-besu to make the tracer work as both `traceStartBlock(BlockHeader, BlockBody, Address)` and `traceStartBlock(ProcessableBlockHeader, Address)` already exist.

PR #1722, opened 2025-01-20
Update Linea Besu to 25.1-develop-448d1a9

### LINKED ISSUE #1721: Update Linea Besu to 25.1-develop-448d1a9

fab-10:
fixes #1721

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
+++ b/arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
@@ -69,7 +69,8 @@ public class Reaper {
   /** Collect the account address read / written by the current transaction */
   private AddressReaper txAddresses = null;
 
-  public void enterBlock(final BlockHeader header, final BlockBody body, final Address miningBeneficiary) {
+  public void enterBlock(
+      final BlockHeader header, final BlockBody body, final Address miningBeneficiary) {
     this.blocks.add(
         BlockSnapshot.of((org.hyperledger.besu.ethereum.core.BlockHeader) header, body));
     this.conflationAddresses.touch(header.getCoinbase()); // ToDo: replace with miningBeneficiary?
--- a/arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
+++ b/arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
@@ -69,10 +69,10 @@ public class Reaper {
   /** Collect the account address read / written by the current transaction */
   private AddressReaper txAddresses = null;
 
-  public void enterBlock(final BlockHeader header, final BlockBody body) {
+  public void enterBlock(final BlockHeader header, final BlockBody body, final Address miningBeneficiary) {
     this.blocks.add(
         BlockSnapshot.of((org.hyperledger.besu.ethereum.core.BlockHeader) header, body));
-    this.conflationAddresses.touch(header.getCoinbase());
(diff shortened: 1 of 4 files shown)


arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
```diff
@@ -69,10 +69,10 @@ public class Reaper {
   /** Collect the account address read / written by the current transaction */
   private AddressReaper txAddresses = null;
 
-  public void enterBlock(final BlockHeader header, final BlockBody body) {
+  public void enterBlock(final BlockHeader header, final BlockBody body, final Address miningBeneficiary) {
     this.blocks.add(
         BlockSnapshot.of((org.hyperledger.besu.ethereum.core.BlockHeader) header, body));
-    this.conflationAddresses.touch(header.getCoinbase());
+    this.conflationAddresses.touch(header.getCoinbase()); // ToDo: replace with miningBeneficiary?
```
fab-10: @lu-pinto @OlivierBBB please update this code accordingly to the new version of `traceStartBlock`

lu-pinto: Good point, I'm not sure if that code is running with Clique, only @OlivierBBB would know.
We could still add it to my PR https://github.com/Consensys/linea-tracer/pull/1711/files. But essentially I added the `miningBeneficiary` to the `BlockCapturer` because it was needed for overriding the interface but not the `Reaper`.

lu-pinto: But my 2 cents - since this was working fine before without any NPEs it's unlikely that they need the proper `miningBeneficiary` there now in this case. It's perfectly legit to have access to the parameter but not use it if not required.

OlivierBBB (approved):
I take it the `0` is the baseline gas in the `transactionIntrinsicGasCost` method. LGTM.

arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
OlivierBBB: I think it makes sense to use the `miningBeneficiary`. If only because
- we have no use for the `header.getCoinbase()` on neither SEPOLIA nor MAINNET
- this reaper is, I take it, responsible for extracting the data used to create replay tests, and replay tests are typically where the divergence between the `miningBeneficiary` and `header.getCoinbase()` appeared

I believe that the above will work regardless, as we will touch the correct coinbase address in any case at the latest when it collects its gas fees. Here we are, seemingly, taking a snapshot of the `0x00..00` address for no reason.

OlivierBBB (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/blockcapture/reapers/Reaper.java
```diff
@@ -69,10 +69,11 @@ public class Reaper {
   /** Collect the account address read / written by the current transaction */
   private AddressReaper txAddresses = null;
 
-  public void enterBlock(final BlockHeader header, final BlockBody body) {
+  public void enterBlock(
+      final BlockHeader header, final BlockBody body, final Address miningBeneficiary) {
     this.blocks.add(
         BlockSnapshot.of((org.hyperledger.besu.ethereum.core.BlockHeader) header, body));
-    this.conflationAddresses.touch(header.getCoinbase());
+    this.conflationAddresses.touch(miningBeneficiary); // ToDo: replace with miningBeneficiary?
```
fab-10:
```suggestion
    this.conflationAddresses.touch(miningBeneficiary);
```

PR #1729, opened 2025-01-22
fix wcp line count for operations coming from block data

### LINKED ISSUE #1728: BLOCKDATA calls to WCP are not counted for trace file limit

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
```diff
@@ -69,42 +66,47 @@ public String moduleKey() {
...
   public void traceStartConflation(final long blockCount) {
-    wcp.additionalRows.add(TIMESTAMP_BYTESIZE); // TODO: check
+    wcp.additionalRows.add(
+        LLARGE // for COINBASE
+            + 6
+            + 6 // for TIMESTAMP
+            + 1
+            + 6 // for NUMBER
+            + 1 // for DIFFICULTY
+            + Bytes.minimalBytes(GAS_LIMIT_MAXIMUM).size() * 4 // for GASLIMIT
```
amkCha, line 76: [for me to understand] why does this one depend on a variable ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
```diff
@@ -69,42 +66,47 @@ public String moduleKey() {
...
-                wcp,
-                euc,
-                chainId,
-                opCode,
-                firstBlockNumber);
-        operations.addLast(operation);
-        // Increase counter to track where we are in the conflation
-        traceCounter++;
-      }
+    if (operations.isEmpty()) {
```
amkCha, line 87: :+1:

amkCha (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: when we call EQ in WCP, it's a One Line Operation, when we call LT, LEQ, GT, GEQ, it takes as many rows as the max byte size of the arguments (capped to LLARGE = 16). I could have write 6 instead of Bytes.minimalBytes(GAS_LIMIT_MAXIMUM).size(), but this way we remember where it comes from. I could have done same for others, but there was no named constant

lorenzogentile404 (approved):

PR #1732, opened 2025-01-23
fix: daily blockchain tests

### LINKED ISSUE #1723: fix: daily blockchain tests
The goal here is to try and figure out what is holding back the daily blockchain tests.

DavePearce:
This fixes the issues running the daily blockchain tests, by doing two things:

1. Updating to the latest version of `go-corset`, which includes some minor optimisations that help reduce cost.
2. Adding a batching parameter for `go-corset` (`-b 1024`) which sets the batch size to max `1024`.  This throttles concurrency to some extent, and reduces memory pressure.

Overall, these changes appear to allow the [daily blockchain tests to pass](https://github.com/Consensys/linea-tracer/actions/runs/12922706274), and hopefully they will consistently pass going forward.

### DIFF
--- a/.github/actions/setup-go-corset/action.yml
+++ b/.github/actions/setup-go-corset/action.yml
@@ -12,4 +12,4 @@ runs:
 
     - name: Install Go Corset
       shell: bash
-      run: go install github.com/consensys/go-corset/cmd/go-corset@77d304f
+      run: go install github.com/consensys/go-corset/cmd/go-corset@fe70003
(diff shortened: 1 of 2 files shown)


OlivierBBB (approved):

OlivierBBB:
Approved. Where does the value 1024 come from and do you know how many batches the previously unthrottled implementation would use ? Also are these batches batches of columns that corset computes or something else ?

letypequividelespoubelles (approved):

DavePearce:
The 1024 figure is just one that seems to work well. I played around with a few. Without a batch size specified it tries to compute as many columns in one go as it can. Given that we have max ~2300
Columns, of which some are input columns and don't need to be computed ... i guess the largest batch probably isn't much larger than 1024. But the difference is obviously enough to tip it back into the green 🤔

DavePearce:
Actually forget that. Its not columns causing the problem but constraints I guess. Which means it might be trying to do all 9000 in one go!! (Since constraints don't have dependencies between themselves)

PR #1733, opened 2025-01-23
clean: remove dead code

### LINKED ISSUE #1449: Very suspicious
https://github.com/Consensys/linea-tracer/blob/cab035d5eef415af7cb0c007bdd569e92aa4ffc5/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/transients/OperationAncillaries.java#L220-L227

It seems that at least the check leading to an early return should be wrt the **negation** of the condition.

letypequividelespoubelles:


OlivierBBB (approved):
I take it these methods were obsolete and not used in the repo anywhere ?

PR #1748, opened 2025-01-28
1660 tests for sload and sstore where one of these instructions has multiple revert reasons

### LINKED ISSUE #1660: Tests for `SLOAD` and `SSTORE` where one of these instructions has multiple revert reasons
We need to write tests where some execution context, at depth e.g. 3 or 4 or more, performs `SLOAD`'s and `SSTORE`'s that will get reverted by _several_ parent contexts reverting. We should even do a test where we have a deep tower of `CALL`'s which alternate in whether or not they `SELF_REVERT` and they all modify the same storage.

Image

lorenzogentile404:


OlivierBBB:
Hey @lorenzogentile404 , the [original issue](https://github.com/Consensys/linea-tracer/issues/1660) describes the case where it's always the same smart contract whose storage changes. I.e. it's performing self calls. From what I saw on our calls the tests don't directly address the issue. We can definitively keep the `A`, `B`, `C`, `D`, `E` version, but the original issue asks for self calls of some smart contract `smcA`.

We should implement two versions
- root reverts
- root does not revert

Where whether a revert takes place depends on e.g. the parity of the integer in call data.

OlivierBBB (approved):
I'm approving albeit a little blindly.

PR #1751, opened 2025-01-29
fix: blockdata line counting & traceEndBlock trigger in ref tests

### LINKED ISSUE #1750: BLOCKDATA reference test constraint issue
https://github.com/Consensys/linea-tracer/actions/runs/13019484899/job/36316653290

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
```diff
@@ -113,7 +115,7 @@ public void popTransaction() {}
 
   @Override
   public int lineCount() {
-    final int numberOfBlock = (operations.size() / opCodes.length);
+    final int numberOfBlock = (operations.size() / opCodes.length) + (conflationFinished ? 0 : 1);
```
letypequividelespoubelles, line 118: fix: as we add blockData operations at the end of the block, while we are still building it, we must add one block to the counter

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -425,7 +427,6 @@ public static void executeTest(final BlockchainReferenceTestCaseSpec spec) {
                 importResult.isImported(),
                 candidateBlock.isValid());
 
-        zkTracer.traceEndBlock(block.getHeader(), block.getBody());
```
letypequividelespoubelles, line 428: fix: we were calling traceEndBlock twice per block

lorenzogentile404 (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
amkCha: Ok so we did need that boolean in the end, and now we know why :+1:

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -412,6 +412,8 @@ public static void executeTest(final BlockchainReferenceTestCaseSpec spec) {
                 ? HeaderValidationMode.LIGHT
                 : HeaderValidationMode.FULL;
 
+        // Note: somehow this function is calling traceEndBlock through
+        // blockValidator.validateAndProcessBlock
```
amkCha, line 416: But it's a method in the besu codebase no ? How can it call a method in our plugin ? :thinking:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: well, the boolean you had was different (and not used):
It was set to false after adding the blockDataOperation, and set to true at the begining of every transaction. It's only usage was, at traceEndTransaction to add the op if true ... which is always the case because we don't have empty block. It could have been usefull if we had empty block we don't want to trace, but it's not the case (Linea doesn't produce empty block)

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
letypequividelespoubelles: What's more surprising is that it doesn't call traceStartBlock. But our ZkTracer is called by Besu's API for tracePreOpcode, traceStartTransaction, etc. If it wasn't the case, we wouldn't be able to be aware of the execution of the evm.

letypequividelespoubelles: But you right, I should have a look in Besu's code to understand why it's not symmetric between traceStartBlock and traceEndBlock

PR #1754, opened 2025-01-30
Low gas stipend precompile calls

### LINKED ISSUE #1730: Low gas stipend precompile calls
For every precompile, write a test where the `ROOT` performs a call to the said precompile, providing a `gas` parameter that is  insufficient. The transaction should have a huge amount of gas (e.g., `1_000_000`), but we are interested in the failure due to out of gas in the precompile (that is insufficient deliberately).

Perform this test both with zero `value` and non-zero value as an argument of the `CALL` to the precompile.

An advanced version of this test is described in https://github.com/Consensys/linea-tracer/issues/1153.

lorenzogentile404:


arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,100 @@
...
+
+    // Common program for all precompile calls
+    program
+        .push(isZeroArgument ? 0 : 1) // retSize
+        .push(isZeroArgument ? 0 : 1) // retOffset
+        .push(argsSize) // argsSize
+        .push(isZeroArgument ? 0 : 1) // argsOffset
+        .push(0) // value
+        .push(precompileAddress) // address
+        .push(1) // gas, that is deliberately insufficient
```
OlivierBBB: I would love for this to be parametric in the following sense. Have some `enum`'s
```rust
valueParameter = ZERO, ONE
gasParameter = ZERO, ONE, COST_MINUS_ONE, COST, COST_PLUS_ONE
```
Remember that in case the value is nonzero the precompile gets a bonus of `2_300` gas.

OlivierBBB: Also in order not to have to compute the memory expansion cost you can pre-populate memory with values. Even just pre-populating e.g. via
```rust
PUSH1 0
PUSH1 224
MSTORE
```

OlivierBBB: I have methods (in https://github.com/Consensys/linea-tracer/pull/1778) to do that which would be useful for
```
SHA2, RIPEMD, IDENTITY
```
For EC precompiles just having 0's would also work.

OlivierBBB: In the `value = ONE` case, if we want to offset the `2_300` gas stipend, you have to make the instructions costly enough:
- `ECRECOVER` always costs `3_000` so it's straight forward
- `ECMUL` cost `6_000` so it's straight forward
- `ECPAIRING` costs at least `45_000` so it's straight forward
- long enough for `SHA2`, `RIPEMD`, `IDENTITY`
- `BLAKE2f` the precompile cost is in memory and again it's simple enough to deduce the right amount of gas

**Note.** We omitted
- `ECADD` has a fixed cost of `150` so it's impossible
- `MODEXP` would be complex, so we can for now exclude it from this and add a new issue to rectify it,

OlivierBBB (changes requested):
More parametric and if necessary pre-populate memory to simplify computations.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+    // - Populate the memory with a large enough number of words for SHA256, RIPEMD160, and ID
+    //   to have precompileCost > gasBonus.
+    final int value = valueParameter.isZeroArgument() ? 0 : 1;
+    final int argsSize; // depends on the called precompile
+    final int argsOffset = valueParameter.isZeroArgument() ? 0 : 1;
+    final int retSize = valueParameter.isZeroArgument() ? 0 : 1;
+    final int retOffset = valueParameter.isZeroArgument() ? 0 : 1;
+
+    // BLAKE2F specific parameters
+    int rFirstByte = valueParameter.isZeroArgument() ? 0 : 0x12;
```
OlivierBBB: I have a slight preference for `rLeadingByte`.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+    } else if (precompileAddress == ALTBN128_PAIRING) {
+      // EC_PAIRING specific parameters
+      argsSize = 192;
+    } else if ((precompileAddress == SHA256
+            || precompileAddress == RIPEMD160
+            || precompileAddress == ID)
+        && valueParameter.isNonZeroArgument()) {
+      // SHA256, RIPEMD160, and ID specific parameters
+      int nWords = 1024;
+      argsSize = nWords * WORD_SIZE; // This guarantees that precompileCost > gasBonus
```
OlivierBBB: The spec calls it $G_\text{callstipend}$, we should thus call it `callStipend` rather than `gasBonus`.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+
+    // In order to actually trigger the insufficient we need to:
+    // - Set a specific args size for BLAKE2F and EC_PAIRING
+    // - Set the r value of BLAKE2F to have precompileCost > gasBonus
+    // - Populate the memory with a large enough number of words for SHA256, RIPEMD160, and ID
+    //   to have precompileCost > gasBonus.
+    final int value = valueParameter.isZeroArgument() ? 0 : 1;
+    final int argsSize; // depends on the called precompile
+    final int argsOffset = valueParameter.isZeroArgument() ? 0 : 1;
+    final int retSize = valueParameter.isZeroArgument() ? 0 : 1;
```
OlivierBBB: I would suggest setting `retSize` to something nonzero always:
```rust
ECRECOVER, SHA2, RIPEMD, ECPAIRING -> WORD_SIZE
ECADD, ECMUL, BLAKE2f -> 2 * WORD_SIZE
MODEXP -> mbs
IDENTITY -> argSize
```

I would also suggest setting `retOffset` to something nontrivial always, e.g. `13`. Making these values dependent on the `valueParameter` doesn't make much sense. The `13` is there just to force overlap in the **call data range** and **return at range**.

What I've done in other PR's (e.g. the one on precompiles that I'm currently working on) is to have two enums which takes on the following values:
```rust
// RelativeRangePositions, to say whether we want our call data ranges and return at ranges to overlap or be disjoint
DISJOINT,
OVERLAP

// WordAlignment, to say whether we want our offsets to be ≡ 0 [16] or not.
ALIGNED
MISALIGNED
```
This could be used to then manufacture some offsets. But in the present PR I'm not convinced this makes much sense. We really care about the pricing. So my suggestion would be to just impose
```java
argOffset = 0;
retOffset = 13;
```
systematically.

OlivierBBB: When you include the MODEXP case in this you will in any case require some values for `bbs`, `ebs` and `mbs`.

OlivierBBB: Also note that what you call `retSize` / `retOffset` / `argSize` / `argOffset` we typically call `rac`, `rao`, `cds`, `cdo`.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+    // Note that we exclude the case of MODEXP as it is treated in a separate test
+    // and the case of ALTBN128_ADD as it has a fixed gas cost of 150.
+    if (valueParameter.isNonZeroArgument()
+        && (gasParameter == GasParameter.COST_MINUS_ONE
+            || gasParameter == GasParameter.COST
+            || gasParameter == GasParameter.COST_PLUS_ONE)
+        && !precompileAddress.equals(ALTBN128_ADD)
+        && !precompileAddress.equals(MODEXP)) {
+      final int gasBonus = 2300;
+      gas -= gasBonus;
```
OlivierBBB: Naming + 2300 is in the `GlobalConstants`.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+  private static int getPrecompileCost(Address precompileAddress, int argsSize, int r) {
+    final int precompileCost;
+    if (precompileAddress.equals(ECREC)) {
+      precompileCost = 3000;
+    } else if (precompileAddress.equals(SHA256)) {
+      precompileCost = (5 + (argsSize + 31) / 32) * 12;
+    } else if (precompileAddress.equals(RIPEMD160)) {
+      precompileCost = (5 + (argsSize + 31) / 32) * 120;
+    } else if (precompileAddress.equals(ID)) {
+      precompileCost = (5 + (argsSize + 31) / 32) * 3;
```
OlivierBBB: 31 and 32 are in the `GlobalConstants` (`WORD_SIZE` and `WORD_SIZE_MO` or so.)

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+    } else if (precompileAddress.equals(MODEXP)) {
+      precompileCost = 200;
+    } else if (precompileAddress.equals(ALTBN128_ADD)) {
+      precompileCost = 150;
+    } else if (precompileAddress.equals(ALTBN128_MUL)) {
+      precompileCost = 6000;
+    } else if (precompileAddress.equals(ALTBN128_PAIRING)) {
+      precompileCost = 45000 + 34000 * (argsSize / 192);
```
OlivierBBB: A lot of these constants are also in the `GlobalConstants`.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+      program
+          .push(rFirstByte) // For simplicity, we only set the first byte of r
+          .push(argsOffset + 2) // offset
+          // Writing rFirstByte at this offset
+          // allows to have r = 0x00000000 or r = 0x00001200
+          .op(OpCode.MSTORE8);
+      argsSize = 213;
+    } else if (precompileAddress == ALTBN128_PAIRING) {
+      // EC_PAIRING specific parameters
+      argsSize = 192;
```
OlivierBBB: `192` and `213` are constants in the `GlobalConstants`.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,260 @@
...
+            || precompileAddress == ID)
+        && valueParameter.isNonZeroArgument()) {
+      // SHA256, RIPEMD160, and ID specific parameters
+      int nWords = 1024;
+      argsSize = nWords * WORD_SIZE; // This guarantees that precompileCost > gasBonus
+      populateMemory(program, nWords, argsOffset);
+    } else {
+      // Default case
+      argsSize = valueParameter.isZeroArgument() ? 0 : 1;
+    }
```
OlivierBBB, line 152: I would suggest extracting this method for setting `argSize` etc ... into its own stand alone method. This may require making these parameters fields of the class.

OlivierBBB (changes requested):
See comments

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/PrecompileUtils.java
```diff
@@ -0,0 +1,128 @@
...
+    return GAS_CONST_RIPEMD + words * GAS_CONST_RIPEMD_WORD;
+  }
+
+  private static int getIDCost(int cds) {
+    final int words = (cds + WORD_SIZE_MO) / WORD_SIZE;
+    return GAS_CONST_IDENTITY + words * GAS_CONST_IDENTITY_WORD;
+  }
+
+  static int getMODEXPCost(int bbs, int mbs, BigInteger exponentLog) {
+    final int words = (Math.max(bbs, mbs) + 7) / 8;
```
OlivierBBB: I wouldn't call that `words`. You are computing how many `long`'s (8 byte integers) make up the largest among base and modulus ... `words` should be reserved for 32-byte chunks. In any case it seems a different word has to be used.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/PrecompileUtils.java
```diff
@@ -0,0 +1,128 @@
...
+    final int fOfMax = words * words;
+    final BigInteger bigNumerator =
+        BigInteger.valueOf(fOfMax).multiply(exponentLog.max(BigInteger.ONE));
```
OlivierBBB: Why is it that we use `BigInteger`'s here ? With our limitations we have that
- `fOfMax` is at most `64 * 64 = 4096`
- `exponentLog` is at most `256 + 8 * (512 - 32) = 4096`
- `bigQuotient` is at most `5592405`

OlivierBBB: It seems in any case that an `int` should suffice, and that `long`'s would be plenty.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+    final int returnAtOffset = 13;
+
+    // BLAKE2F specific parameters
+    final int rLeadingByte = argumentCase.isZeroCase() ? 0 : 0x12;
+    final int r = rLeadingByte << 8;
+
+    // MODEXP specific parameters
+    final int bbs = modexpCostGT200 ? 0x01 : 0x02;
+    final int ebs = modexpCostGT200 ? 0x06 : 0x03;
+    final int mbs = modexpCostGT200 ? 0x19 : 0x04;
```
OlivierBBB: It seems surprising that these computations are carried out regardless of the precompile, rather than within the relevant `case`.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+        .op(OpCode.MSTORE)
+        .push(ebsPadded)
+        .push(32) // offset
+        .op(OpCode.MSTORE)
+        .push(mbsPadded)
+        .push(64) // offset
+        .op(OpCode.MSTORE)
+        .push(bemPadded)
+        .push(96) // offset
+        .op(OpCode.MSTORE);
```
OlivierBBB: This isn't the right way to do it. The issue is that you are confined to testing short data because you can only `MSTORE` a 32B integer. So `bemPadded` has to be at most 32 bytes long. However it seems that your own scenarios violate this condition, e.g. when `modexpCostGT200 = true` you get
```
bbs + ebs + mbs = 0x26 = 38 > 32
```

OlivierBBB: You should use the technique I sent you guys over slack and which you'll find in my precompile PR. This avoids having to prepare data and `MSTORE`'ing it. You just `EXTCODECOPY` the stuff right into memory from an account that's been prepared to contain exactly the string you want.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+      callDataSize = PRC_BLAKE2F_SIZE;
+      prepareBlake2F(program, rLeadingByte, callDataOffset);
+    } else if (precompileAddress == ALTBN128_PAIRING) {
+      callDataSize = PRC_ECPAIRING_SIZE;
+    } else if ((precompileAddress == SHA256
+            || precompileAddress == RIPEMD160
+            || precompileAddress == ID)
+        && argumentCase.isNonZeroCase()) {
+      final int nWords = 1024;
+      callDataSize = nWords * WORD_SIZE; // This guarantees that precompileCost > gasBonus
```
OlivierBBB: We use the term `callStipend` rather than `gasBonus`

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+      prepareSha256Ripemd160Id(program, nWords, callDataOffset);
+    } else if (precompileAddress == MODEXP) {
+      callDataSize = 96 + bbs + ebs + mbs;
+      exponentLog = prepareModexp(bbs, mbs, ebs, callDataSize, program);
+    } else {
+      // ECADD, ECMUL, ECRECOVER cases
+      callDataSize = argumentCase.isZeroCase() ? 0 : 1;
```
OlivierBBB: I don't understand this: why would we want a CDS of 0 or 1 ? Genuine question, there may be some reason that I'm not getting.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+            Bytes.concatenate(bbsPadded, ebsPadded, mbsPadded, bemPadded),
+            BigInteger.valueOf(callDataSize),
+            BigInteger.valueOf(bbs),
+            BigInteger.valueOf(ebs),
+            BigInteger.valueOf(mbs))
+        .max(BigInteger.ONE);
+  }
+
+  /**
+   * Computes the returnAtCapacity based on the precompile address, and callDataSize in the case of
```
OlivierBBB, line 282: `ReturnAtCapacity` isn't something one computes, it's something one is free to set as one pleases in a CALL-type instruction. It seems to me you are computing the `ReturnDataSize` of (successful) calls to the various precompiles. These aren't up for debate, they are indeed determined by the precompile.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+   * ID.
+   *
+   * @param precompileAddress the address of the precompile contract.
+   * @param callDataSize the call data size. Beyond the case of ID, this value is ignored.
+   * @param mbs the modulo byte size. Beyond the case of MODEXP, this value is ignored.
+   * @return the computed return rac.
+   */
+  private static int getReturnAtCapacity(Address precompileAddress, int callDataSize, int mbs) {
```
OlivierBBB: `ReturnAtCapacity` isn't something one computes, it's something one is free to set as one pleases in a CALL-type instruction. It seems to me you are computing the `ReturnDataSize` of (successful) calls to the various precompiles. These aren't up for debate, they are indeed determined by the precompile.

OlivierBBB: You are free to use `r@c` = the predicted `rds`, but the method should bear a different name regardless.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class LowGasStipendPrecompileCallTests {
+
+  // Enums for the different testing scenarios
+  enum ArgumentCase {
+    ZERO,
+    NON_ZERO;
```
OlivierBBB, line 62: What is the `Argument` you are referencing in the `ArgumentCase` name ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/OobOperation.java
```diff
@@ -487,6 +458,44 @@ public void populateColumnsForPrecompile(MessageFrame frame) {
     }
   }
 
+  // Support method for MODEXP
+  public static BigInteger computeExponentLog(
```
OlivierBBB: It's surprising that this should return a `BigInteger`. As if any of `bbs` / `ebs` / `mbs` is large we `throw` anyway, and the end result is going to be at most `8 * 512`. Can't we use some `log2` method for (unsigned) `long`s or so ?

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+  enum GasCase {
+    ZERO,
+    ONE,
+    COST_MINUS_ONE,
+    COST,
+    COST_PLUS_ONE;
+  }
+
+  /**
+   * Parameterized test for low gas stipend precompile call.
```
OlivierBBB: I would add to the doc string something like this.
```
In this family of tests we call every precompile with hand crafted gas parameters in the CALL. The frame itself is provided with ample gas, but we choose the gas parameter to be equal to the execution cost of the precompile (up to -1, 0, +1.)

This allows us to deliberately trigger failures in the precompiles due to insufficient gas and test that, on the contrary, calls to precompiles with *exactly* the right amount of gas are successful (given that other conditions are met in terms of the contents of the call data.)
```

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
OlivierBBB: Also note that this is where you get the
```
0x01 + 0x06 + 0x19 = 0x26 = 38 > 32
```
issue mentioned below.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+    final int returnAtCapacity = getReturnAtCapacity(precompileAddress, callDataSize, mbs);
+
+    // Compute the precompile cost
+    final int precompileCost =
+        getPrecompileCost(precompileAddress, callDataSize, bbs, mbs, exponentLog, r);
+
+    // Compute the gas stipend in the different testing scenarios
+    int gas = getGas(gasCase, precompileCost);
+
+    // In case funds are sent to the precompile contract (argumentCase == NON_ZERO)
```
OlivierBBB: So `ArgumentCase` refers to `value` I take it ? In that case why not call it `ValueParameter` or something similar ?

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+    // a gas stipend of 2300 is added to the transaction.
```
OlivierBBB: It's added to the precompile frame's initial gas.

OlivierBBB: You can (and should ?) provide the _transaction_ with essentially limitless gas, e.g. `61_000_000`. The deduction you are talking about applies to the `gasParameter` of the CALL instruction targeting the precompile.

**Edit.** The `1_000_000` you provide to the transaction is plenty.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
```diff
@@ -0,0 +1,327 @@
...
+    // We now deduce that gas stipend from the gas given to the transaction to trigger
+    // insufficient gas for the precompile call in the non-trivial cases (COST_MINUS_ONE, COST,
+    // COST_PLUS_ONE).
+    // Note that we exclude the case of MODEXP as it is treated in a separate test
+    // and the case of ECADD as it has a fixed gas cost of 150.
```
OlivierBBB: `... as it has a fixed gas cost of 150 < 2_300.`

OlivierBBB (changes requested):
Several changes are in place. Otherwise looks good to me.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
lorenzogentile404: Indeed, it is not that interesting from a testing perspective. I'll just use 1 or some arbitrary value.

arithmetization/src/test/java/net/consensys/linea/zktracer/precompiles/LowGasStipendPrecompileCallTests.java
lorenzogentile404: I'll make the method more general, however the sum is 0x20 = 32.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/OobOperation.java
```diff
@@ -487,6 +464,28 @@ public void populateColumnsForPrecompile(MessageFrame frame) {
...
+        doublePaddedCallData.slice(96 + bbs, min(ebs, 32)).toUnsignedBigInteger();
+
+    if (ebs <= 32 && leadingBytesOfExponent.signum() == 0) {
+      return 0;
+    } else if (ebs <= 32 && leadingBytesOfExponent.signum() != 0) {
+      return log2(leadingBytesOfExponent, RoundingMode.FLOOR);
+    } else if (ebs > 32 && leadingBytesOfExponent.signum() != 0) {
+      return 8 * (ebs - 32) + log2(leadingBytesOfExponent, RoundingMode.FLOOR);
+    } else {
+      return 8 * (ebs - 32);
```
OlivierBBB, line 485: Re-reading this makes me wonder whether we have tests where we deliberately set the `leadingBytesOfExponent` to zero ...

OlivierBBB (approved):
LGTM

PR #1757, opened 2025-01-30
Reset `linea-constraints` back a few commits

### LINKED ISSUE #1756: chore: check defensive padding is working
This is a simple test to check that defensive padding is working.  Specifically, the approach is to reset `linea-constraints` to a known place where it will break with defensive padding and then ... check that it does break.

DavePearce:
This takes us back to before the fix which was spotted with manual padding, and should be automatically spotted with defensive padding.

**DO NOT MERGE EVER**


DavePearce:
This test was passed --- defensive padding did indeed find the problem.

PR #1758, opened 2025-01-31
test: tests masking of JUMPDEST under PUSHX

### LINKED ISSUE #309: Add tests to ROM module
- Add unit tests where
  - [x] the bytecode is a sequence of 32 consecutive pushes (PUSH1, PUSH2, ... PUSH32) and confront the constructed `PUSH_PARAMETER` with the one obtained by Besu.
  - [x] the bytecode ends abruptly with a PUSH instruction such that the bytecode itself is too short to construct the `PUSH_PARAMETER` --- tests the padding stuff
  - [x] the masking of bytes in the OPCODE column when they are to be interpreted as parameters of a PUSH
  - [x] the masking of JUMPDEST in particular under these cases
  - [x] the IS_VALID_JUMPDEST behaves correctly wrt PUSH masking
- Check against reference tests.

letypequividelespoubelles:


lorenzogentile404 (commented):
This PR contains tests already available in `JumpDestinationVettingTest`.

PR #1762, opened 2025-01-31
test: add SHAKIRA tests

### LINKED ISSUE #707: Add test cases for SHAKIRA_DATA

letypequividelespoubelles:


arithmetization/src/test/java/net/consensys/linea/zktracer/module/shakira/ShakiraInputsExtensiveTests.java
```diff
@@ -0,0 +1,150 @@
...
+              .push(
+                  rightPadTo(
+                      Bytes.concatenate(
+                          Bytes.of(PUSH2.byteValue()),
+                          Bytes.ofUnsignedShort(size),
+                          Bytes.of(PUSH4.byteValue()),
+                          Bytes.ofUnsignedInt(offset),
+                          Bytes.of(RETURN.byteValue())),
+                      WORD_SIZE))
+              .op(MSTORE)
```
OlivierBBB, line 137: As discussed on slack, wrong order of arguments. This will trigger an `MXPX` as the offset is
```rust
0x 61 xx 63 xx yy zz tt f3 00 .. 00
```

lorenzogentile404 (changes requested):
I believe the way you defined `SEED` makes the test non-deterministic, as the execution order is not always the same.  An approach I used somewhere else is initialise SEED inside the `MethodSource` with an hardcoded seed, so it is independent from the execution order. For `instructionSpecificBytecode` you may initialise `SEED` at every invokation using a seed that depends on the arguments of the call.

lorenzogentile404 (approved):

PR #1777, opened 2025-02-05
feat: enable popping bundle of transactions

### LINKED ISSUE #1727: Popping transaction bundles
The `zkTracer` must have the ability to pop multiple transactions at once, that is, transactions that belong to a given **bundle**. 

One solution is to keep our current `StackedSet`/`StackedList` mechanisms, to start tracing at `traceStartBundle` and  and to trigger the `pop`/`commit` (or `merge`?) operations at `traceEndBundle`. This solution isn't favoured by @fab-10. IIRC, and following explanations by @ahamlat, the `zkTracer` would be the only user of these new methods.

Alternatively we can achieve the same result in `traceStartTransaction` and trigger the `pop`/`commit` (or `merge`?) operations at `traceEndTransaction` when a bundle starts / when it ends.

**Notes.**
- the issue will become mostly obsolete with the _limitless prover_ (for everything except for precompiles)
- tracing a bundle that becomes untraceable after several transactions were already traced is time consuming and could be problematic

letypequividelespoubelles:


ahamlat (approved):
First review, LGTM.
Let's ship it, I will do a deeper performance analysis on the release.

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -304,19 +300,24 @@ public void traceContextExit(MessageFrame frame) {
...
+   * commitTransactions()}
+   */
+  public void popTransactions() {
+    hub.popTransactions();
+  }
+
+  public void commitTransactions() {
+    hub.commitTransactions();
+  }
+
```
OlivierBBB, line 320: Wouldn't it make sense to call it e.g. `popTransactionBundle` / `commitTransactionBundle` or `popBundle` / `commitBundle` ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/CountOnlyOperation.java
```diff
@@ -25,32 +25,28 @@
...
-   * can't be pop
-   */
-  public void enter() {
-    countCommitedToTheConflation += countInTransaction;
-    countInTransaction = 0;
+  @Getter private int countInTransactionBundle = 0;
+
+  public void commitTransactions() {
+    countCommitedToTheConflation += countInTransactionBundle;
+    countInTransactionBundle = 0;
```
OlivierBBB, line 32: Here we use both `Transactions` and `TransactionBundle`. My personal preference would be `TransactionBundle` everywhere.

letypequividelespoubelles: noted

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedList.java
```diff
@@ -29,90 +27,49 @@
...
     Preconditions.checkState(conflationFinished, "Conflation not finished");
-    return operationsCommitedToTheConflation;
-  }
-
-  public boolean isEmpty() {
-    return size() == 0;
-  }
-
-  public boolean contains(Object o) {
-    return operationsInTransaction.contains(o) || operationsCommitedToTheConflation.contains(o);
```
OlivierBBB, line 110: I take it `contains` was never used ? Don't we use this (or some variant of it) when adding stuff to the stacked set ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -90,21 +76,17 @@ public boolean isEmpty() {
   }
 
   public boolean contains(Object o) {
-    return operationsInTransaction().contains(o) || operationsCommitedToTheConflation().contains(o);
+    return operationsInTransactionBundle().contains(o)
+        || operationsCommitedToTheConflation().contains(o);
```
OlivierBBB, line 80: We do have a `contains` for `StackedSet`'s but not `StackedList`'s (which explains my previous question.)

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  public boolean isEmpty() {
+    return size() == 0;
+  }
+
+  public boolean contains(Object o) {
+    return operationsInTransactionBundle().contains(o)
+        || operationsCommitedToTheConflation().contains(o);
+  }
```
OlivierBBB: Now I'm confused. For `StackedSet`'s I understand the need for a `contains` method. `StackedList`'s on the other hand contain operations that are typically ordered and have to be carried out / traced no matter what. In essence: we don't  omit duplicates. Why do we need a contains for `StackedList`'s then ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -497,17 +500,17 @@ public void traceStartTransaction(final WorldView world, final Transaction tx) {
...
+      state.processingPhase(TX_SKIP);
       new TxSkipSection(this, world, transactionProcessingMetadata, transients);
     } else {
       if (transactionProcessingMetadata.requiresPrewarming()) {
-        state.setProcessingPhase(TX_WARM);
+        state.processingPhase(TX_WARM);
         new TxPreWarmingMacroSection(world, this);
       }
-      state.setProcessingPhase(TX_INIT);
+      state.processingPhase(TX_INIT);
```
OlivierBBB, line 516: Was this made `fluent` or something? I find the `setBla(X)` syntax more appealing than its `fluent` alternative.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/state/State.java
```diff
@@ -64,7 +65,10 @@ public void incrementMxpStamp() {
     mxpStamp++;
   }
 
-  @Getter @Setter HubProcessingPhase processingPhase;
+  @Getter
+  @Setter
+  @Accessors(fluent = true)
```
OlivierBBB, line 70: Hm 👀

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/state/HubTraceSections.java
```diff
@@ -13,27 +13,29 @@
...
 /**
- * Stores all the trace sections associated to the same transaction, stored in chronological order
- * of creation.
+ * Stores all the trace sections associated to one transaction, stored in chronological order of
+ * creation.
  */
+@Getter
 @Accessors(fluent = true)
-public class TxTrace {
+public class HubTraceSections {
```
OlivierBBB: We could alternatively just call it `TraceSections` without reference to the HUB (which is the module that contains this class anyway.)

OlivierBBB (approved):
LGTM, only cosmetic feedback
- we should replace`Transactions` with `TransactionBundle`)
- some questions around the `contains` for `StackedList`.

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/StackedList.java
letypequividelespoubelles: Indeed we don't need at all. IIRC, it was needed for one test only at the time I implemented the StackedList stuff. But apparently this is not needed anymore, so i deleted it.

fab-10 (approved):
Bundle tests on sequencer side with this branch are going well

PR #1781, opened 2025-02-06
chore: update `go-corset` to latest commit

DavePearce:
This updates `go-corset` to `75a88c5` as this contains some performance optimisations of interest.  Specifically, simplification of inverse columns.

### DIFF
--- a/.github/actions/setup-go-corset/action.yml
+++ b/.github/actions/setup-go-corset/action.yml
@@ -12,4 +12,4 @@ runs:
 
     - name: Install Go Corset
       shell: bash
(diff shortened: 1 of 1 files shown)


OlivierBBB (approved):
LGTM. I imagine that in general using a given commit over is preferable over using `@latest` ?

DavePearce:
> LGTM. I imagine that in general using a given commit over is preferable over using `@latest` ?

Yes, for the release process.  Otherwise, its hard to tell what version of `go-corset` a given release was tested against.

PR #1783, opened 2025-02-07
Extensive `OOGX` tests involving memory expansion + gas estimation tool `BytecodeRunner.runOnlyForGasCost` + `MSIZE` (#1783)

### LINKED ISSUE #1685: Nontrivial values tests for `stack-consistency` testing
The `OOGX` tests and normal "banal" tests are missing stuff. It's the same type of issue we already had with `PUSH`'s but for `SLOAD` (and potentially for `SSTORE`). The issue is that in the tests storage is empty when tested. So `SLOAD` (which triggers `OOGX`) dredges up 0, which is indistinguishable from not writing anything. But when storage is nonempty the associated constraint `storage-instruction---setting-storage-slot-values---doing-for-SLOAD` blows up in case of an `OOGX`.

We may also have the same issue with `MLOAD / MSTORE / MSTORE8 / CALLDATALOAD` if the `OOGX` tests have, respectively:
- empty memory
- empty call data

We need ensure nontrivial results for the following opcodes, both a "banal" (i.e. no `OOGX`) and `OOGX` versions, for the following:
```rust
SLOAD
SSTORE

// special crafted values
SIGNEXTEND
BYTE
SHR
SAR

// 0
ISZERO

// 1 and 1
AND
EQ

// some address in the state with nontrivial balance and code
BALANCE
EXTCODEHASH

// a value that is close to NUMBER
BLOCKHASH

// have nonempty RAM or nonempty CALLDATA
MLOAD
CALLDATALOAD

// should be called after having expanded memory once
MSIZE

// generic 7 * i + 5 will do:
ADD
MUL
SUB
DIV
SDIV
MOD
SMOD
ADDMOD
MULMOD
EXP
GT
SGT
EQ
SHL
PUSHX
SWAPX
DUPX

// generic 128 - 3 * i or so
LT
SLT
```

lorenzogentile404:
Closes #1685 
- `BytecodeRunner.runOnlyForGasCost`  Method to calculate cost dynamically - based on LondonGasCalculator (Besu)
- OOGX for opcodes with no mem expansion : list from `opCodeToOpCodeDataMap.values()` without opcodes with mem expansion cost
- OOGX for opcodes with mem expansion cost : MSTORE, MSTORE8, MLOAD, CALLDATACOPY, CODECOPY, EXTCODECOPY, RETURN, RETURNDATACOPY, REVERT, CREATE, LOGX
- MSIZE test: with non zero value and post memory expansion

Local benchmark done on 366 tests with
- calculate gas cost manually 1min40s
- calculate gas cost dynamically 1min30s
- calculate gas cost by running `BytecodeRunner.run` twice  2min50s 

Test coverage improvement 
+7 lines covered
+4 conditions covered
Results
**+0.3% condition coverage on Unit tests**
**+0.1% overall coverage score (with blockchain ref tests)**


arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -150,24 +138,21 @@ void outOfGasExceptionCallTest(
...
     program
         .push(0) // return at capacity
         .push(0) // return at offset
         .push(0) // call data size
         .push(0) // call data offset
         .push(value) // value
         .push("ca11ee") // address
-        .push(1000) // gas
+        .push(stipend) // gas
         .op(OpCode.CALL);
```
lorenzogentile404, line 142: Is `stipend` set to 0 and never changed? You probably wanted to change it based on `value`.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -176,26 +161,60 @@ void outOfGasExceptionCallTest(
...
-          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    /*
+                long gasCostExt =
+        GAS_CONST_G_TRANSACTION
+                + // base gas cost
+                (isWarm ? GAS_CONST_G_VERY_LOW + GAS_CONST_G_COLD_ACCOUNT_ACCESS : 0) // PUSH + BALANCE
+                + 7 * GAS_CONST_G_VERY_LOW // 7 PUSH
+                + callGasCost(value != 0, targetAddressExists, isWarm); // CALL
+    */
+
```
lorenzogentile404: Do we still need this comment?

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -176,26 +161,60 @@ void outOfGasExceptionCallTest(
...
 
+  /*
+      private long callGasCost(boolean transfersValue, boolean targetAddressExists, boolean isWarm) {
+      Preconditions.checkArgument(
+              !(isWarm && !targetAddressExists), "isWarm implies targetAddressExists");
+      return (transfersValue ? GAS_CONST_G_CALL_VALUE : 0)
+              + (targetAddressExists ? 0 : (transfersValue ? GAS_CONST_G_NEW_ACCOUNT : 0))
+              + (isWarm ? GAS_CONST_G_WARM_ACCESS : GAS_CONST_G_COLD_ACCOUNT_ACCESS);
+    }
+  */
```
lorenzogentile404: This was written with the limitation of not taking into consideration memory expansion. We probably can completely remove it.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -255,4 +259,76 @@ void outOfGasExceptionSLoad(int cornerCase) {
...
+            .op(OpCode.JUMPDEST)
+            .push(OpCode.JUMPDEST.byteValue())
+            .compile();
+
+    BytecodeRunner bytecodeRunner = BytecodeRunner.of(bytecode);
+    long gasCost;
+    if (cornerCase == -1) {
+      // JUMP needs a valid bytecode to jump to
+      // Calculate the gas cost to trigger OOGX on JUMP and not on the last but one opcode
+      gasCost = GAS_CONST_G_TRANSACTION + GAS_CONST_G_VERY_LOW + GAS_CONST_G_MID;
```
lorenzogentile404, line 238: I suggest to write here "JUMP needs JUMPDEST to jump to"

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -255,4 +259,76 @@ void outOfGasExceptionSLoad(int cornerCase) {
...
+            .compile();
+
+    BytecodeRunner bytecodeRunner = BytecodeRunner.of(bytecode);
+
+    long gasCost;
+    if (cornerCase == -1) {
+      // JUMP needs a valid bytecode to jump to
+      // Calculate the gas cost to trigger OOGX on JUMPI and not on the last but one opcode
+      gasCost = GAS_CONST_G_TRANSACTION + 2 * GAS_CONST_G_VERY_LOW + GAS_CONST_G_HIGH;
+    } else {
```
lorenzogentile404, line 269: Same as comment above.

testing/src/main/java/net/consensys/linea/testing/BytecodeRunner.java
```diff
@@ -64,36 +66,46 @@ public static BytecodeRunner of(Bytes byteCode) {
...
 
   // Ad-hoc gasLimit and accounts
   public void run(Long gasLimit, List<ToyAccount> additionalAccounts) {
-    this.run(Wei.fromEth(1), gasLimit, additionalAccounts);
+    this.run(Wei.fromEth(1), gasLimit, additionalAccounts, Bytes.EMPTY);
   }
 
   // Ad-hoc senderBalance, gasLimit and accounts
   public void run(Wei senderBalance, Long gasLimit, List<ToyAccount> additionalAccounts) {
+    this.run(senderBalance, gasLimit, additionalAccounts, Bytes.EMPTY);
```
lorenzogentile404, line 103: In a future PR we may want to find a more elegant way than overloading to emulate default arguments, maybe via annotations or so, but I believe this has low priority.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/MemoryTests.java
```diff
@@ -147,4 +147,20 @@ void revertAfterLog2() {
...
+  void checkMSizeAfterMemoryExpansion() {
+    BytecodeCompiler program = newProgram();
+    program
+        .push(0xFF)
+        .push(0)
+        .op(OpCode.MSTORE)
+        .op(OpCode.MSIZE)
+        .push(0xFF)
+        .push(1) // expand memory
+        .op(OpCode.MSTORE)
```
OlivierBBB: Obvious remark: both `MSTORE`'s expand memory, the first to size `0x20`, the second to size `0x40`.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+
+  @ParameterizedTest
+  @ValueSource(ints = {-1, 0, 1})
+  void outOfGasExceptionCallDataCopy(int cornerCase) {
+    BytecodeCompiler program = BytecodeCompiler.newProgram();
+
+    Bytes calldata =
+        Bytes.fromHexString("0x7FFFFFFFFFFFFF00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF");
+    program
+        .push(31) // value
```
OlivierBBB: This parameter is `size` rather than `value`.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+    bytecodeRunner.run(Wei.fromEth(1), gasCost + cornerCase, List.of(), calldata);
+    if (cornerCase == -1) {
+      assertEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    } else {
+      assertNotEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    }
```
OlivierBBB: This paragraph could be extracted into a method I imagine ? It's duplicated quite a number of times.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+        // Create the contract
+        .push(41)
+        .push(0)
+        .push(0)
+        .op(OpCode.CREATE)
+        .push(32) // size
+        .push(0) // offset
+        .push(33) // destoffset
+        .op(OpCode.DUP4) //
+        .op(OpCode.EXTCODECOPY);
```
OlivierBBB, line 193: It would make sense to rename this test to
```java
void outOfGasExceptionWarmExtCodeCopy(int cornerCase) { ... }
```
(the address in question will be warm after the successful deployment) and to add a second test called e.g. `outOfGasExceptionColdExtCodeCopy` where you
- define a `ToyAccount` with address `c0deadd7e55`, provide it with byte code that is e.g. simply `"ff".repeat(foreignCodeSize)` where you set `final int foreignCodeSize = 70;`, say, and add it to the list of accounts in the test
- then run the following code in the root
```java
program
        .push(foreignCodeSize + 3) // size
        .push(11) // offset
        .push(33) // destoffset
        .push("c0deadd7e55") // address
        .op(OpCode.EXTCODECOPY);
```
This test targets a cold account and will read out of bounds of its code.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+  void outOfGasExceptionReturn(int cornerCase) {
+    BytecodeCompiler program = BytecodeCompiler.newProgram();
+
+    program
+        .push(
+            Bytes.fromHexString(
+                "0xF00000000000000000000000000000000000000000000000000000000000FFAA")) // value
+        .push(0) // offset
+        .op(OpCode.MSTORE)
+        .push(3) // value
```
OlivierBBB: `value` is really a `size`

OlivierBBB: There should be a second version of this test where the transaction is a `DEPLOYMENT_TRANSACTION`. The pricing of the `RETURN` is different in this case (you pay 200 gas for every byte you are attempting to deploy.) The `program` you provide would have to be passed as the `Transaction.payload` instead. We have several tests where we use deployment transactions.

The version above could be renamed to `outOfGasExceptionForReturnFromMessageCallTransaction` and the deployment version `outOfGasExceptionForReturnFromDeploymentTransaction`.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+        .push(Bytes.fromHexString("0xFFFFFFFF")) // gas
+        .op(OpCode.STATICCALL)
+        // 3. Clean the stack
+        .op(OpCode.POP)
+        .op(OpCode.POP)
+        // 4. Return data copy
+        .push(32) // size
+        .push(0) // offset
+        .push(65) // destoffset, trigger mem expansion
+        .op(OpCode.RETURNDATACOPY);
```
OlivierBBB, line 295: This is more of a side note than anything.

The above could be achieved more directly. Currently you
- pre-populate the state with an account at address `c0de` with some code
- `EXTCODECOPY` this code in full
- use this code as the init code of a `CREATE` which deploys a new SMC with code I'll call `returnDataProviderCode`
- `STATICCALL` into this account, thus producing some nonempty return data

You could shortcircuit the first steps and instead
- pre-populate the state with an account at address `c0de` with `returnDataProviderCode`
- `STATICCALL` into the `c0de` account, thus producing some nonempty return data

The bytecode would be
```java
program
        .push(0) // byte size of return data
        .push(0) // retOffset
        .push(0) // byte size calldata
        .push(0) // argsOffset
        .op("c0de") // address of 'return data provider' account
        .op(OpCode.GAS) // gas
        .op(OpCode.STATICCALL)
        // 3. Clean the stack
        .op(OpCode.POP)
        // 4. Return data copy
        .push(32) // size
        .push(0) // offset
        .push(65) // destoffset, trigger mem expansion
        .op(OpCode.RETURNDATACOPY);
```

OlivierBBB: I would also suggest to add a variant of the above where we do the following
```java
program
        .push(0) // byte size of return data
        .push(0) // retOffset
        .push(0) // byte size calldata
        .push(0) // argsOffset
        .op("c0de") // address of 'return data provider' account
        .op(OpCode.GAS) // gas
        .op(OpCode.STATICCALL)
        // 3. Clean the stack
        .op(OpCode.POP)
        // 4. Doubly exceptional return data copy
        .op(OpCode.RETURNDATASIZE)
        .push(1)
        .op(OpCode.ADD); // size = RDS + 1, which will trigger the `returnDataCopyException`
        .push(0) // offset
        .push(65) // destoffset, trigger mem expansion
        .op(OpCode.RETURNDATACOPY);
```
We could name this test to `doubleExceptionReturnDataCopyTestOogxAndRdcx`. We have so far not written tests where we deliberately trigger several exceptions at once.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+      assertNotEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(ints = {-6420, -6419, -6418, 100, 101, 102})
+  /*
+  Deployment code: "0x7F7EFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF60005260206000F3"
```
OlivierBBB, line 337: What's the reason for using a `PUSH31` rather than a `PUSH32` as the first operation of the deployed code ? Just asking to understand what is going on. It seems to me that the code you are deploying is
```go
0x007effff..ff
```
I.e.
```rust
STOP
PUSH31 0xff .. ff
```
So that the deployed code, when called, will `STOP` immediately.

OlivierBBB: I guess it's not being called so it doesn't matter. Still an odd choice IIUC.

OlivierBBB: My bad, I got mixed up. The deployed code is
```rust
PUSH31 0xff .. ff
```
no `STOP`.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+        // Create the contract
+        .push(41)
+        .push(0)
+        .push(0)
+        .op(OpCode.CREATE); // No constructor so code executed and runtime code set to return value
+
+    Bytes pgCompile = program.compile();
+    BytecodeRunner bytecodeRunner = BytecodeRunner.of(pgCompile);
+
+    long gasCost = bytecodeRunner.runOnlyForGasCost();
```
OlivierBBB, line 366: Are we sure this is the correct `gasCost` ? I'm asking because a naïve implementation of `runOnlyForGasCost` would compute the total gas consumed by the transaction, say. But this would include the `PUSH`'s and `MSTORE`'s and `RETURN`'s of the deployment inside the `CREATE`. Are we sure those costs aren't taken into account when running `runOnlyForGasCost` ?

What we need here is the cost just up to an including the `upfrontGasCost` of the `CREATE`.

OlivierBBB: Oh, I guess this is what the `6420` etc is about ...

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+      assertEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    } else if (cornerCase <= 100) {
+      assertNotEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+      assertEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection(2).commonValues.tracedException());
```
OlivierBBB, line 380: Is this `OUT_OF_GAS_EXCEPTION` happening inside of the deployment frame ?

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -255,4 +238,76 @@ void outOfGasExceptionSLoad(int cornerCase) {
...
+      assertNotEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(ints = {-1, 0, 1})
+  void outOfGasExceptionJumpi(int cornerCase) {
```
OlivierBBB, line 250: Here, too, we have the opportunity to test simultaneous exceptions, in this case:
```java
jumpException
outOfGasException
```
And it would be trivial to implement, just switch the second push to
```java
    final Bytes bytecode =
        BytecodeCompiler.newProgram()
            .push(1) // pc = 0, 1
            .push(6) // pc = 2, 3
            // ...
```

OlivierBBB: It would make sense to create a new class for these `MultiExceptionTests`.

OlivierBBB: Which would include the `RETURNDATACOPY` double exception test mentioned earlier.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
```diff
@@ -0,0 +1,604 @@
...
+    if (cornerCase == -1) {
+      assertEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    } else {
+      assertNotEquals(
+          OUT_OF_GAS_EXCEPTION,
+          bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
+    }
+  }
```
OlivierBBB, line 413: In terms of `MultiExceptionsTest` one could have a simple test where one triggers simultaneously
```java
staticException
outOfGasException
```
or
```java
staticException
stackUnderflowException
```
One needs two smart contracts with byte code
```java
BytecodeCompiler simpleLog0 = BytecodeCompiler.newProgram();
simpleLog0.push(0).push(0).op(LOG0);

BytecodeCompiler stackUnderflowLog0 = BytecodeCompiler.newProgram();
stackUnderflowLog0.push(0).op(LOG0);
```
Which are called upon through a `STATICCALL` with `gas = 3 + 3 + 375 + cornerCase` / `gas = 3 + 375 + cornerCase`.

OlivierBBB (commented):
Left several comments

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/MemoryTests.java
amkCha: You're right, it felt better to go from a non-zero value to a non-zero value, but it's strictly the same behavior. Let me simplify it :+1:

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
amkCha: Good idea, I created a Utils method :ok_hand:

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
amkCha: Yes exactly, can it be checked in a better way ? It's the case where the deployment frame has enough gas to run the code but not enough to cover the depositFee, so it triggers an OOGX

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
amkCha: I created the issue here to follow up on `MultiExceptionTests` https://github.com/Consensys/linea-tracer/issues/1838

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
amkCha: Issue created for `MultiExceptionTests` https://github.com/Consensys/linea-tracer/issues/1838

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
amkCha: For the 1st comment, I refactored the bytecode, thanks for the explanation. The shortcircuit makes much more sense. I was using evm/playground, but we have more tooling for testing. 
For 2nd comment here, I opened the issue `MultiExceptionTests` https://github.com/Consensys/linea-tracer/issues/1838

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
amkCha: Added `outOfGasExceptionColdExtCodeCopy ` test :ok_hand:

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
amkCha: Yes, the test is parametric on different values we adjust the `gasCost` with ( `{-6420, -6419, -6418, 100, 101, 102}̀` ) to create 3 scenarios : 

  1. OOGX for CREATE before deployment: remove 6400 (depositFee) + deployment code exec cost (18)
  2. OOGX for CREATE after deployment: enough gas for child creation, but not enough to complete deployment code or deposit (we remove between 6419 and 6418 for instance)
  3. No OOGX for CREATE after deployment: we have to add 1/64th of 6418 to gasCost to account for gasAvailableForChildCreate (63/64 rule)

It's acually a `MultiExceptionTest`. Might move it to the new test class in subsequent PR

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasMemExpExceptionTest.java
OlivierBBB: No, LGTM just making sure I understood properly 🙂

OlivierBBB (approved):
LGTM

PR #1784, opened 2025-02-07
test: mixture of non reverted, self reverted, and get reverted LOGX

### LINKED ISSUE #671: Test reverting LOGs and the impact on the MMU module and its STAMP
We must ensure that the MMU_STAMP from the HUB and the MMU module are in sync, see #397. We should thus have tests with
- [ ] reverted internal transactions that produce (reverted) LOG's
  - [ ] with reverts caused by the current context (CONTEXT_SELF_REVERT = 1)
  - [ ] and with reverts caused by some parent context reverting (CONTEXT_GETS_REVERTED = 1)
- [ ] both reverted and unreverted LOG's

letypequividelespoubelles:


lorenzogentile404 (commented):
Unless testing the concatenation of all these programs is what you wish to test, I would make a parametric shorter test that executes the programs separately (or maybe keep both versions?).

letypequividelespoubelles:
> Unless testing the concatenation of all these programs is what you wish to test, I would make a parametric shorter test that executes the programs separately (or maybe keep both versions?).

It is indeed the concatenation of all this LOGX operation that are of interest, as we want to check that the LOG / MMU stamp behave as expected wrt to (self/get)-reverted in middle of others non reverted LOGX

lorenzogentile404 (approved):

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/RevertingLogsTests.java
```diff
@@ -0,0 +1,205 @@
...
+                  "0x000007031c400000000007031c400000000007031c400000000007031c400000")) // topic 4
+          .push(
+              Bytes.fromHexString(
+                  "0x000007031c300000000007031c300000000007031c300000000007031c300000")) // topic 3
+          .push(
+              Bytes.fromHexString(
+                  "0x000007031c200000000007031c200000000007031c200000000007031c200000")) // topic 2
+          .push(
+              Bytes.fromHexString(
+                  "0x000007031c100000000007031c100000000007031c100000000007031c100000")) // topic 1
```
OlivierBBB: It would make sense to declare some static `Bytes` in the class
```java
private static final Bytes TOPIC_1 = Bytes.fromHexString("0x000007031c100000000007031c100000000007031c100000000007031c100000")
// etc ...
```
So that you can simply e.g. `.push(TOPIC_2).push(TOPIC_1)`.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/RevertingLogsTests.java
```diff
@@ -0,0 +1,205 @@
...
+          .push(18) // size
+          .push(0x1) // offset
+          .op(OpCode.LOG4)
```
OlivierBBB: Should be `LOG3`

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/RevertingLogsTests.java
```diff
@@ -0,0 +1,205 @@
...
+        .push(0) // retSize
+        .push(0) // retOffset
+        .op(OpCode.MSIZE) // arg size
+        .push(0) // argOffset
+        .push(1) // value
+        .push(address) // address
+        .push(100000) // gas
+        .op(OpCode.CALL)
+        .compile();
+  }
```
OlivierBBB, line 148: It could be interesting to have two versions:
- `call`
- `staticCall`

Since LOG's will raise the static violation. That would give you an alternative self reverting LOG scenario.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/RevertingLogsTests.java
```diff
@@ -0,0 +1,205 @@
...
+import org.hyperledger.besu.crypto.KeyPair;
+import org.hyperledger.besu.crypto.SECP256K1;
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.datatypes.TransactionType;
+import org.hyperledger.besu.datatypes.Wei;
+import org.hyperledger.besu.ethereum.core.Transaction;
+import org.junit.jupiter.api.Test;
+
+public class RevertingLogsTests {
```
OlivierBBB, line 40: Could you add some documentation on what it is we are trying to test ?

OlivierBBB (changes requested):
see comments

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/RevertingLogsTests.java
```diff
@@ -0,0 +1,205 @@
...
+
+  private static final Bytes CALLDATACOPY =
+      newProgram()
+          .op(OpCode.CALLDATASIZE) // size
+          .push(0) // offset
+          .push(0) // dest offset
+          .op(OpCode.CALLDATACOPY)
+          .compile();
```
OlivierBBB, line 48: The purpose of this byte code chunk is to populate memory. Maybe let the name reflect that, e.g. calling it
```rust
POPULATE_MEMORY
```
Or so. Also you could achieve the same goal more reliably with a `PUSH32` / `MSTORE` combo (reliable in the sense that you don't need to remember to provide the transaction with a payload.) We do BTW have `populateMemory` methods at various places.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mmu/RevertingLogsTests.java
letypequividelespoubelles: indeed a very good idea

OlivierBBB (approved):
LGTM

Issue #1772, opened 2025-02-05
The nature of the "base" context of tests matters and must be included in parametric tests


OlivierBBB:
The last bug that remained in the EVM test suite highlights the importance of writing tests two ways: where the "base" (e.g. root of call frame tree) context is either
- `CONTRACT_CREATION`
- `MESSAGE_CALL`

Specifically the issue was when exploring the `scenario/CALL_EOA_SUCCESS_WILL_REVERT` where the caller was undergoing deployment and the deployment reverted. The issue was that when "undoing" the CALL at REVERT time we were using a more updated snapshot of the caller account, specifically one where the deployment had already been reverted. This clashed with `account_consistency` constraints.

PR #1788, opened 2025-02-10
test: add selfdestructing COINBASE

### LINKED ISSUE #1662: `SELFDESTRUCT`'ing `COINBASE` address tests
We require tests where the either
- the `recipient`
- the `coinbase`
- the two are the same or not
- either one is deployed during the transaction or already exists

`SELFDESTRUCT`, with or without reverting transaction. The right way to do this is as a parametric multi transaction test where in a follow up transaction we query which ever account `SELFDESTRUCT`ed (via `BALANCE` / `EXTCODEHASH` / `EXTCODESIZE` / `EXTCODECOPY` / `CALL` / `SELFDESTRUCT` (as the heir.)

letypequividelespoubelles:


arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/selfdestructTests/SelfdestructCoinbaseTests.java
```diff
@@ -0,0 +1,157 @@
...
+    deployedAccounts.add(senderAccount);
+    if (coinBaseDeployed) {
+      deployedAccounts.add(selfDestructorCoinbaseAccount);
+    }
+
+    ToyExecutionEnvironmentV2.builder()
+        .accounts(deployedAccounts)
+        .transactions(List.of(selfdestruction, checkingCoinbase))
+        .zkTracerValidator(zkTracer -> {})
+        .coinbase(depAddress)
```
OlivierBBB, line 147: Why is the `coinbase` set to `depAddress` ?

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/selfdestructTests/SelfdestructCoinbaseTests.java
```diff
@@ -0,0 +1,157 @@
...
+import org.hyperledger.besu.ethereum.core.Transaction;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class SelfdestructCoinbaseTests {
+
+  @ParameterizedTest
+  @MethodSource("selfDestructCoinbaseInputs")
+  void selfdestructCoinbaseTests(
```
OlivierBBB, line 71: Could we add some documentation on this test ?

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/selfdestructTests/SelfdestructCoinbaseTests.java
```diff
@@ -0,0 +1,157 @@
...
+            .payload(selfDestructorCoinbaseAccount.getCode())
+            .build();
+
+    final Transaction checkingCoinbase =
+        ToyTransaction.builder()
+            .sender(senderAccount)
+            .keyPair(senderKeyPair)
+            .value(Wei.of(123))
+            .gasLimit(100000L)
+            .payload(BytecodeCompiler.newProgram().op(OpCode.COINBASE).op(OpCode.BALANCE).compile())
```
OlivierBBB: In terms of checks I would add
```rust
EXTCODESIZE
EXTCODEHASH
```
Also I would suggest adding an `ToyAccount coinbaseCheckerAccount` rather than doing a deployment. If only because it will be easier to do multiple checks on the `COINBASE` if we expand these tests some more.

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/selfdestructTests/SelfdestructCoinbaseTests.java
letypequividelespoubelles: For the case "root context is deployment" (https://github.com/Consensys/linea-tracer/issues/1772) and  "coinbase / recipient address collision".

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/selfdestructTests/SelfdestructCoinbaseTests.java
letypequividelespoubelles: sure

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/selfdestructTests/SelfdestructCoinbaseTests.java
letypequividelespoubelles: done

OlivierBBB (approved):
LGTM

PR #1797, opened 2025-02-11
1792 cleaning some todos

### LINKED ISSUE #1792: cleaning some TODOs

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -854,7 +844,8 @@ private void exitDeploymentFromDeploymentInfoPov(MessageFrame frame) {
      * address A; bytecode A executes exactly the same CREATE2 raising the Failure Condition F for
      * address B;
      */
-    if (failureConditionForCreates) {
+    if (this.currentTraceSection() instanceof CreateSection
+        && ((CreateSection) this.currentTraceSection()).scenarioFragment.isFailedCreate()) {
```
OlivierBBB, line 848: LGTM. The only question I have is why `isFailedCreate` encompasses both
```rust
CREATE_FAILURE_CONDITION_WILL_REVERT
CREATE_FAILURE_CONDITION_WONT_REVERT
```
The original condition for setting `failedCreate = true` and thus `hub.failureConditionForCreates = true` was simply that the scenario, after refinement, be
```rust
CREATE_FAILURE_CONDITION_WONT_REVERT
```

This bit matters when executing `Hub.exitDeploymentFromDeploymentInfoPov(...)`. At that time we will not have had time for a **Failure Condition F** `CREATE(2)` to transition to the `REVERT` variant of this scenario. The attempted `CREATE(2)` was unexceptional, and in order to switch to `CREATE_FAILURE_CONDITION_WILL_REVERT` we will need at least another instruction carried out by the creator to either raise an exceptional halting condition or execute on the `REVERT` opcode.

What the above discussion also shows is that the original hack had an issue: we don't always `traceOpCode()` e.g. when encountering a `SUX` / `SOX`. We were thus running the risk of not resetting the `Hub.failureConditionForCreates` when expected. Not sure if this could have caused a bug, i.e. not sure whether we could have entered `Hub.exitDeploymentFromDeploymentInfoPov(...)` with that bit still inadvertently on ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/scenario/CreateScenarioFragment.java
```diff
@@ -57,13 +57,19 @@ public static CreateScenario switchToRevertingScenario(final CreateScenario curr
...
   public boolean isAbortedCreate() {
     return scenario == CreateScenario.CREATE_ABORT;
   }
 
+  public boolean isFailedCreate() {
+    return scenario.isAnyOf(
+        CreateScenario.CREATE_FAILURE_CONDITION_WILL_REVERT,
+        CreateScenario.CREATE_FAILURE_CONDITION_WONT_REVERT);
+  }
+
```
OlivierBBB, line 70: See previous comment

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/storage/StorageFragment.java
```diff
@@ -44,25 +42,6 @@ public final class StorageFragment implements TraceFragment {
...
-    // TODO: maybe remove in the future ?
-    checkArgument(match);
-    checkArgument(containsActualStorageSlotIdentifier);
-
-    final boolean isFirstOccurrence =
-        current.get(storageSlotIdentifier).getFirstOccurrence() == this;
-    final boolean isFinalOccurrence =
-        current.get(storageSlotIdentifier).getFinalOccurrence() == this;
-
-    // tracing
```
OlivierBBB, line 65: I guess this all was bloat that has been made obsolete by new `go-corset` tricks.

OlivierBBB (approved):
LGTM, left a comment about the `isFailedCreate` method.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/storage/StorageFragment.java
letypequividelespoubelles: yes

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
letypequividelespoubelles: yes, I can go back to only WONT revert case, althought it should be the same ...

PR #1800, opened 2025-02-12
tests: rlpTxn tests

### LINKED ISSUE #1793: `RLP_TXN` testing
We need extensive parametric tests.

letypequividelespoubelles:


lorenzogentile404 (changes requested):
There is still a TODO. Is it something to include in this PR? In general, could you add a short description of the purpose of the test or point to Java issue describing it? I just made the SEED non-static to avoid non-determinism in the case more tests are added in the same class in the future.

letypequividelespoubelles:
@lorenzogentile404 added doc. And yes, the TODO will remain, as I don't know how to solve it ATM

lorenzogentile404 (approved):

PR #1817, opened 2025-02-17
fix: no overflow for BlakeRounds precompile limits count

### LINKED ISSUE #1814: Make sur `BlakeRoundCount` cannot overflow
Blake `count`'s are two fold:
- one that counts the number of calls made to that precompile
- one that tallies the number of rounds

The system limit on the first is 0 (i.e. we aren't allowed to make any calls to Blake currently.) If and when that changes we will need to be careful about the second limit. The issue being that the numbers `r` we tally via `+= r` are 4 byte integers which may cause overflows if the second `count` is an `int`.

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/BlakeRounds.java
```diff
@@ -15,18 +15,66 @@
...
     return "PRECOMPILE_BLAKE_ROUNDS";
   }
+
+  @Override
+  public void addPrecompileLimit(final int count) {
+    throw new UnsupportedOperationException("Not implemented");
+  }
+
+  public void addPrecompileLimit(final Bytes r) {
+    Preconditions.checkArgument(r.bitLength() <= 32, "r is at most 4 bytes long");
```
OlivierBBB: I wonder whether it would make more sense to constrain it to
```java
r.size() <= 4
```
as `r` ought to be extracted as the first 4 bytes of call data, and `r.bitlength()` ignores the first bytes if they are zero.

OlivierBBB: We may even want to assert `r.size() == 4`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/BlakeRounds.java
```diff
@@ -15,18 +15,66 @@
...
+    final BigInteger rBI = r.toUnsignedBigInteger();
+    // check if r is greater or equal to Integer.MAX_VALUE
+    if (rBI.compareTo(INTEGER_MAX_VALUE_BI) >= 0) {
+      transactionBundleContainsIllegalOperation(true);
+      return;
```
OlivierBBB, line 55: Currently `BLAKE2f` is completely illegal. We should therefore set
```java
transactionBundleContainsIllegalOperation(true);
```
**unconditionally**, in particular even if `r` is zero.

letypequividelespoubelles: I do not agree on this, we have another module `BlakeEffectiveCall` that counts the number of calls to `Blake2f` The limit in the sequencer for this module must be 0 (Idk if that's the case in prod already), and not set in `BlakeRounds`.

OlivierBBB: One way could be to define `INTEGER_MAX_VALUE_BI = BigInteger.ZERO`.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/precompileLimits/ModexpIllegalOperationTests.java
```diff
@@ -13,7 +13,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package net.consensys.linea.zktracer.containers;
+package net.consensys.linea.zktracer.module.precompileLimits;
```
OlivierBBB, line 16: Will this change be ok if we apply it on top of the old release `beta-v1.2.0-rc3` ? Which doesn't for instance contain the changes related to `transactionBundles` ?

letypequividelespoubelles: I guess if you apply this change on top of https://github.com/Consensys/linea-tracer/commit/2109e842037b96973316c605be82b0e32939302a it's gonna be ok (and as you need this fix, it's ok) ... but I can revert it, it just makes more sense to have this tests in this package, but it's not very important

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/BlakeRounds.java
OlivierBBB: Yeah, I agree, I forgot about the other count.

OlivierBBB: The `INTEGER_MAX_VALUE_BI = BigInteger.ZERO` thing would still work (with some mandatory renaming) and make sense. I would consider it more coherent than using `BigInteger.valueOf(MAX_VALUE)`. We also achieve that both counts (`total_number_of_calls` and `total_number_of_rounds`) agree when we've overstepped with our current limitations.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/precompiles/BlakeRounds.java
```diff
@@ -15,18 +15,67 @@
...
 import net.consensys.linea.zktracer.container.stacked.CountOnlyOperation;
+import org.apache.tuweni.bytes.Bytes;
 
 @Getter
 @Accessors(fluent = true)
 public final class BlakeRounds implements CountingOnlyModule {
   private final CountOnlyOperation counts = new CountOnlyOperation();
+  @Setter private boolean transactionBundleContainsIllegalOperation = false;
+
+  private static final BigInteger INTEGER_MAX_VALUE_BI = BigInteger.valueOf(MAX_VALUE);
```
OlivierBBB, line 37: Maybe rename to `MAX_NUMBER_OF_BLAKE_ROUNDS` and set it to `BigInteger.ZERO`.

letypequividelespoubelles: I still don't understand why you want this. In teh sequencer we have a limit of 0 for this module, so why we should change the lineCount to have all the time MAX_INTEGER instead of the real value ? It's the sequencer job to confront the real value to the limit we want (which is 0).

OlivierBBB: I guess the question is why would we want to compare to `MAX_VALUE` in the first place ?
```java
   if (rBI.compareTo(INTEGER_MAX_VALUE_BI) >= 0) {
      transactionBundleContainsIllegalOperation(true);
      return;
```
The value `Integer.MAX_VALUE` is meaningless in the present context and doesn't correlate with any value e.g. found in the limit files. We do however in the `limits.toml` under `tests`, which is the file that contains the up to date limits, have the value `0`.

OlivierBBB (approved):
Let's merge it

PR #1831, opened 2025-02-19
Improve MMIO line counting performance

ahamlat:
CPU profiling on ERC20 and ERC721 transactions during performance load testing showed that `isFastOperation` method in `MmioData` class, takes up to 25% of the block selection time.

**ERC20 transactions**

image

**ERC721 transactions**

image

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
@@ -141,7 +141,7 @@ public class Mmio implements Module {
     final boolean isRamToRamTwoTarget = (mmioData.instruction() == MMIO_INST_RAM_TO_RAM_TWO_TARGET);
     final boolean isRamVanishes = (mmioData.instruction() == MMIO_INST_RAM_VANISHES);
     int mmioLineCountingInstruction = lineCountOfMmioInstruction(mmioData.instruction());
-    for (short ct = 0; ct < mmioLineCountingInstruction ; ct++) {
+    for (short ct = 0; ct < mmioLineCountingInstruction; ct++) {
       trace
           .cnA(Bytes.minimalBytes(mmioData.cnA()))
           .cnB(Bytes.minimalBytes(mmioData.cnB()))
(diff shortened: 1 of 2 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
```diff
@@ -140,8 +140,8 @@ void trace(Trace trace, MmioData mmioData, final int stamp) {
     final boolean isRamToRamTwoSource = (mmioData.instruction() == MMIO_INST_RAM_TO_RAM_TWO_SOURCE);
     final boolean isRamToRamTwoTarget = (mmioData.instruction() == MMIO_INST_RAM_TO_RAM_TWO_TARGET);
     final boolean isRamVanishes = (mmioData.instruction() == MMIO_INST_RAM_VANISHES);
-
-    for (short ct = 0; ct < lineCountOfMmioInstruction(mmioData.instruction()); ct++) {
+    int mmioLineCountingInstruction = lineCountOfMmioInstruction(mmioData.instruction());
```
letypequividelespoubelles, line 143: This shouldn't impact the linecount performances as this is called at commit time (= tracing time, ie when we write the .lt file), way after calling lineCount(). With this change, at tracing time, you'll compute a second time the line counting, instead of reading the value only.

letypequividelespoubelles (changes requested):
see comment

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
ahamlat: I'm not sure to understand your comment. The change I did is to avoid computing  lineCountOfMmioInstruction for each iteration. In this change, it is done only once. There is nothing else that should change here.

letypequividelespoubelles: > I'm not sure to understand your comment.

Hum, that's normal, I misread your change, so my comment is totally stupid, my bad ^^

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mmio/Mmio.java
OlivierBBB: Could you explain @ahamlat  how computing
```java
int mmioLineCountingInstruction = lineCountOfMmioInstruction(mmioData.instruction());
```
ahead of the for loop impacts performance ?

OlivierBBB: Is it otherwise being recomputed with every iteration of the loop ?

OlivierBBB (commented):
LGTM

PR #1843, opened 2025-02-20
Optimize OpCodes class

ahamlat:
From CPU profiling on different tests, I noticed that `CallFrame.frame` call takes around 7% of the selection time, and most of the time is related to getting the opcode data related to the opcode label or number

image

image

This PR stores all the opcode data in an ArrayList, and the index in the array list represents to number of the opCode, ex. Opcode `DIV` will be stored in index `4` (`0x04`).


arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodes.java
```diff
@@ -32,8 +31,8 @@
 public class OpCodes {
   private static final JsonConverter YAML_CONVERTER = JsonConverter.builder().enableYaml().build();
 
-  public static Map<Integer, OpCodeData> valueToOpCodeDataMap;
-  public static Map<OpCode, OpCodeData> opCodeToOpCodeDataMap;
+  private static final short OPCODES_LIST_SIZE = 256;
+  static List<OpCodeData> opCodes = new ArrayList<>(OPCODES_LIST_SIZE);
```
letypequividelespoubelles: Could we give it a better name ? maybe `opcodesList` ? 
Also, would it make sense to merge OpCode and Opcodes calss ?

letypequividelespoubelles (approved):
LGTM. Were you able to do a perf test on this PR to see the improvement ?

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodes.java
ahamlat: I like the having both, Opcodes as a utility class, and OpCode as the ENUM of all the EVM opcodes.

PR #1850, opened 2025-02-25
clean: some cleaning

### LINKED ISSUE #1834: rm dep number / status infinity

### LINKED ISSUE #1849: cleaning

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/Oob.java
```diff
@@ -37,17 +37,16 @@
 /** Implementation of a {@link Module} for out of bounds. */
 @RequiredArgsConstructor
 @Accessors(fluent = true)
-public class Oob implements OperationListModule<OobOperation> {
-  // TODO @Lorenzo why it's not a StateLess module ?
+public class Oob implements OperationSetModule<OobOperation> {
```
letypequividelespoubelles, line 40: @lorenzogentile404 are you ok with this ?

OlivierBBB: @letypequividelespoubelles pointed out for instance that some OOB tests may be using the output of the latest `OobOperation` ... in which case switching to `StackedSet`'s may cause issues.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
```diff
@@ -45,12 +49,14 @@ public String moduleKey() {
   }
 
   @Override
-  public void tracePreOpcode(MessageFrame frame) {
-    operations.add(
-        new AddOperation(
-            OpCode.of(frame.getCurrentOperation().getOpcode()),
-            Bytes32.leftPad(frame.getStackItem(0)),
-            Bytes32.leftPad(frame.getStackItem(1))));
+  public void tracePreOpcode(MessageFrame frame, OpCode opcode, short exception) {
```
amkCha: Question : why are we doing this check in many modules ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
```diff
@@ -76,21 +77,19 @@ public void traceStartBlock(
   }
 
   @Override
-  public void tracePreOpcode(MessageFrame frame) {
-    final OpCode opCode = OpCode.of(frame.getCurrentOperation().getOpcode());
-    checkArgument(opCode == OpCode.BLOCKHASH, "Expected BLOCKHASH opcode");
+  public void tracePreOpcode(MessageFrame frame, OpCode opcode, short exception) {
+    if (opcode == BLOCKHASH && Exceptions.none(exception)) {
```
amkCha: Why do we test except as none here and only oogx in other modules ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/SstoreSection.java
```diff
@@ -111,7 +111,8 @@ public SstoreSection(Hub hub, WorldView worldView) {
     commonValues.refundDelta(
         Hub.GAS_PROJECTOR
             .of(hub.currentFrame().frame(), hub.opCode())
-            .refund()); // TODO should use Besu's value
+            .refund()); // Note: we can't use Besu's refund value, as our is only for non-reverting
```
amkCha, line 114: Shouldn't it be the same value though ?

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobRdcTest.java
```diff
@@ -376,9 +376,9 @@ void testReturnDataCopyUsingIdentityPrecompileMaxPosRds() {
     // Chunk with index 1 is the one corresponding to IDENTITY precompile
     // precompileCost = (5 + ceil) * 3 where ceil = 1
     // cds is trivially 0 in initReturnDataCopyProgramUsingIdentityPrecompile
-    assertEquals(
-        bytecodeRunner.getHub().oob().operations().get(1).getPrecompileCost(),
-        BigInteger.valueOf(18));
+    // assertEquals(
```
amkCha: Checking, you guys want to keep that commented ?

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionAtContextReEntry.java
```diff
@@ -0,0 +1,174 @@
...
+    // message call to a simple smart contract
+    program
+        .push(0) // r@c
+        .push(0) // r@c
+        .push(0) // r@c
+        .push(0) // r@c
+        .push(storageKey)
+        .op(SLOAD) // deployment address
+        .push(0xff) // gas
+        .op(STATICCALL);
```
amkCha, line 82: To be sure it matches the name, is this call going to be unsuccessful ?

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionAtContextReEntry.java
```diff
@@ -0,0 +1,174 @@
...
+import static net.consensys.linea.zktracer.opcode.OpCode.*;
+
+import net.consensys.linea.testing.BytecodeCompiler;
+import net.consensys.linea.testing.BytecodeRunner;
+import net.consensys.linea.zktracer.module.hub.Hub;
+import org.apache.tuweni.bytes.Bytes;
+import org.junit.jupiter.api.Test;
+
+/**
+ * We explore what happens if the first instruction after resuming the execution of a frame is
```
amkCha, line 27: Great commenting on that test. Makes it easy to review :+1:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/SstoreSection.java
OlivierBBB: We can't: refunds in Besu are tallied as the instructions (specifically: `SSTORE`) roll in. If a frame reverts, the refund counter reverts to its value pre-frame. Our arithmetization does things differently: we only add the refund to the counter if the frame does not revert. This cannot be decided in the moment. You have to know whether the frame will revert or not, which becomes apparent only at transaction end.

It's a design choice. It could be done "live" and undone (like we undo e.g. `BALANCE`/`WARMTH` changes) but at the time this was designed this seemed like the simplest option.

TBH this could be something we change in the future. It wouldn't be a hard change and would homogenise our approach.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionAtContextReEntry.java
OlivierBBB: The transactions in all 4 tests in this class are supposed to finish on an exception (a `stackUnderflowException` induced by an `ADD` in the root. Thus all transactions are supposed to fail.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobRdcTest.java
letypequividelespoubelles: it's @lorenzogentile404 's test. We could keep those checks, Idk how important they are

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionAtContextReEntry.java
letypequividelespoubelles: THX @OlivierBBB  ;)

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionAtContextReEntry.java
OlivierBBB: To answer your question directly (after misunderstanding it initially) : yes, this CALL will fail: the first instruction it executes is an `ADD` on an empty stack. This raises a stackUnderflowException, causing the CALL to fail, consume all of it's gas, and the parent frame to resume execution.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
letypequividelespoubelles: the exception check ? 

I just refactor the logic we had before while killing the `Signals` class. 

Now you say so, I find quite inefficient to check for exceptions each time, we should just trace the operation, even if we don't need it. Having one more operation per exception vs checking the exceptions for each opcode ... I think it worths it. 

Would you be ok @OlivierBBB  ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/oob/Oob.java
lorenzogentile404: If I remember correctly, I do not assume that operations are ordered in the test related to OOB. I check for example that there exist an operation in which `insufficientGasForPrecompile` has been detected within the operation for the handling of a certain precompile call. We need to ensure this is the case for every test involving OOB however.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/Add.java
letypequividelespoubelles: This PR's current version:
- check for exception at each opcode to know if trace the operation.
Pros: we trace only what we need
Cons: a LOT of exceptions check (and 99,99% are useless, as no exceptions)

@amkCha 's proposal:
- don't check for exception, and trace it anyway. 
Pros: no exceptions check while tracing / sequencing 
Cons: when an exception occurs, we trace one ADD/BIN/SHF/etc operation, that can adds up to 16 rows in the trace ...

OlivierBBB: Exceptions are rare, so most of the time we will indeed be checking and finding no issue. I would rather we remain strict about only tracing the instructions that are required. This respects the HUB logic. Also the exceptions check involves comparing a short to 0, it shouldn't have any meaningful impact on performance.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/oob/OobRdcTest.java
letypequividelespoubelles: @lorenzogentile404 will uopdate the check

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
letypequividelespoubelles: I just wrote the same logic as before. But you're right, we don't want trace anything, whatever the exception is, am I right @OlivierBBB  ? 
It's for BIN SHF ADD MUL EXT MOD WCP BLOCKHASH ... we never need the exceptional operation ?

lorenzogentile404 (approved):
I adjusted where needed the `Oob` business, so LGTM.

amkCha (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
OlivierBBB: We never need the exceptional operation @letypequividelespoubelles. I guess checking none would make sense. I think this is sometimes written with `oogx`, sometimes with `none`, way because these operations can only fail due to:
- stack underflow (but then they don't make it to `traceOpcode`)
- out of gas (in which case they do make it to here, given that the stack was fine that is)

But we should be able to switch to `none(...)` for those instructions

letypequividelespoubelles:
> I adjusted where needed the `Oob` business, so LGTM.

CI are failing now ;)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
letypequividelespoubelles: sounds good. I'll factorise the exception check

PR #1852, opened 2025-02-26
fix(statetests): lower go mem limit

amkCha:
Testing to make general state tests pass with lower go mem limit (set like blockchain ref tests)


letypequividelespoubelles (approved):
LGTM

PR #1862, opened 2025-02-28
feat: support embedded metadata in trace files

### LINKED ISSUE #1861: feat: embed metadata into trace files
The goal here is to utilise the improved trace file format to embed metadata relevant to the trace for the purposes of debugging, and also for configuration checks.  In particular, the goal is to embed block range, tracer version, chain id and the constaints commit used to generate the trace files.

DavePearce:
This adds support for embedding metadata into the generated trace file.  The metadata is sourced from two places: the `zkevm.bin` file, and also some information gathered by the tracer (e.g. block range, etc).

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
@@ -25,6 +25,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import net.consensys.linea.plugins.config.LineaL1L2BridgeSharedConfiguration;
@@ -73,6 +75,11 @@ public class ZkTracer implements ConflationAwareOperationTracer {
   /** Accumulate all the exceptions that happened at tracing time. */
   @Getter private final List<Exception> tracingExceptions = new FiniteList<>(50);
 
+  // Fields for metadata
+  private long startBlock = Long.MAX_VALUE;
+  private long endBlock = 0;
+  private BigInteger chainId;
+
   public ZkTracer() {
     this(
         LineaL1L2BridgeSharedConfiguration.TEST_DEFAULT,
@@ -85,6 +92,7 @@ public class ZkTracer implements ConflationAwareOperationTracer {
 
   public ZkTracer(
       final LineaL1L2BridgeSharedConfiguration bridgeConfiguration, BigInteger chainId) {
+    this.chainId = chainId;
     this.hub = new Hub(bridgeConfiguration.contract(), bridgeConfiguration.topic(), chainId);
     for (Module m : this.hub.getModulesToCount()) {
       if (!spillings.containsKey(m.moduleKey())) {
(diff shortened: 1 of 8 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -317,4 +337,9 @@ public Map<String, Integer> getModulesLineCount() {
     modulesLineCount.put("BLOCK_TRANSACTIONS", hub.cumulatedTxCount());
     return modulesLineCount;
   }
+
+  public static byte[] getMetadataBytes(Map<String, String> metadata) throws IOException {
+    ObjectWriter mapper = new ObjectMapper().writer();
```
FlorianHuc: This can be made static to avoid reinstantiating it each time.

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -73,6 +75,11 @@ public class ZkTracer implements ConflationAwareOperationTracer {
   /** Accumulate all the exceptions that happened at tracing time. */
   @Getter private final List<Exception> tracingExceptions = new FiniteList<>(50);
 
+  // Fields for metadata
+  private long startBlock = Long.MAX_VALUE;
```
FlorianHuc: The logic to update start and end block is not straightforward to get. Can't we set them when the zktracer is updated? You have all the info when you instantiate the zktracer object in GenerateConflatedTracesV2

FlorianHuc (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -73,6 +75,11 @@ public class ZkTracer implements ConflationAwareOperationTracer {
...
+  private long endBlock = 0;
+  private BigInteger chainId;
```
letypequividelespoubelles: shouldn't it be final ?

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
DavePearce: Agreed, done.

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
DavePearce: Agreed, done.

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
DavePearce: Ok, I have resolved this now ... but it took a few changes to thread through the relevant information.

DavePearce:
@letypequividelespoubelles @FlorianHuc I have now updated this PR to additionally embed line counts.  However, there is currently an issue (to some extent) because it doesn't embed line counts for non-tracing modules, such as keccack.  I need to figure out where that information comes from.

FlorianHuc (approved):

letypequividelespoubelles:
> @letypequividelespoubelles @FlorianHuc I have now updated this PR to additionally embed line counts. However, there is currently an issue (to some extent) because it doesn't embed line counts for non-tracing modules, such as keccack. I need to figure out where that information comes from.

Added it

PR #1866, opened 2025-03-03
fix: l2l1Log count

### LINKED ISSUE #728: check the formula to count the number of `KECCAK` is up to date

### LINKED ISSUE #1863: [Counting] Reviewing the counting for the L2 to L1 logs
We recently found an issue in the counting logic for the L2L1 logs. It was invisible because the L2L1 logs are quite sparse. In the prover, we have a hard limit on how many L2L1 logs can happen at once in one conflated batch. The issue is about reviewing how this limit is taken into account by the conflator and the sequencer to ensure it is correctly enforced.

# Definition

L2L1Logs counts the numbers logs happening in a conflated execution batch, such that:
* The topic[0] is the keccak-hash of the event signature `MessageSent(address,address,uint256,uint256,uint256,bytes,bytes32)` which is equal to `0xe856c2b8bd4eb0027ce32eeaf595c21b0b6b4644b326e5b7bd80a1cf8db72e6c`
* The emitting address is the L2-bridge contract address. (`0x971e727e956690b9957be6d51Ec16E73AcAC83A7` on sepolia)

# Expected behaviour

* The sequencer should count the events satisfying the two above conditions and ensure it does not seal a block overflowing this limit.
* The conflator should ensure it does not create conflated execution batches that are overflowing the limit.

letypequividelespoubelles:
Issues that are fixed:
- we weren't calling those modules at traceEndTx, so there were never triggered, so their lineCount remained at 0 all the time.
- for l2l1Logs, we were counting the sum of the log data size, instead of the nb of calls
-  all the computation were completely outdated, reviewed with @AlexandreBelling

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
+++ b/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
@@ -28,9 +28,12 @@ public record LineaL1L2BridgeSharedConfiguration(Address contract, Bytes topic)
   private static Bytes LINEA_L2L1TOPIC =
       Bytes.fromHexString("0xe856c2b8bd4eb0027ce32eeaf595c21b0b6b4644b326e5b7bd80a1cf8db72e6c");
 
-  public static final LineaL1L2BridgeSharedConfiguration SEPOLIA_DEFAULT =
+  private static final Address SEPOLIA_L2L1LOGS_SMC =
+      Address.fromHexString("0x971e727e956690b9957be6d51Ec16E73AcAC83A7");
+
+  public static final LineaL1L2BridgeSharedConfiguration TEST_DEFAULT =
       LineaL1L2BridgeSharedConfiguration.builder()
-          .contract(Address.fromHexString("0x971e727e956690b9957be6d51Ec16E73AcAC83A7"))
+          .contract(Address.fromHexString("0x7e577e577e577e577e577e577e577e577e577e57"))
           .topic(LINEA_L2L1TOPIC)
           .build();
 
--- a/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
+++ b/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
@@ -25,6 +25,7 @@ import org.hyperledger.besu.datatypes.Address;
 public record LineaL1L2BridgeSharedConfiguration(Address contract, Bytes topic)
     implements LineaOptionsConfiguration {
 
+  // = Hash(MessageSent(address,address,uint256,uint256,uint256,bytes,bytes32))
   private static Bytes LINEA_L2L1TOPIC =
       Bytes.fromHexString("0xe856c2b8bd4eb0027ce32eeaf595c21b0b6b4644b326e5b7bd80a1cf8db72e6c");
 
--- a/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
+++ b/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
@@ -24,6 +24,16 @@ import org.hyperledger.besu.datatypes.Address;
 @Builder(toBuilder = true)
 public record LineaL1L2BridgeSharedConfiguration(Address contract, Bytes topic)
     implements LineaOptionsConfiguration {
+
+  private static Bytes LINEA_L2L1TOPIC =
+      Bytes.fromHexString("0xe856c2b8bd4eb0027ce32eeaf595c21b0b6b4644b326e5b7bd80a1cf8db72e6c");
+
+  public static final LineaL1L2BridgeSharedConfiguration SEPOLIA_DEFAULT =
+      LineaL1L2BridgeSharedConfiguration.builder()
+          .contract(Address.fromHexString("0x971e727e956690b9957be6d51Ec16E73AcAC83A7"))
+          .topic(LINEA_L2L1TOPIC)
+          .build();
+
   public static final LineaL1L2BridgeSharedConfiguration EMPTY =
       LineaL1L2BridgeSharedConfiguration.builder()
           .contract(Address.ZERO)
--- a/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
+++ b/arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
@@ -36,14 +36,4 @@ public record LineaL1L2BridgeSharedConfiguration(Address contract, Bytes topic)
           .contract(Address.fromHexString("0x7e577e577e577e577e577e577e577e577e577e57"))
           .topic(LINEA_L2L1TOPIC)
           .build();
-
-  public static final LineaL1L2BridgeSharedConfiguration EMPTY =
-      LineaL1L2BridgeSharedConfiguration.builder()
-          .contract(Address.ZERO)
-          .topic(Bytes.EMPTY)
-          .build();
-
-  public boolean isEmpty() {
-    return this.contract.equals(Address.ZERO) || this.topic.isEmpty();
-  }
 }
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
@@ -75,12 +75,12 @@ public class ZkTracer implements ConflationAwareOperationTracer {
 
   public ZkTracer() {
     this(
-        LineaL1L2BridgeSharedConfiguration.SEPOLIA_DEFAULT,
+        LineaL1L2BridgeSharedConfiguration.TEST_DEFAULT,
         Bytes.fromHexString("c0ffee").toUnsignedBigInteger());
   }
 
   public ZkTracer(BigInteger nonnegativeChainId) {
-    this(LineaL1L2BridgeSharedConfiguration.SEPOLIA_DEFAULT, nonnegativeChainId);
+    this(LineaL1L2BridgeSharedConfiguration.TEST_DEFAULT, nonnegativeChainId);
   }
 
   public ZkTracer(
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
@@ -75,12 +75,12 @@ public class ZkTracer implements ConflationAwareOperationTracer {
 
   public ZkTracer() {
     this(
-        LineaL1L2BridgeSharedConfiguration.EMPTY,
+        LineaL1L2BridgeSharedConfiguration.SEPOLIA_DEFAULT,
         Bytes.fromHexString("c0ffee").toUnsignedBigInteger());
   }
 
   public ZkTracer(BigInteger nonnegativeChainId) {
-    this(LineaL1L2BridgeSharedConfiguration.EMPTY, nonnegativeChainId);
+    this(LineaL1L2BridgeSharedConfiguration.SEPOLIA_DEFAULT, nonnegativeChainId);
   }
 
   public ZkTracer(
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
@@ -85,7 +85,6 @@ public class ZkTracer implements ConflationAwareOperationTracer {
 
   public ZkTracer(
       final LineaL1L2BridgeSharedConfiguration bridgeConfiguration, BigInteger chainId) {
-    ;
     this.hub = new Hub(bridgeConfiguration.contract(), bridgeConfiguration.topic(), chainId);
     for (Module m : this.hub.getModulesToCount()) {
       if (!spillings.containsKey(m.moduleKey())) {
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/CountingOnlyModule.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/CountingOnlyModule.java
@@ -40,7 +40,7 @@ public interface CountingOnlyModule extends Module {
     return counts().lineCount();
   }
 
-  default void addPrecompileLimit(final int count) {
+  default void addLimit(final int count) {
     Preconditions.checkArgument(count >= 0, "Must be positive");
     counts().add(count);
   }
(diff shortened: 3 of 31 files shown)


arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
```diff
@@ -24,6 +24,16 @@
 @Builder(toBuilder = true)
 public record LineaL1L2BridgeSharedConfiguration(Address contract, Bytes topic)
     implements LineaOptionsConfiguration {
+
+  private static Bytes LINEA_L2L1TOPIC =
```
FlorianHuc, line 29: is this topic chain agnostic?

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -75,17 +75,16 @@ public class ZkTracer implements ConflationAwareOperationTracer {
 
   public ZkTracer() {
     this(
-        LineaL1L2BridgeSharedConfiguration.EMPTY,
+        LineaL1L2BridgeSharedConfiguration.SEPOLIA_DEFAULT,
```
FlorianHuc: don't we have a check that this should be non empty in the code, with a failure if it's empty? Setting it as sepolia by default would make the check useless. What would hapen if we use sepolia default on mainnet?

arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
letypequividelespoubelles: I think so, but not sure.

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
letypequividelespoubelles: Normally, the ZkTracer wo the LineaL1L2BridgeSharedConfiguration is used only by test, but yes, should be good to add some checks.

For testing purpose, I could create a new address, instead of the sepolia prod one, would you prefer ?

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/KeccakBlocksTests.java
```diff
@@ -0,0 +1,103 @@
...
+    // check lineCount of Keccak
+    final int txRlpSize = tx.encoded().size();
+    final int txKeccak =
+        numberOfKeccakBloc(txRlpSize)
+            + 1
+            + 1; // numberOfKeccakBloc(txRlpSize) + 1 for the tx + 2 for EcRecover
+    final int rlpAddrKeccak = 1 + 1; // 1 for CREATE, 1 for CRETAE2
+    assertEquals(txKeccak + rlpAddrKeccak, keccak.lineCount());
+
+    // check lineCount of l2Block
```
amkCha, line 100: Maybe a comment on the 20 +38 ? Or is it obvious ?

amkCha (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpaddr/RlpAddrOperation.java
```diff
@@ -42,19 +43,25 @@ public final class RlpAddrOperation extends ModuleOperation {
   private final Bytes32 keccak;
 
   // CREATE operation
-  public RlpAddrOperation(Bytes32 rawDepAddress, OpCode opCode, BigInteger nonce, Address address) {
+  public RlpAddrOperation(
+      Keccak keccak, Bytes32 rawDepAddress, OpCode opCode, BigInteger nonce, Address address) {
     this(rawDepAddress, opCode, nonce, address, Bytes32.ZERO, Bytes32.ZERO);
+    // We hash RLP (Address + nonce) which is at most 1 + (1+20) + (1+8) = 31 bytes
+    keccak.addLimit(31);
```
amkCha: Not sure there is a best way to do this, but if there are other hardcoded limits elsewhere, we might want to put them together and have them in 1 file

arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
OlivierBBB: Wasn't this topic the hash of the associated method signature? If so then the answer should be yes.

**Edit.** According to [this issue](https://github.com/Consensys/linea-tracer/issues/1863) it's the hash of `MessageSent(address,address,uint256,uint256,uint256,bytes,bytes32)`, which is borne out using e.g. [this tool](https://emn178.github.io/online-tools/keccak_256.html).

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpaddr/RlpAddrOperation.java
letypequividelespoubelles: it's the only place where we see this constant ...

arithmetization/src/main/java/net/consensys/linea/zktracer/container/module/CountingOnlyModule.java
```diff
@@ -40,7 +40,7 @@ default int lineCount() {
     return counts().lineCount();
   }
 
-  default void addPrecompileLimit(final int count) {
+  default void addLimit(final int count) {
     Preconditions.checkArgument(count >= 0, "Must be positive");
```
OlivierBBB: Rien à voir but the correct way to express `≥0` is **nonnegative**.
**Positive** means `>0`.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/KeccakBlocksTests.java
letypequividelespoubelles: adding it, it comes from the libeCount formula in l2Block, for one tx / one block

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -255,7 +255,9 @@ private List<Module> precompileLimitModules() {
         modexpEffectiveCall,
         ripemdBlocks,
         blakeEffectiveCall,
-        blakeRounds);
+        blakeRounds,
+        l2Block,
+        l2L1Logs);
```
OlivierBBB, line 260: Was that the omission you mentioned @letypequividelespoubelles ?

letypequividelespoubelles: exactly. Sorry, should have pointed to the code.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -379,10 +381,14 @@ public List<Module> getModulesToCount() {
...
-    l2Block = new L2Block(l2l1ContractAddress, LogTopic.of(l2l1Topic));
-    l2L1Logs = new L2L1Logs(l2Block);
+    if (l2l1ContractAddress.equals(TEST_DEFAULT.contract())) {
+      log.info("WARN: Using default testing L2L1 contract address");
+    }
+    l2L1Logs = new L2L1Logs();
+    l2Block = new L2Block(l2L1Logs, l2l1ContractAddress, LogTopic.of(l2l1Topic));
     keccak = new Keccak(ecRecoverEffectiveCall, l2Block);
     shakiraData = new ShakiraData(wcp, sha256Blocks, keccak, ripemdBlocks);
+    rlpAddr = new RlpAddr(this, trm, keccak);
```
OlivierBBB, line 391: Were we also missing the `rlpAddr` ?

arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
OlivierBBB: It may be worthwhile to add this as a comment on the definition of `LINEA_L2L1TOPIC`.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/ecdata/EcData.java
```diff
@@ -89,18 +89,16 @@ public void callEcData(
...
+      case PRC_ECMUL -> ecMulEffectiveCall.addLimit(ecDataOperation.internalChecksPassed() ? 1 : 0);
+      case PRC_ECRECOVER -> ecRecoverEffectiveCall.addLimit(
           ecDataOperation.internalChecksPassed() ? 1 : 0);
       case PRC_ECPAIRING -> {
         // ecPairingG2MembershipCalls case
         // NOTE: the other precompile limits are managed below
         // NOTE: see EC_DATA specs Figure 3.5 for a graphical representation of this case analysis
         if (!ecDataOperation.internalChecksPassed()) {
-          ecPairingG2MembershipCalls.addPrecompileLimit(0);
+          ecPairingG2MembershipCalls.addLimit(0);
```
OlivierBBB: `addLimit` seems like a misnomer. For count only operations are updating a tally, so `updateTally` seems more appropriate.

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/L2L1LogsTests.java
```diff
@@ -0,0 +1,180 @@
...
+    final KeyPair senderKeyPair = new SECP256K1().generateKeyPair();
+    final Address senderAddress =
+        Address.extract(Hash.hash(senderKeyPair.getPublicKey().getEncodedBytes()));
+    final ToyAccount senderAccount =
+        ToyAccount.builder().balance(Wei.fromEth(123)).nonce(12).address(senderAddress).build();
+
+    final ToyAccount LOG_ACCOUNT =
+        ToyAccount.builder()
+            .balance(Wei.fromEth(1))
+            .address(Address.wrap(Bytes.repeat((byte) 1, Address.SIZE)))
```
OlivierBBB, line 109: Is this supposed to work ? Maybe related to test configuration stuff, but the address `0x0101..0101` matches neither the real one (`0x971e727e956690b9957be6d51Ec16E73AcAC83A7`) nor the other test address (`0x7e57..7e57`).

**Edit.** I see that you also have transactions targeting the `l2l1LogSMC` account.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
letypequividelespoubelles: no, but as we were missing counting the Keccak coming from RLP_ADDR for deployment / create /create2, so I had to change the constructor of RLP_ADDR

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/KeccakBlocksTests.java
```diff
@@ -0,0 +1,109 @@
...
+
+    final Keccak keccak = toyWorld.getHub().keccak();
+    final L2Block l2Block = toyWorld.getHub().l2Block();
+
+    // check lineCount of Keccak
+    final int txRlpSize = tx.encoded().size();
+    final int txKeccak =
+        numberOfKeccakBloc(txRlpSize)
+            + 1
+            + 1; // numberOfKeccakBloc(txRlpSize) + 1 for the tx + 2 for EcRecover
```
OlivierBBB: The description does not match the action (+1, +1 vs +1, +2)

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/L2L1LogsTests.java
letypequividelespoubelles: it's supposed to NOT trigger the L2L1Block, yes, I have a chackArgument that the linecount is 0 at the end of the tx

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/KeccakBlocksTests.java
letypequividelespoubelles: indeed, it's +1+1, thx

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
```diff
@@ -127,17 +92,34 @@ public List<Trace.ColumnHeader> columnHeaders() {
 
   @Override
   public void traceEndTx(TransactionProcessingMetadata tx) {
+    numberOfTransactions.add(1);
+
     for (Log log : tx.getLogs()) {
-      if (log.getLogger().equals(l2l1Address) && log.getTopics().contains(l2l1Topic)) {
-        this.l2l1LogSizes.peek().add(log.getData().size());
+      if (isL2L1Log(log)) {
+        l2l1LogSizes.add(log.getData().size());
```
OlivierBBB, line 99: Do we now whether the messages can be arbitrary size when the relevant L2/L1 logging method is called ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
```diff
@@ -55,69 +60,29 @@ public String moduleKey() {
...
-            L1_MSG_INDICES_BYTES * txCount
-            + ABI_LEN_BYTES
-            + ABI_OFFSET_BYTES
-            + // the L1 messages
-            ABI_LEN_BYTES
-            + ABI_OFFSET_BYTES; // abi overheads for the blockdata struct.
-
-    return l1Size;
+        // Calculates the data size related to the block
+        + nbBlock * (TIMESTAMP_BYTESIZE + Hash.SIZE + NB_TX_IN_BLOCK_BYTESIZE);
```
OlivierBBB, line 85: I can imagine what the `TIMESTAMP_BYTESIZE` contributes to, what do the other two terms represent ? In particular is the `+ 2` i.e. `+ NB_TX_IN_BLOCK_BYTESIZE` linked to the transactions being assigned a number or so ?

OlivierBBB (approved):
LGTM. I left a couple more comments / questions.

arithmetization/src/main/java/net/consensys/linea/plugins/config/LineaL1L2BridgeSharedConfiguration.java
OlivierBBB: @letypequividelespoubelles

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
OlivierBBB: Wow 👀 several important pieces that got lost along the way ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
letypequividelespoubelles: @AlexandreBelling told it's fixed data size. You're afraid of overflow ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
letypequividelespoubelles: NB_TX_IN_BLOCK_BYTESIZE just tells the nb of transactions we have in each blocks, and this info takes 2 bytes in the encryption

PR #1874, opened 2025-03-05
fix: corsetBlockProcessor override issue

amkCha:
Besu `MainnetBlockProcessor` class was updated lately for tx parallelization and the override on `CorsetBlockProcessor` fails and need to be updated


letypequividelespoubelles (approved):
LGTM

PR #1879, opened 2025-03-06
Fix/general state test bis

amkCha:
- Find optimization between GOMEMLIMIT and parallelization :for running 105759 tests General State tests  
           - before : **5h30min**
           - after : **2h10min**

- Ignore General State tests failing with `Invalid width`error message


letypequividelespoubelles (approved):
LGTM

PR #1882, opened 2025-03-06
chore: create a proper module to count transactions

### LINKED ISSUE #1880: add txCount module

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/BlockTransactions.java
```diff
@@ -0,0 +1,59 @@
...
+  public void popTransactionBundle() {}
+
+  @Override
+  public void commitTransactionBundle() {}
+
+  @Override
+  public int lineCount() {
+    final int hubNumberOfTx = hub.state.txCount();
+    final int txnDataNumberOfTx = hub.txnData().operations().size();
+    assert (hubNumberOfTx == txnDataNumberOfTx);
```
amkCha, line 50: Are the two different numbers calculated differently ? What happens if the `assert` fails ?

letypequividelespoubelles: > Are the two different numbers calculated differently ? 

yes, txnData is triggered at at endTransaction, whereas hub is triggered at startTransaction. For the first one, we count the nulmber of TxnDataOperation, for the second, a little bit more complicated, but equivalent of the number of transaction, as it's the number of state, and we create a new state at each tx.  As the line counting happens after the end of the transaction, it must be the same. 

> What happens if the assert fails ?

We are ****** :) More seriously, it's a real issue if this happens, but even without the assert, we would have issues, as the constraints (especially lookups between txnData and HUB) would fail.

amkCha (approved):

PR #1884, opened 2025-03-06
fix(allcov): class source

amkCha:
- Classes not specified in jacoco test reports for nightly, weekly and fastReplay
- Fix name for report upload on fastReplay


DavePearce (approved):
LGTM

PR #1886, opened 2025-03-07
feat: update modules to report their own spillage info

### LINKED ISSUE #1885: feat: incorporate spillage information from `Trace.java`
The goal here is to automatically incorporate spillage information from the generated `Trace.java` file, rather than reading the `spillings.toml` file.  As such, that file can be deleted.

DavePearce:
Modules now report their own spillage information which, in most cases, is obtained from the generated Trace.java file.  As such, there is no longer any need for the spillings.toml file and associated machinery for reading it.


letypequividelespoubelles (approved):
LGTM, thx

PR #1890, opened 2025-03-11
feat: utilise parameterised zkevmbin

### LINKED ISSUE #1889: feat: utilise parameterised `zkevm.bin`
The goal here is to use a single `zkevm.bin` file which is parameterised, thereby preventing the need for `zkevm_for_reference_tests.bin` and `zkevm_for_old_linea.bin`.  This will exploit new functionality available in `go-corset`.

DavePearce:


arithmetization/src/test/java/net/consensys/linea/replaytests/ReplayTests.java
```diff
@@ -173,45 +163,16 @@ void modexpTriggeringNonAlignedFirstLimbSingleSourceMmuModexp() {
...
 
   @Test
   void incorrectCreationCapture() {
-    replay(LINEA_MAINNET, "4323985.mainnet.json.gz");
-  }
-
-  @Disabled
-  @ParameterizedTest
-  @MethodSource("replayBlockTestSource")
-  void replayBlockTest(int blockNumber) {
```
DavePearce: @lorenzogentile404 I removed this because it doesn't seem to be used, and I think it won't work anymore either since it relies on the `capture.pl` script which also doesn't work anymore (due to missing shadow node).  But, if you prefer to keep it ... that's fine.  I can put it back in easily enough.

arithmetization/src/main/java/net/consensys/linea/plugins/continoustracing/ContinuousTracingBlockAddedListener.java
```diff
@@ -59,7 +60,9 @@ public void onBlockAdded(final AddedBlockContext addedBlockContext) {
           try {
             final CorsetValidator.Result traceResult =
                 continuousTracer.verifyTraceOfBlock(
-                    blockHeader.getNumber(), blockHash, new ZkTracer());
+                    blockHeader.getNumber(),
+                    blockHash,
+                    new ZkTracer(Bytes.fromHexString("c0ffee").toUnsignedBigInteger()));
```
DavePearce: I don't understand how this chain id makes sense, but this is exactly what was happening before.  I think perhaps we should just use `LINEA_MAINNET` here?

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -61,29 +62,20 @@ public class ZkTracer implements ConflationAwareOperationTracer {
...
-        LineaL1L2BridgeSharedConfiguration.TEST_DEFAULT,
-        Bytes.fromHexString("c0ffee").toUnsignedBigInteger());
+  public ZkTracer(BigInteger nonnegativeChainId) {
+    this(LINEA_MAINNET(nonnegativeChainId));
   }
 
-  public ZkTracer(BigInteger nonnegativeChainId) {
-    this(LineaL1L2BridgeSharedConfiguration.TEST_DEFAULT, nonnegativeChainId);
+  public ZkTracer(ChainConfig chain) {
+    this(LineaL1L2BridgeSharedConfiguration.TEST_DEFAULT, chain);
```
DavePearce: @letypequividelespoubelles Should we fold `LineaL1L2BridgeSharedConfiguration` into `ChainConfig`?  They seem related.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
```diff
@@ -70,7 +69,7 @@ public void traceStartConflation(final long blockCount) {
             + 1
             + 6 // for NUMBER
             + 1 // for DIFFICULTY
-            + (bigIntegerToBytes(GAS_LIMIT_MAXIMUM).size() * 4) // for GASLIMIT
+            + (bigIntegerToBytes(chain.gasLimitMaximum).size() * 4) // for GASLIMIT
```
DavePearce, line 72: @lorenzogentile404 The constant `GAS_LIMIT_MAXIMUM` still exists in class `Trace`.  However, it should never be used as it only ever holds the default value (i.e. the value for Linea mainnet).  In the future, I might add a feature so we can decide which constants are visible in the `Trace.java` file (e.g. `:public` or something).  We would then make this constant not visible, so as to protect against someone accidentally using it.

arithmetization/src/main/java/net/consensys/linea/plugins/continoustracing/ContinuousTracingBlockAddedListener.java
OlivierBBB: IIRC this value `c0ffee` was chosen to be recognizable and to fix a NPE in certain (?) unit tests.

arithmetization/src/test/java/net/consensys/linea/replaytests/Issue1031Tests.java
```diff
@@ -28,6 +28,6 @@ public class Issue1031Tests {
 
   @Test
   void issue_1090_range_10_20() {
-    replay(LINEA_MAINNET, "10-20.mainnet.json.gz");
+    replay(OLD_LINEA_MAINNET, "10-20.mainnet.json.gz");
```
OlivierBBB: Was the chain id (I take it) of mainnet changed ?

arithmetization/src/test/java/net/consensys/linea/replaytests/ReplayTests.java
OlivierBBB: We have an issue to update the capture scripts https://github.com/Consensys/linea-tracer/issues/1892

plugins/src/main/java/net/consensys/linea/plugins/continoustracing/ContinuousTracingPlugin.java
```diff
@@ -93,7 +94,8 @@ public void start() {
 
     besuEvents.addBlockAddedListener(
         new ContinuousTracingBlockAddedListener(
-            new ContinuousTracer(traceService, new CorsetValidator(), tracingConfiguration),
+            new ContinuousTracer(
+                traceService, new CorsetValidator(ChainConfig.LINEA_MAINNET), tracingConfiguration),
```
OlivierBBB: Names such as `SLACK_SHADOW_NODE_WEBHOOK_URL` in this class strongly suggest that `ContinuousTracer` was indeed related to the shadow node. And should likely be removed altogether.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
OlivierBBB: Does anyone remember what the `* 4` accomplishes ?

arithmetization/src/main/java/net/consensys/linea/corset/GoCorsetValidator.java
```diff
@@ -138,6 +147,24 @@ private List<String> buildCommandLine(Path traceFile, String zkEvmBin) {
...
+   * @param options
+   */
+  private void setChainConstants(List<String> options) {
+    if (chain != LINEA_MAINNET) {
+      options.add("-Sblockdata.GAS_LIMIT_MINIMUM=" + chain.gasLimitMinimum.toString());
+      options.add("-Sblockdata.GAS_LIMIT_MAXIMUM=" + chain.gasLimitMaximum.toString());
+      if (chain.blockGasLimitEnabled) {
+        options.add("-Sblockdata.GAS_LIMIT_ENABLE=1");
+      } else {
+        options.add("-Sblockdata.GAS_LIMIT_ENABLE=0");
```
OlivierBBB, line 163: What is the convention for `S` and `D` for the `-Dblockchain` and `-Sblockdata` options ?

arithmetization/src/main/java/net/consensys/linea/zktracer/ChainConfig.java
```diff
@@ -0,0 +1,110 @@
...
+          LINEA_CHAIN_ID,
+          false,
+          BigInteger.valueOf(LINEA_GAS_LIMIT_MINIMUM),
+          BigInteger.valueOf(LINEA_GAS_LIMIT_MAXIMUM));
+
+  /**
+   * Represents Linea sepolia prior to the block gas limit being enforced for the purposes of
+   * running existing replay tests.
+   */
+  public static final ChainConfig OLD_LINEA_SEPOLIA =
```
OlivierBBB: Ok, this answers my previous question about what the `OLD` `ChainConfig`'s represent.

arithmetization/src/main/java/net/consensys/linea/zktracer/ChainConfig.java
```diff
@@ -0,0 +1,110 @@
...
+
+  /** Determines minimum gas limit for this chain */
+  public final BigInteger gasLimitMinimum;
+
+  /** Determins maximum gas limit for this chain */
+  public final BigInteger gasLimitMaximum;
+
+  private ChainConfig(
+      int chainId,
+      boolean gasLimitEnabled,
```
OlivierBBB, line 95: Tiniest nitpick: maybe we could rename this to `fixedGasLimitEnabled` ?

OlivierBBB (approved):
LGTM, left some comments

arithmetization/src/test/java/net/consensys/linea/replaytests/Issue1031Tests.java
DavePearce: @OlivierBBB no, the chain id is unchanged for linea mainnet.  What has changed, however, is the fixed gas limit constraint.  All of our existing replay tests were generated on a version of Linea which didn't have this restriction.  @letypequividelespoubelles called this "old linea" and I've just borrowed that.  Using the chain configuration `OLD_LINEA_MAINNET` here means that the `zkevm.bin` will be run with the constant `GAS_LIMIT_ENABLE=0` and, as a result, the constraint `block-gas-limit-value` is disabled (see [here](https://github.com/Consensys/linea-constraints/blob/master/blockdata/processing/gaslimit/linea.lisp))

plugins/src/main/java/net/consensys/linea/plugins/continoustracing/ContinuousTracingPlugin.java
DavePearce: Yeah, so we should discuss this ... and if it is indeed only for the shadow node then we should remove it.  More code just makes life harder.

arithmetization/src/main/java/net/consensys/linea/corset/GoCorsetValidator.java
DavePearce: Yeah, so there are three things here at least:

* `-Dblockchain=Ethereum`.   This is a gradle property definition which will be historical whence this PR is merged.
* `-Dcommit="${GIT_COMMIT}"` (as seen in the `Makefile` for `linea-constraints`).  This is a `go-corset` command which "defines" a piece of metadata in the `zkevm.bin` file.
* `-Sblockdata.GAS_LIMIT_ENABLE=0` (as seen above).  This is a `go-corset` command which "sets" the value of a given "externalised constant" (e.g. `(defconst (X :extern) 1)` is an example of one).  All externalised constants have a default value, but this can be override using the `-S` option as illustrated here.

I mean, yeah, the naming is somewhat arbitrary.  Potentially `-C` might be better than `-S` as this command is about "externalised [C]onstants".  Certainly, it can be changed.

arithmetization/src/main/java/net/consensys/linea/zktracer/ChainConfig.java
DavePearce: Yes, we could.  I will action that now.

DavePearce: Potentially we also want to update `linea-constraints` so the constant is named e.g. `FIXED_GAS_LIMIT_ENABLE` for example.

arithmetization/src/main/java/net/consensys/linea/zktracer/ChainConfig.java
```diff
@@ -0,0 +1,110 @@
...
+import static net.consensys.linea.zktracer.Trace.LINEA_SEPOLIA_CHAIN_ID;
+
+import java.math.BigInteger;
+
+/**
+ * Provides s single point of configuration for running the tracer on different chains which, aside
+ * from Linea mainnet, are all currently for the purposes of testing.
+ */
+public class ChainConfig {
+  /** Represents Linea mainnet as it stands today which enforces the block gas limit. */
```
letypequividelespoubelles: In all following comment, I would add `2 billion` to have `enforces the 2 billions block gas limit`

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
letypequividelespoubelles: yes, it would be a really good idea. Need to ask the relevant address and Topic0 for mainnet / sepolia to @AlexandreBelling

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/Blockdata.java
letypequividelespoubelles: yes, it will disappear soon.
It's to count (an upper limit of) the number of rows we have to add to the line count of WCP, as we call WCP only at the end of the block. The *4 is because there are 4 rows of calling WCP, where the biggest arg is chain.gasLimitMaximum

letypequividelespoubelles (approved):
LGTM. Good idea to merge the L2L1 info in chainConfig

arithmetization/src/test/java/net/consensys/linea/replaytests/ReplayTests.java
lorenzogentile404: Indeed, the capture script has been updated and the PR should be merged soon.

lorenzogentile404 (changes requested):
LGTM, except I would not remove the parametric replay test, as the `capture.pl` script is under update.

lorenzogentile404 (approved):

PR #1891, opened 2025-03-11
Feat/state manager tests part 1

amkCha:
Implement workaround for @bogdanbear state manager tests
- TransactionAccount
- TransactionStorage
- BlockwideDeployment
as information/columns are not available in the Hub anymore

### DIFF
--- a/testing/src/main/java/net/consensys/linea/testing/MultiBlockExecutionEnvironment.java
+++ b/testing/src/main/java/net/consensys/linea/testing/MultiBlockExecutionEnvironment.java
@@ -17,6 +17,7 @@ package net.consensys.linea.testing;
 
 import static net.consensys.linea.zktracer.Trace.LINEA_BLOCK_GAS_LIMIT;
 
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


testing/src/main/java/net/consensys/linea/testing/MultiBlockExecutionEnvironment.java
```diff
@@ -37,6 +39,9 @@ public class MultiBlockExecutionEnvironment {
 
   private final List<BlockSnapshot> blocks;
 
+  public static final BigInteger CHAIN_ID = BigInteger.valueOf(1337);
+  private static final ZkTracer tracer = new ZkTracer(CHAIN_ID);
+
```
amkCha, line 44: Instantiate it in the class to be able to access the tracer in the `getHub` method

testing/src/main/solidity/TestingBase.sol
```diff
@@ -258,7 +271,8 @@ abstract contract TestingBase {
      */
     function deployWithCreate2(
         bytes32 _salt,
-        bytes memory _bytecode
+        bytes memory _bytecode,
+        bool _revertFlag
```
amkCha, line 275: @bogdanbear Just checking that we should keep the changes in the  `TestingBase` contract

bogdanbear: In my opinion, having the possibility to revert might be useful—but if you think you do not need it, feel free to remove it. If you take that route, you will have to remove the revert flags from the tests.

bogdanbear (approved):

PR #1894, opened 2025-03-11
feat: count the right number of keccak and not an upper bound

### LINKED ISSUE #1893: feat: exact value for Keccak line count

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -54,10 +54,10 @@ public int lineCount() {
 
     return
     // From tx RLPs, used both for both the signature verification and the
-    // public input computation. As l2Block.sizesRlpEncodedTxs().lineCount() gives the size of the
-    // concatenation of all
-    // the RLP-encoded transactions, we add txCount to not miss keccak blocks.
-    (numberOfKeccakBloc(l2Block.sizesRlpEncodedTxs().lineCount()) + txCount)
+    // public input computation.
+    l2Block.sizesRlpEncodedTxs().getAll().stream()
```
lorenzogentile404: there should be a method called `.stream() `in the `sizesRlpEncodedTxs()` if it is a` StackedSet` or `StackedList` that does this directly.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
```diff
@@ -71,7 +72,7 @@ public void popTransactionBundle() {
   @Override
   public int lineCount() {
 
-    return sizesRlpEncodedTxs.lineCount()
+    return sizesRlpEncodedTxs.getAll().stream().reduce(0, Integer::sum)
```
lorenzogentile404: Same here, `getAll().stream()` should already be done by a named named `stream()` of `StackedList`.

lorenzogentile404 (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
OlivierBBB: I would imagine that most modules implementing the `stackedList` (and `stackedSet`) use essentially this implementation. It may make sense to have a `default` implementation of the `lineCount` method which is precisely this.

OlivierBBB (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
letypequividelespoubelles: I removed it, simplified and make it more perf.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
letypequividelespoubelles: > I would imagine that most modules implementing the stackedList (and stackedSet) use essentially this implementation. It may make sense to have a default implementation of the lineCount method which is precisely this.

Yes, this is the purpose of the `ModuleOperationStackedList`. All tracing module (RLP_*, MMU, etc) implements this.

OlivierBBB: My point was whether we should have a default implementation that does something à la 
```java
this.getAll().stream().reduce(0, Integer::sum);
```

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
```diff
@@ -49,23 +49,20 @@ public void updateTally(final int count) {
...
+    counts.lineCount()
         // From ecRecover precompiles,
         // This accounts for the keccak of the recovered public keys to derive the
         // addresses. This also accounts for the transactions signatures
         // verifications.
-        + (txCount + ecRecoverCount) * numberOfKeccakBloc(PUBKEY_BYTES)
-
-        // From deployed contracts, number of Keccak block for SHA3 and CREATE2, and from RLP_ADDR
-        + counts.lineCount();
+        + (txCount + ecRecoverCount) * numberOfKeccakBloc(PUBKEY_BYTES);
```
OlivierBBB, line 65: Did anything change here ? It's not clear from the git diff ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
```diff
@@ -107,7 +109,11 @@ public void traceEndTx(TransactionProcessingMetadata tx) {
     // overhead for each transaction (32 bytes for an offset, and 32 bytes for
     // to encode the length of each sub bytes array). This overhead is also
     // incurred by the top-level array, hence the +1.
-    sizesRlpEncodedTxs.add(tx.getBesuTransaction().encoded().size());
+    final int txDataSize = tx.getBesuTransaction().encoded().size();
+    sizesRlpEncodedTxs.add(txDataSize);
+    // Counts the number of Keccak from tx RLPs, used both for both the signature verification and
+    // the public input computation.
+    keccak.updateTally(txDataSize);
```
OlivierBBB, line 116: I take it the prover hashes both the raw transaction RLP and the re-RLP-ization of the subset required for signature verification ? How are both accounted for here ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
letypequividelespoubelles: We never use `this.getAll().stream().reduce(0, Integer::sum);` It was the only place where we used it, and I just removed it, so ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/Keccak.java
letypequividelespoubelles: just here, no. But now, all the rest is in `counts`, updated with `updateTally` when necessary

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
letypequividelespoubelles: hum, no in the formula, the prover hashes only the RLP-ization of the full tx. That's a good point that we don't count the keccak of the RLP-ization of the subset ...

OlivierBBB: If we want a quick and dirty solution we can just multiply `txDataSize` by 2. Obviously that defeats the purpose of this PR 😛

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L2Block.java
```diff
@@ -107,7 +109,15 @@ public void traceEndTx(TransactionProcessingMetadata tx) {
...
+    // TODO: this is for the RLP of the unsigned transaction, not the signed one. This is an upper
+    // bond. Waiting for Besu to expose the method that enables to compute the RLP of the un signed
+    // tx
+    keccak.updateTally(txDataSize);
```
OlivierBBB, line 121: Ok 🙂. Tiny nitpick: `upper bound` + I would rephrase it to e.g.
```java
    // TODO: this accounts for the message (hash) which the raw transaction signed.
    // Recall that said message is assembled (re-RLP-ized) from fields of the raw transaction.
    // This is an upper bound. Waiting for Besu to expose the method which computes said re-RLP-ized message (length.)
```

OlivierBBB: As I recall we already have access to said re-RLP-ized message in the RLP_TXN module, do we not ? It was a check that the message which we construct coincides with that of Besu.

letypequividelespoubelles: yes, but we compute it at commit time. I've asked @fab-10 to make the method public in Besu, so we'll be able to retrieve it once besu version is updated in the tracer repo.

PR #1895, opened 2025-03-11
Feat/validation reference test flow

amkCha:
- Add validation flow for reference tests when we open a PR
- Triggers a 4min30s flow broken down to : 2min of repo checkout + 2.5 min of generating tests and running 1 test
image

- Additional Fixes :
1. upload of report for Nightly
2. upload of report for Weekly

### DIFF
--- a/.github/workflows/gradle-ethereum-tests.yml
+++ b/.github/workflows/gradle-ethereum-tests.yml
@@ -29,9 +29,9 @@ jobs:
           JAVA_OPTS: -Dorg.gradle.daemon=false
 
       - name: Run General State Reference Tests
-        run: GOMEMLIMIT=40GiB ./gradlew referenceGeneralStateTests
+        run: GOMEMLIMIT=20GiB ./gradlew referenceGeneralStateTests
         env:
-          REFERENCE_TESTS_PARALLELISM: 3
+          REFERENCE_TESTS_PARALLELISM: 4
           JAVA_OPTS: -Dorg.gradle.daemon=false
           CORSET_FLAGS: disable
           GOCORSET_FLAGS: -vw -b1024 --ansi-escapes=false --report --air
--- a/.github/workflows/validate-reference-test-flow.yml
+++ b/.github/workflows/validate-reference-test-flow.yml
@@ -33,6 +33,7 @@ jobs:
         run: ./gradlew -Dblockchain=Ethereum referenceBlockchainTests -x spotlessCheck
         timeout-minutes: 15
         env:
+          GENERATE_REF_BLOCKCHAIN_TESTS: false
           JAVA_OPTS: -Dorg.gradle.daemon=false
           CORSET_FLAGS: disable
           GOCORSET_FLAGS: -b1024 -vw --ansi-escapes=false --report --air
--- a/.github/workflows/validate-reference-test-flow.yml
+++ b/.github/workflows/validate-reference-test-flow.yml
@@ -1,5 +1,5 @@
 name: "Validate Reference Test Flow"
-on: [push, pull_request]
+on: [pull_request]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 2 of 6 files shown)


.github/workflows/gradle-ethereum-tests.yml
```diff
@@ -29,9 +29,9 @@ jobs:
           JAVA_OPTS: -Dorg.gradle.daemon=false
 
       - name: Run General State Reference Tests
-        run: GOMEMLIMIT=40GiB ./gradlew -Dblockchain=Ethereum referenceGeneralStateTests
+        run: GOMEMLIMIT=64GiB ./gradlew -Dblockchain=Ethereum referenceGeneralStateTests
```
DavePearce: I would be cautious of increasing the `GOMEMLIMIT` so high.  Its probably going to have the opposite effect to what you want.  I don't think we ever need a limit above `20GiB` or so.  With 2 threads in parallel, we have a usage of 2 x 64Gb = 128Gb. That is the machines limit, which means it will certainly run out of resources on big traces which push those limits.

What you had before was also problematic: 40Gb x 3 = 120Gb ... still too high for a machine which (I believe) only have 128GB.  The blockchain reference tests uses 4 x 20Gb = 80Gb ... which is well under the machine limit.

.github/workflows/validate-reference-test-flow.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+
+      - name: Setup Test Environment
+        uses: ./.github/actions/setup-environment
+        with:
+          go-corset: true
+
+      - name: Run Reference Blockchain test sample
+        run: ./gradlew -Dblockchain=Ethereum referenceBlockchainTests -x spotlessCheck
```
DavePearce: Hmmmm, this doesn't look right to me ... it will run all the tests, no?  I think you need e.g. `--tests "BlockchainReferenceTest_55"` where `55` is just one of the tests I picked at random.  Not sure what a good example is, but anyway that might be fine.

.github/workflows/validate-reference-test-flow.yml
```diff
@@ -0,0 +1,34 @@
+name: "Validate Reference Test Flow"
```
DavePearce: Not sure about the name?  May just "Run One Blockchain Test` or "One Blockchain Test" or `A Blockchain Test`.  I guess it doesn't matter, but I don't find the name super enlightening.

reference-tests/build.gradle
```diff
@@ -36,6 +36,11 @@ configurations.configureEach {
 tasks.register("generateBlockchainReferenceTests", RefTestGenerationTask) {
   refTestTemplateFilePath = "src/test/resources/templates/BlockchainReferenceTest.java.template"
   refTests = "src/test/resources/ethereum-tests/BlockchainTests"
+  // Added to run validation on reference test flow - in this case, we only generate 1 test file
```
DavePearce: Ah, I see what you are doing.  You don't need to do this --- gradle already has support for running a single test with the `--tests` option.  See above.

.github/workflows/validate-reference-test-flow.yml
DavePearce: For reference, `BlockchainReferenceTest_104` is the one which contains the test you chose ... `stChainId`.

```
grep "stChainId" `find * -name "BlockchainReferenceTest_*"`
```

.github/workflows/validate-reference-test-flow.yml
amkCha: Changed the naming :ok_hand:

.github/workflows/gradle-ethereum-tests.yml
amkCha: You're right, I was just trying to make it shorter than 5h30 min. And 40GB x 3 had worked on my test branch prior to merging. But in any case, I've changed it to your suggested 4 x 20GB and am waiting for an answer to see if we can just scrap the State test flow alltogether if it's a duplicate :ok_hand:

reference-tests/build.gradle
amkCha: So what I was trying to do is not just launch a single test but also generate the file of only 1 test i/o generating 500 test files and running only one. But in any case, might be better to generate all the files to mimick the real process that launches every evening as much as possible. It adds a minute to the flow which makes it 5 and is still fair

.github/workflows/validate-reference-test-flow.yml
amkCha: As commented below, I was only generating 1 test file in the `generated` folder, so it was launching only 1 test as it only had one available, but I've changed that in the end (see comment below)

DavePearce (approved):
LGTM

PR #1905, opened 2025-03-13
feat: remove tracing module line counts from metadata

### LINKED ISSUE #1904: feat: remove tracing module line counts from metadata
The purpose of this is just avoid a potentially hazardous use of the metadata.  Specifically, using the line count of a tracing module for anything other than simple diagnostics is dangerous.  Furthermore, the line count information for all tracing modules is already available in the trace itself.  Therefore, removing this from the metadata just prevents a potential mishap.

DavePearce:
This PR removes line counts from the embedded metadata for _tracing_ modules.  Thus, only line counts for non-tracing modules is retained.


letypequividelespoubelles (approved):
LGTM

PR #1907, opened 2025-03-13
feat: remove ContinuousTracingPlugin

### LINKED ISSUE #1906: chore: remove `ContinuousTracingPlugin`
I think this plugin is not used (am just trying to check this) and, hence, it should be removed.

DavePearce:
This simply removes all the code related to the continuous tracing plugin which was previously used on the shadow node.


letypequividelespoubelles (approved):
RIP shadow node

PR #1909, opened 2025-03-14
fix: L2L1 config for tracing and counting RPC

### LINKED ISSUE #1908: fix L2L1 config for `GenerateLineCountsV2`

letypequividelespoubelles:


DavePearce (approved):
LGTM. I guess it means whoever starts the plugins will need to provide the CLI arguments now? I guess whoever that is will figure it out!

DavePearce:
Also you are sure not necessary for `GenerateConflatedTracesV2`? I would personally do that as well unless there is a good reason not to.

amkCha (approved):

PR #1920, opened 2025-03-18
feat: add manual release workflow

### LINKED ISSUE #1919: feat: trigger release workflow on manual release
The goal here is to change the release workflow so that, instead of being triggered by a `git push --tags` event, it is triggered by the manual creation of a release.  This looks to be possible via the [`release` trigger](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release).

DavePearce:
This workflow is intended eventually to replace the existing release workflow.  Instead of creating a release, this reacts to a manually created release by uploading the necessary assets.

### DIFF
--- a/.github/workflows/manual-release.yml
+++ b/.github/workflows/manual-release.yml
@@ -1,0 +1,67 @@
+# The purpose of this workflow is to upload the necessary assets when
+# a release is made manually on github.
+name: Manual Release Assets
+
+on:
+  release:
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: Specify tag where assets should be added.
+        required: true
+        type: string
+        default: "v1.2.3"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          submodules: false
+
+      - name: Setup Environment
+        # setup go-corset
+        uses: ./.github/actions/setup-environment
+        
+      - name: Configure Tag
+        id: config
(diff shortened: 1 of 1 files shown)


OlivierBBB:
Will this work in a way where you can manually do a release from a given branch / PR ?

DavePearce:
> Will this work in a way where you can manually do a release from a given branch / PR ?

@OlivierBBB Well, it works when you goto [this page](https://github.com/Consensys/linea-tracer/releases) and select "draft a new release".  When you do that, you have the option to use a different branch (which I hadn't noticed before)!  So I guess the answer to your question is yes (though not sure I properly understood your question).

OlivierBBB:
This being a workflow I imagined one would be triggering it from the GitHub Actions page, possibly with a drop down menu to select the branch of which you want to do a release. So far we've mostly used `arith-dev` but we did some releases from feature branches recently for @bogdanbear and @gusiri which is why I'm asking.

I'm just asking to understand how we will trigger it and what our options will be.

.github/workflows/manual-release.yml
```diff
@@ -0,0 +1,67 @@
...
+
+on:
+  release:
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: Specify tag where assets should be added.
+        required: true
+        type: string
+        default: "v1.2.3"
```
letypequividelespoubelles, line 13: isn't dangerous to have a default ?

letypequividelespoubelles (approved):
trust you :)

PR #1934, opened 2025-03-20
feat: tweak weekly tests

DavePearce:
Ready to merge.


.github/workflows/gradle-weekly-tests.yml
```diff
@@ -11,9 +11,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  weekly-tests:
+  weekly-replay-tests:
```
amkCha: Are they replay ? Just checking

.github/workflows/gradle-prc-call-tests.yml
```diff
@@ -0,0 +1,48 @@
+name: Weekly Precompile Call Tests
```
amkCha: It was probably done on purpose, but do you think we should separate those weeklys from the others ? It's having us manage another flow and another file. They could run on the same github-action with 2 runners in // though and upload 2 different reports

.github/workflows/gradle-weekly-tests.yml
DavePearce: Hmmm, I thought they were.  Based on what Olivier said last night though, you have a point.  I will check that now.

.github/workflows/gradle-prc-call-tests.yml
DavePearce: So, the reason I did this is that I believe (perhaps incorrectly) that the test profile for the weeklys is quite different from the PRC call tests.  However, I will look at that now, since I previously thought the other weeklys were all replays.  Key point being that we cannot run the replays with `X_PARALLELISM=26`.

DavePearce: Ok, so I think you are right here --- nice one.  We can potentially do the whole lot under a single "weekly" tag.  I will try to set that up now.

DavePearce: Done. Lets see if it works.

.github/workflows/gradle-weekly-tests.yml
DavePearce: Done.  No longer marked as replay.

amkCha (approved):

PR #1936, opened 2025-03-21
feat: TRM redesign

### LINKED ISSUE #1927: TRM redesign

letypequividelespoubelles:


DavePearce (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/WcpCall.java
```diff
@@ -0,0 +1,68 @@
...
+@Accessors(fluent = true)
+public class WcpCall {
+  private final UnsignedByte instruction;
+  private final Bytes arg1Hi;
+  private final Bytes arg1Lo;
+  private final Bytes arg2Hi;
+  private final Bytes arg2Lo;
+  private final boolean result;
+
+  public WcpCall(Wcp wcp, byte instruction, Bytes arg1, Bytes arg2) {
```
amkCha, line 36: I think we needed that method here as well https://github.com/Consensys/linea-tracer/blob/arith-dev/arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/BlockdataOperation.java#L264 
Maybe we add a TODO to refacto and use that method after you merge :ok_hand:

amkCha (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/WcpCall.java
letypequividelespoubelles: We need it not only for blockdata, but almost all modules calling WCP, so euc, mmio, txndata, hub, mxp, shakira, blakemodexp, ecData, oob, blockhash, soon rlpTxn ...

PR #1955, opened 2025-04-01
feat: improve weekly tests ci utilisation

### LINKED ISSUE #1952: feat: improve weekly tests CI utilisation
At the moment, the CI utilisation for the weekly tests is not optimal.  Specifically:

* **(Weekly Unit Tests)** these exhibit a very low amount of parallelism for the majority of the time.

* **(Weekly PRC CallTests)** these have good parallelism upto 20cores (which is our self imposed limit) but very modest memory requirements.  In theory, we should easily be able to hit 32cores and peak performance.

DavePearce:


arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/sixtyThreeSixtyFourths/SixtyThreeSixtyFourthsTests.java
```diff
@@ -80,6 +82,7 @@
  */
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@Execution(ExecutionMode.CONCURRENT)
 public class SixtyThreeSixtyFourthsTests {
```
lorenzogentile404, line 81: `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` seems to be not necessary anymore as the method sources may be static.

arithmetization/src/test/java/net/consensys/linea/zktracer/SignedOperationsExtensiveTest.java
```diff
@@ -51,16 +49,18 @@ void signedComparisonsModDivTest(OpCode opCode, String a, String b) {
     bytecodeRunner.run();
   }
 
-  Stream<Arguments> signedComparisonsModDivTestSource() {
+  private static Stream<Arguments> signedComparisonsModDivTestSource() {
+    Random rng = new Random(123);
+
```
lorenzogentile404, line 54: What is the advantage of declaring `rng` inside the method source here? Then we need to pass it every time to `randomBytes`, that seems to be not ideal. Is there an advantage from an efficiency perspective?

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/sixtyThreeSixtyFourths/SixtyThreeSixtyFourthsTests.java
DavePearce: Ok, because they no longer use a `static Random` field?  At least I can't see one.

arithmetization/src/test/java/net/consensys/linea/zktracer/SignedOperationsExtensiveTest.java
DavePearce: Well, there are pros / cons.  This is just one way to solve the problem.  The other is to use `@Execution(ExecutionMode.CONCURRENT)` which I did in other cases.  We can do that here, though passing `rng` to `randomBytes` everytime is not a big deal from my perspective.  Personally, if it were me, I would remove as many annotations as possible and stick with the simplest approach available.  Here, we have removed `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` which, already, caused problems (where as for the other examples we have to add another annotation to compensate).  I did add in `@Accessors(fluent = true)` and `@ExtendWith(UnitTestWatcher.class)` to be consistent with other tests, though I am unsure whether they are strictly necessary.  If not, they should be removed across the board.

arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/sixtyThreeSixtyFourths/SixtyThreeSixtyFourthsTests.java
DavePearce: I will remove this then.

lorenzogentile404 (approved):
LGTM

PR #1965, opened 2025-04-03
EIP-3855-PUSH0

### LINKED ISSUE #1962: EIP-3855: PUSH0
spec: https://github.com/Consensys/linea-specification/issues/63

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
```diff
@@ -77,7 +76,7 @@ public static OpCodeData forNonOpCodes(int value) {
    * @return <code>true</code> if this opcode is a <code>PUSHx</code>
    */
   public boolean isPush() {
-    return (0x60 <= value) && (value < 0x80);
+    return (EVM_INST_PUSH0 <= value) && (value <= EVM_INST_PUSH32);
```
letypequividelespoubelles: real change

arithmetization/src/main/resources/opcodes.yml
```diff
@@ -579,6 +579,15 @@ opcodes:
       delta: 0
       staticGas: G_JUMP_DEST
       flag4: true
+  - mnemonic: PUSH0
```
letypequividelespoubelles: real change

OlivierBBB: LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/RamSettings.java
OlivierBBB, line 1: Was this not used anymore ?

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
OlivierBBB: How is this `isPush()` method being used ? We for sure don't want to use it in the ROM.

OlivierBBB (approved):
LGTM modulo the question of `isPush()`.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/RamSettings.java
letypequividelespoubelles: I guess it was used at the time in the instruction decoder / hub. It was used to tell where the exo data is comming from for each opcode. But now it's not used at all, dead code.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
letypequividelespoubelles: it's used in the inst decoder and in the hub. 

I continue conv in slack

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/StackFragment.java
```diff
@@ -137,21 +138,21 @@ private StackFragment(
     }
 
     this.commonFragmentValues = commonFragmentValues;
-    this.pushValue = opCode.isPush() ? EWord.of(getPushValue(hub)) : EWord.ZERO;
+    this.pushValue = opCode.isPushNotZero() ? EWord.of(getPushValue(hub)) : EWord.ZERO;
   }
 
   private Bytes getPushValue(Hub hub) {
-    checkState(hub.opCode().isPush());
+    checkState(hub.opCode().isPushNotZero());
```
OlivierBBB, line 145: I would rename it to `isNontrivialPush`

OlivierBBB: I would rename it to `isNontrivialPush`

arithmetization/src/main/resources/opcodes.yml
OlivierBBB: Won't this clash with the London instruction decoding ?

**Edit.** I see that you have an `@Override tracePushZero` so no.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
```diff
@@ -67,21 +66,24 @@ public static OpCodeData forNonOpCodes(int value) {
...
    * @return <code>true</code> if this opcode is a <code>PUSHx</code>
    */
-  public boolean isPush() {
-    return (0x60 <= value) && (value < 0x80);
+  public boolean isPushNotZero() {
+    return (EVM_INST_PUSH1 <= value) && (value <= EVM_INST_PUSH32);
+  }
+
+  public boolean isPushZero() {
+    return EVM_INST_PUSH0 == value;
```
OlivierBBB: Purely stylistic comment, I would prefer
```java
return value == EVM_INST_PUSH0;
```

arithmetization/src/main/resources/opcodes.yml
OlivierBBB: I would have expected that we have several files
```
opcodes_LONDON.yml
opcodes_SHANGHAI.yml
...
```

OlivierBBB: Also the MXP pricing of `CREATE(2)` will change with Shanghai. I would expect it's easier to have distinct yml files.

OlivierBBB: Another point is e.g. in Paris we will switch names for `DIFFICULTY` / `PREVRANDAO`.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
```diff
@@ -67,21 +66,24 @@ public static OpCodeData forNonOpCodes(int value) {
             false,
             false,
             false),
-        new RamSettings(DataLocation.NONE, DataLocation.NONE),
         new Billing(GasConstants.G_ZERO, BillingRate.NONE, MxpType.NONE));
   }
 
   /**
-   * A method singling out <code>PUSHx</code> instructions.
+   * A method singling out <code>PUSHx</code> with X != 0 instructions.
```
OlivierBBB: Grammar comment
```java
A method singling out <code>PUSHX</code> instructions with X != 0.
```

OlivierBBB (dismissed):
LGTM but I think it would be cleaner if we used separate `opcodes_HARDFORKNAME.yml` files and used the `Fork`/`ChainConfig`/`ProtocolSpec` or what not to select the correct yml file to build the instruction decoder.

Also I take it that (unexceptional) `PUSH_POP` instructions are just stack instructions we don't need to modify their tracing ? E.g. in the `PushPopSection` if there even is such a thing ?

arithmetization/src/main/resources/opcodes.yml
letypequividelespoubelles: yout two last point are indeed good point ...

testing/src/main/java/net/consensys/linea/testing/BytecodeCompiler.java
```diff
@@ -44,6 +45,7 @@ private BytecodeCompiler() {}
    * @return an instance of {@link BytecodeCompiler}
    */
   public static BytecodeCompiler newProgram() {
+    loadOpcodes(LONDON);
```
letypequividelespoubelles: if we don't do this, then the (static) opcodeDataList is not set with set of (fork dependant) opcode list in the testing framework

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -82,6 +83,7 @@ public ZkTracer(
    * @param chain
    */
   public ZkTracer(ChainConfig chain) {
+    loadOpcodes(chain.fork);
```
letypequividelespoubelles, line 86: this is the (only ?) main change : the list of opcode is generated using a forkOpcode.yaml

testing/src/main/java/net/consensys/linea/testing/BytecodeCompiler.java
DavePearce: Why not take the fork as a parameter?

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
DavePearce: Where did you get the `yaml` files from?  Did you write them or are they from e.g. BESU?

DavePearce (approved):
LGTM

testing/src/main/java/net/consensys/linea/testing/BytecodeCompiler.java
letypequividelespoubelles: Thanks for the question :) 
Indeed I would prefer to have only one parameter, instead of here and in ToyExecV2. 
That's done.

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
letypequividelespoubelles: yaml files are ours, it's in arith/src/main/ressources. They are part of modified files of this PR

Issue #1153, opened 2024-09-10
Precompile tests where we run up agains the (63/64)-ths stuff


OlivierBBB:
We need to write tests with very simple bytecode. The specifics of these tests are as follows:
- message call transaction to a very simple contract
- bytecode: some pushes followed by e.g. `GAS` followed by `CALL` / `STATICCALL` to a precompile
- provide the transaction with a very low amount of gas

The goal is to trigger the `(63/64)-ths` business. We want to do this so that the `callee` will not meet its gas requirements.

In other words we must extract several unit tests (for all precompiles) from the issue #1136. See the documentation.

This is more advanced version of the test described  in https://github.com/Consensys/linea-tracer/issues/1730. In the present test, we **_cannot_** give enough gas to run the precompile, while in the simpler version we deliberately do not give enough gas.

OlivierBBB:
Gas calculation

```rust
int gasPreCall = frame.getRemainingGas() // prior to the CALL-type instruction processing
int gasUpfront =
    (value == 0)
        ? 100 // precompiles are warm
        : (precompileExistsInState)
            ? 9_000 + 100
            : 25_000 + 9_000 + 100;

checkState(gasPreCall >= gasUpfront);

int x = gasPreCall - gasUpfront;
int k = x / 64;
int l = x - 64 * k;
// x = 64 * k + l with 0 ≤ l < 64 is the euclidean division
// this k and l is what we are after

int stipend = (value == 0) ? 0 : 2_300;

int calleeGas = 63 * k + l + stipend;
```
You also compute the precompile cost and the desired `targetCalleeGas `
```rust
int prcCost = precompileExecutionCost() // cost of precompile execution
int targetCalleeGas = switch (costParam) {
    case COST_MO -> prcCost - 1;
    case COST -> prcCost;
}
```

--------------

At this point the goal is to work backwards from the desired conclusion:
```rust
63 * k + l + stipend = targetCalleeGas
0 ≤ k
0 ≤ l < 64
```
You work backwards. The above gets you uniquely the right `k` and `l`. Then you deduce the `x`, and from `x` you deduce `gasPreCall`. At that point you add up all the previous costs (the pushes, the initial memory expansion so as to incur no memory expansion during the `CALL`-type instruction, the `21_000` for the transaction) and you get the transaction gas limit.

--------

[...]

PR #1978, opened 2025-04-09
Feat/multi exception tests

### LINKED ISSUE #1838: MultiExceptionTests
Follow up on the OOGX task https://github.com/Consensys/linea-tracer/pull/1783

`MultiExceptionsTest`  could have a simple test where one triggers simultaneously

- [x] for `outOfGasExceptionLog0`, see https://github.com/Consensys/linea-tracer/pull/1783/files#r1962176828

```
staticException
outOfGasException
```

or

```
staticException
stackUnderflowException
```

One needs two smart contracts with byte code

```java
BytecodeCompiler simpleLog0 = BytecodeCompiler.newProgram();
simpleLog0.push(0).push(0).op(LOG0);

BytecodeCompiler stackUnderflowLog0 = BytecodeCompiler.newProgram();
stackUnderflowLog0.push(0).op(LOG0);
```

Which are called upon through a STATICCALL with gas = 3 + 3 + 375 + cornerCase / gas = 3 + 375 + cornerCase.

- [x] for `outOfGasExceptionJumpi`, see https://github.com/Consensys/linea-tracer/pull/1783/files#r1962163327

test simultaneous exceptions, in this case:
```
jumpException
outOfGasException
```
And it would be trivial to implement, just switch the second push to
```java
    final Bytes bytecode =
        BytecodeCompiler.newProgram()
            .push(1) // pc = 0, 1
            .push(6) // pc = 2, 3
            // ...
```

- [x] for `outOfGasExceptionReturnDataCopy`, see https://github.com/Consensys/linea-tracer/pull/1783/files#r1962067971

add a variant of the above where we do the following
```java
program
        .push(0) // byte size of return data
        .push(0) // retOffset
        .push(0) // byte size calldata
        .push(0) // argsOffset
        .op("c0de") // address of 'return data provider' account
        .op(OpCode.GAS) // gas
        .op(OpCode.STATICCALL)
        // 3. Clean the stack
        .op(OpCode.POP)
        // 4. Doubly exceptional return data copy
        .op(OpCode.RETURNDATASIZE)
        .push(1)
        .op(OpCode.ADD); // size = RDS + 1, which will trigger the `returnDataCopyException`
        .push(0) // offset
        .push(65) // destoffset, trigger mem expansion
        .op(OpCode.RETURNDATACOPY);
```
We could name this test to `doubleExceptionReturnDataCopyTestOogxAndRdcx`. We have so far not written tests where we deliberately trigger several exceptions at once.

amkCha:
Closes #1838 

In this test, we trigger all subsets possible of exceptions (except stack exceptions) at the same time for each opcode.
List of the combinations tested below
- RDCX & OOGX : RETURNDATACOPY
- RDCX & MXPX : RETURNDATACOPY
- JUMPX & OOGX : JUMP, JUMPI
  - triggering JUMPX by jumping outside of the bytecode
  - triggering JUMPX by jumping within the bytecode but not landing on a valid jump destination
- STATIC & OOSX : SSTORE
- STATIC & OOGX : LOG0, LOG1, LOG2, LOG3, LOG4, SSTORE, SELFDESTRUCT, CREATE, CREATE2, CALL
- STATIC & MXPX : LOG0, LOG1, LOG2, LOG3, LOG4, CREATE, CREATE2, CALL
- STATIC & ROOB : LOG0, LOG1, LOG2, LOG3, LOG4, CREATE, CREATE2, CALL
- InvalidCodePrefix & OOGX : RETURN
- InvalidCodePrefix & MaxCodeSize : RETURN
- MaxCodeSize & OOGX : RETURN
- InvalidCodePrefix & MaxCodeSize & OOGX : RETURN

Note : As MXPX is a subcase of OOGX, we don't test MXPX & OOGX
Note2 : For Shanghai, will need to add combinations with initcodesize exception for CREATE and CREATE2


arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/OutOfGasExceptionTest.java
```diff
@@ -268,7 +269,8 @@ void outOfGasExceptionJumpi(int cornerCase) {
     if (cornerCase == -1) {
       // JUMPI needs JUMPDEST to jump to
       // Calculate the gas cost to trigger OOGX on JUMPI and not on the last but one opcode
-      gasCost = GAS_CONST_G_TRANSACTION + 2 * GAS_CONST_G_VERY_LOW + GAS_CONST_G_HIGH;
+      // 21000L intrinsic gas cost + 3L PUSH + 10L JUMP, and we retrieve 1
```
amkCha, line 272: Fix on gas cost calculation to trigger OOGX on JUMP and not JUMPDEST

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
```diff
@@ -18,9 +18,30 @@
...
 public class ExceptionUtils {
+
+  public static Bytes address1 =
+      Bytes.fromHexString("0x1FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF");
+  public static Bytes address2 =
+      Bytes.fromHexString("0x2FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF");
+  public static Bytes address3 =
+      Bytes.fromHexString("0x3FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF");
+  public static Bytes address4 =
+      Bytes.fromHexString("0x4FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF");
```
OlivierBBB, line 40: Addresses are 20 byte integers and if they are larger (32 bytes) they get trimmed of their 12 leading bytes. In other words the four addresses above are the same
```rust
raw address:     0x?FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
trimmed address:                         0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```
To have distinct addresses you should use
```rust
// distinct addresses
address1 = 0x1FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
address2 = 0x2FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
address3 = 0x3FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
address4 = 0x4FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```

OlivierBBB: You seem to be using these fields as topics of `LOGX` instructions, in which case the above issue doesn't apply. But it would make sense to rename them to `topicX`.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
```diff
@@ -33,4 +54,165 @@ static void assertEqualsOutOfGasIfCornerCaseMinusOneElseAssertNotEquals(
...
+        .push(0) // byte size of return data
+        .push(0) // retOffset
+        .push(0) // byte size calldata
+        .push(0) // argsOffset
+        .push("c0de") // Address of account
+        .push(gas) // gas
+        .op(OpCode.STATICCALL);
+  }
+
+  public static BytecodeCompiler getPgStaticCallToCodeAccount() {
```
OlivierBBB: What does `pg` stand for ? Program ?

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
```diff
@@ -33,4 +54,165 @@ static void assertEqualsOutOfGasIfCornerCaseMinusOneElseAssertNotEquals(
...
+    return program;
+  }
+
+  public static BytecodeCompiler getPgCreateInitCodeWithReturnStartByteAndSize(
+      String hexString, int returnSize) {
+    BytecodeCompiler initProgram = BytecodeCompiler.newProgram();
+    initProgram
+        .push(hexString)
+        .push(0)
+        .op(OpCode.MSTORE8)
```
OlivierBBB, line 214: Given the `MSTORE8` operation it would make sense to replace the `hexString` with a `byte` or an `int` (with a check of the form `0 ≤ x < 256`)

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
```diff
@@ -33,4 +54,165 @@ static void assertEqualsOutOfGasIfCornerCaseMinusOneElseAssertNotEquals(
...
+          .op(OpCode.CREATE2);
+      default -> {}
+    }
+
+    return program;
+  }
+
+  public static BytecodeCompiler getPgCreateInitCodeWithReturnStartByteAndSize(
+      String hexString, int returnSize) {
+    BytecodeCompiler initProgram = BytecodeCompiler.newProgram();
```
OlivierBBB, line 210: One can add a little documentation of the form:
```java
/**
* The {@code initProgram} inserts a single byte {@code x} at offset 0 into RAM
* and returns {@code returnSize} bytes starting at offset 0. There are four cases
* of interest:
* <ul> 
* <li> {@code x} is <b>0xEF</b> and {@code returnSize} > 0</li>
* <li> {@code x} is <b>0xEF</b> and {@code returnSize} = 0</li>
* <li> {@code x} is anything but <b>0xEF</b> and {@code returnSize} > 0</li>
* <li> {@code x} is anything but <b>0xEF</b> and {@code returnSize} = 0</li>
* </ul>
* Only the first one should raise the <b>invalidCodePrefixException</b>.
*/
```

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
```diff
@@ -0,0 +1,461 @@
...
+ */
+
+@ExtendWith(UnitTestWatcher.class)
+public class MultiExceptionTest {
+
+  @Test
+  void rdcAndOogExceptionsReturnDataCopy() {
+    boolean MXPX = true;
+    boolean RDCX = true;
+    final ToyAccount codeProviderAccount = getAccountForCodeAddress(return32BytesFFBytecode);
```
OlivierBBB: The name `getAccountForCodeAddress` is confusing to me. What this does is create (and return) an account with given bytecode at address `c0de`. Maybe `endowCodeAddressWithGivenByteCode(...)` or so

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
```diff
@@ -0,0 +1,461 @@
...
+  }
+
+  @Test
+  void invalidCodePrefixAndOogExceptionForCreate() {
+    // We run gas cost calculation on program without Invalid Code Prefix exception
+    String startByte = Integer.toHexString(0xee);
+    BytecodeCompiler programWithoutICP =
+        getPgCreateInitCodeWithReturnStartByteAndSize(startByte, 1);
+    BytecodeRunner bytecodeRunner = BytecodeRunner.of(programWithoutICP.compile());
+    long gascost = bytecodeRunner.runOnlyForGasCost();
```
OlivierBBB: I am slightly concerned about this part (though it seems to work.) The issue I have was that by running `runOnlyForGasCost` you don't have fine grain control over what you provide the `CREATE`. The problem is that what you provide the child context of a CREATE(2) is 63/64ths of the left over gas after paying for the upfront gas cost of the CREATE(2) opcode. But the 63/64 ths function may take on the same value for `g` and `g - 1` for the right value of `g := remainingGas`, see my comment here: https://github.com/Consensys/linea-tracer/issues/1153.

OlivierBBB: Note that this issue of

> sometimes `g` and `g - 1` return the same value under the map `x -> x - (x/64)`

does not arise with `g` and `g - 2`. So maybe the answer is to systematically use `cornerCase = -2, 0, +2` instead of `cornerCase = -1, 0, +1`.

OlivierBBB: But this will only work for a single layer of `CALL`'s / `CREATE`'s

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
```diff
@@ -0,0 +1,461 @@
...
+        .push(initProgramByteSize)
+        .push(0)
+        .push(0)
+        .op(OpCode.CREATE);
+
+    BytecodeRunner bytecodeRunner = BytecodeRunner.of(program.compile());
+    // We run the program with a gas cost that triggers OOGX
+    // We calculate all opcodes gas before the RETURN opcode
+    // 32027L = 3L PUSH + 3L PUSH + 6L MSTORE + 3L PUSH + 3L PUSH + 3L PUSH + 32000L CREATE +
+    // (7/64)(negligible) + 3L PUSH + 3L PUSH
```
OlivierBBB: What does `(7/64)(negligible)` refer to ? Same question below about the `(16/64)(negligible)` line.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
```diff
@@ -33,4 +54,165 @@ static void assertEqualsOutOfGasIfCornerCaseMinusOneElseAssertNotEquals(
...
+        .op(OpCode.GAS) // gas
+        .op(OpCode.STATICCALL);
+  }
+
+  public static BytecodeCompiler getProgramRDCFromStaticCallToCodeAccount(
+      boolean withRDCX, boolean withMXPX) {
+    // if withMXPX, we set an offset to trigger MXPX else regular MXP
+    Bytes offsetRDC =
+        withMXPX
+            ? Bytes.fromHexStringLenient("0xFFFFFFFF")
```
OlivierBBB: (From memory) the `MXPX` exception actually triggers when either
```rust
- size ≥ 256^4
- size ≠ 0 and offset ≥ 256^4
```
so is `0xFFFFFFFF` enough or should we use `0x0100000000` instead ?

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
```diff
@@ -33,4 +54,165 @@ static void assertEqualsOutOfGasIfCornerCaseMinusOneElseAssertNotEquals(
...
+    if (withRDCX) {
+      programStartWithStaticCall
+          .push(1)
+          .op(OpCode.ADD); // size = RDS + 1, which will trigger the `returnDataCopyException`
+    }
+    programStartWithStaticCall
+        .push(0) // offset
+        .push(offsetRDC) // destoffset, trigger mem expansion
+        .op(OpCode.RETURNDATACOPY);
+    // Bytes.fromHexStringLenient("0xFFFFFFFF")
```
OlivierBBB: If the above comment is correct then this needs to be updated, too.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
```diff
@@ -0,0 +1,461 @@
...
+  }
+
+  @Test
+  public void staticAndOutOfSStoreExceptions() {
+    BytecodeCompiler pg = BytecodeCompiler.newProgram();
+
+    pg.push(0).push(0).op(OpCode.SSTORE);
+
+    int gasCostToTriggerOutOfSStore = 3 + 3 + GAS_CONST_G_CALL_STIPEND - 1;
+    // 21000L is the intrinsic gas cost of a transaction and 3L is the gas cost of PUSH1
```
OlivierBBB: Comment looks obsolete

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
```diff
@@ -0,0 +1,461 @@
...
+
+    BytecodeCompiler program = simpleProgramEmptyStorage(opCode);
+    Bytes pgCompile = program.compile();
+    BytecodeRunner bytecodeRunner = BytecodeRunner.of(pgCompile);
+    long gasCostTx = bytecodeRunner.runOnlyForGasCost();
+
+    /*
+    for CREATE/CREATE2, Static Exception happens before deployment, so we test OOGX before deployment
+    We remove 6400 (depositFee) + deployment code exec cost (18) from gas cost calculated
+     */
```
OlivierBBB: I don't understand why the `depositFee` enters the picture here. The `depositFee` gets paid at the end of the lifetime of e.g. a `CREATE`-spawned execution context. It is associated with a `RETURN` at the end of the life-cycle of a deployment. For this fee to be paid you have to first enter said execution context.

Here we are attempting to test the collision of a `staticException` and an `outOfGasException` at the moment of the `CREATE(2)` opcode. The opcode that must throw these exceptions. In particular the plan is to never enter the deployment context (wherein a `depositFee` may apply at the end.)

So the simplest way to create this exception is to do
```
// root context of transaction, call stack depth = 0
// do various PUSH'es
STATICCALL(address = "0xc0de", gas = gasParam, callData = {}, returnAt = {})

// c0de execution in a static context, call stack depth = 1
PUSH1 0x00
PUSH1 0x00
PUSH1 0x00 // remainingGas = gasParam - 3 * 3
CREATE

// would cost 32_000, but violates static context
// gasParam should therefore be
// 3 * 3 + 32_000 - 1 or so if we want to simultaneously 
// have staticException & outOfGasException
```

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
```diff
@@ -0,0 +1,461 @@
...
+  static Stream<OpCode> opCodesForStaticAndOogExceptionList() {
+    List<OpCode> opCodesListArgument =
+        Arrays.asList(
+            /*            OpCode.LOG0,
+            OpCode.LOG1,
+            OpCode.LOG2,
+            OpCode.LOG3,
+            OpCode.LOG4,
+            OpCode.SSTORE,
+            OpCode.SELFDESTRUCT,*/
```
OlivierBBB: Should this be uncommented ?

OlivierBBB (dismissed):
Various comments. I have not yet reviewed all of it. I would suggest breaking the main test class into smaller files, just to simplify navigation etc. This can be done either
- by opcode (family)
- by combination of exceptions

The first option would seem the simplest mentally.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
amkCha: refactored it with signature `getAccountForAddressWithBytecode(Address addr, Bytes bytecode)` to make it explicit  :+1:

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: you're right, it's renamed :ok_hand:

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: I renamed the function to `getProgramStaticCallToCodeAccount`

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: Done, and added the check

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: thanks for writing it, added in the file :)

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: Removed

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
amkCha: Omg, Yes !

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
amkCha: Yes, not very clear - it means that it's not going to round to even 1gas, ie < 0.5, so i'll write just that in the comment

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: Briefly reading the spec, that's what I see indeed - but `0xFFFFFFFF` is enough and does trigger MXPX - we might need to check that 
image

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
amkCha: If I understand your concern correctly, I think we are saying the same thing here actually (and it might also be me not being very clear in my comments). 

I run the whole program once to calculate the gas cost like so 
```
BytecodeCompiler program = simpleProgramEmptyStorage(OpCode.CREATE);
    Bytes pgCompile = program.compile();
    BytecodeRunner bytecodeRunner = BytecodeRunner.of(pgCompile);
    long gasCostTx = bytecodeRunner.runOnlyForGasCost();
```
as this program has gone until the end of the life-cycle of a deployment, I retrieve the amount that went into deployment
- 6400 (deposit fee)
- 18 (deployment code exec gas cost)
- 1 to trigger an OOGX
= 6419 which is my cornerCase for the `CREATE`
This triggers an OOGX at the moment of the `CREATE` opcode, simultaneous to the STATIC exception

(same logic for CREATE2 as well)

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
amkCha: you're right, for this one, as I cannot pilot the 63/64 because I don't go through a StaticCall, there needs to be a check to see if OOGX was triggered - to be safe (and not have to check), i'll go with `-2`

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: I updated the value to `0x0100000000`  to account for the MXP redesign :ok_hand:

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
```diff
@@ -33,4 +56,180 @@ static void assertEqualsOutOfGasIfCornerCaseMinusOneElseAssertNotEquals(
           bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
     }
   }
+
+  public static ToyAccount getAccountForAddressWithBytecode(Address addr, Bytes bytecode) {
+    return ToyAccount.builder()
```
lorenzogentile404, line 61: In case then you need this bytecode in memory, there is somewhere already implemented using `EXTCODECOPY`. I can find it, if needed.

lorenzogentile404 (approved):
LGTM.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: Yes, let me know how to put bytecode in memory - for this one it's generic to be able to call an account with a bytecode to execute with a static call for example

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MultiExceptionTest.java
amkCha: Validated on a call, added a comment

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/ExceptionUtils.java
amkCha: Actually, the MXPX is triggered by offset + size check, so as size is 31, we trigger the MXPX with both offset values (`0xFFFFFFFFF` and `0x01000000000`)

PR #1980, opened 2025-04-11
update besu and prepare for release

letypequividelespoubelles:


gradle.properties
OlivierBBB, line 1: This is the main  change to file, right ?

OlivierBBB (approved):

gradle.properties
letypequividelespoubelles: Yes

Issue #1974, opened 2025-04-08
[Arithmetization Testing] Create a test environment that uses an in thread Besu node


gauravahuja:
- Configure and start an in thread Besu node with linea tracing and shomei plugin. 
- Create a new test environment where 
  - Each unit test should start a fresh besu node
  - After processing the transaction in the test invoke an rpc to generate conflated trace file 
  - This trace file will be validated by corset.
  - Ensure easy migration of existing tests to this new test environment.

gauravahuja:
https://github.com/hyperledger/besu/pull/8572

PR #1981, opened 2025-04-14
ToyExecutionEnvironment can run transactions by starting a besu node

gauravahuja:
#1974 

- Create Besu GenesisConfig with accounts balances and code from the ToyExecutionEnvironment accounts.
- Spin up Besu Node running in a separate thread. This Besu node is setup
  - with the genesis config created in first step
  - with the tracer plugin
- Send transactions from the ToyExecutionEnvironment to the Besu Node via web3j APIs.
- Generate the conflated trace file for the blocks that contain the transactions.


gradle/tests.gradle
```diff
@@ -127,7 +127,25 @@ jacocoTestReport {
 }
 
 tasks.test.configure {
-  dependsOn(buildZkevmBin)
+  dependsOn(buildZkevmBin, "copyPlugins")
```
DavePearce, line 130: So, the purpose of `copyPlugins` is to place all of the plugins together in one place?

DavePearce (approved):
It looks fine to me, as far as I can tell.  I am unsure about how robust the RPC call mechanism will be.

gradle/tests.gradle
gauravahuja: Yes, for now it is just the tracer plugin, but later it will be updated to also download the shomei plugin. The plugins are put in one directory and when the besu node is spun up in each test it looks at this directory to find all the plugins to add.

PR #1988, opened 2025-04-22
1970 mxp redesign

### LINKED ISSUE #1970: `MXP` redesign
Implementation of https://github.com/Consensys/linea-specification/pull/103

lorenzogentile404:
Redesign of the Mxp module for Cancun where MCOPY is added

The Mxp operation can follow 5 scenarii depending on the opcode. Each scenario executes wcp and euc computations and inherits from the previous scenario as computations are cumulative

**1. MSize scenario**
- no computation

**2. Trivial scenario**
- computes size1IsZero and size2IsZero

**3. Mxpx scenario** 
- computes size1IsZero and size2IsZero
- computes mxpxExpression

**4. State update with word pricing scenario**
- computes size1IsZero and size2IsZero
- computes mxpxExpression
- computes state update (wordsNew,cMemNew) 
- and extraGasCost for word pricing opcodes

**5. State update with byte pricing scenario**
 - computes size1IsZero and size2IsZero
 - computes mxpxExpression
 - computes state update (wordsNew,cMemNew) 
 - and extraGasCost for byte pricing opcodes


arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/BlockdataOperation.java
```diff
@@ -305,6 +305,8 @@ private Bytes eucCall(int w, EWord arg1, EWord arg2) {
 
     res[w] = euc.callEUC(arg1, arg2).quotient();
 
+    exoInst[w] = UnsignedByte.of(EVM_INST_DIV);
```
letypequividelespoubelles: ? EUC has only one instruction, and it's not DIV

arithmetization/src/main/java/net/consensys/linea/zktracer/module/euc/EucCall.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.experimental.Accessors;
+import net.consensys.linea.zktracer.types.UnsignedByte;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+
+@Getter()
+@Accessors(fluent = true)
+public class EucCall {
+  private final UnsignedByte instruction;
+  private final Bytes arg1Hi;
```
letypequividelespoubelles: there are no high part for arg1 and arg2. EUC takes two long as input, it can't deal with bigger input

arithmetization/src/main/java/net/consensys/linea/zktracer/module/euc/EucCall.java
```diff
@@ -0,0 +1,37 @@
...
+  private final Bytes arg1Lo;
+  private final Bytes arg2Hi;
+  private final Bytes arg2Lo;
+  private final Bytes result;
+
+  public EucCall(Euc euc, Bytes arg1, Bytes arg2) {
+    // TODO : check instruction
+    this.instruction = UnsignedByte.of(EVM_INST_DIV);
```
letypequividelespoubelles: nope

arithmetization/src/main/java/net/consensys/linea/zktracer/module/euc/EucCall.java
letypequividelespoubelles: a long or a 8 Bytes

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpComputation.java
```diff
@@ -0,0 +1,182 @@
...
+import net.consensys.linea.zktracer.module.euc.Euc;
+import net.consensys.linea.zktracer.module.euc.EucCall;
+import net.consensys.linea.zktracer.module.hub.fragment.imc.MxpCall;
+import net.consensys.linea.zktracer.module.wcp.Wcp;
+import net.consensys.linea.zktracer.module.wcp.WcpCall;
+import net.consensys.linea.zktracer.opcode.OpCode;
+import org.apache.tuweni.bytes.Bytes;
+
+public class MxpComputation {
+  private final Wcp wcp;
```
letypequividelespoubelles: I'd prefer having a list of MxpExoCall, easier to read

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpComputation.java
```diff
@@ -0,0 +1,182 @@
...
+  private final Euc euc;
+  public final List<WcpCall> wcpCalls;
+  public final List<EucCall> eucCalls;
+  public final boolean[] wcpFlags;
+  public final boolean[] eucFlags;
+
+  @Builder
+  public MxpComputation(Wcp wcp, Euc euc, int nRows) {
+    this.wcp = wcp;
```
letypequividelespoubelles: I think we don't need to keep EUC and WCP in memory

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpOperationV3.java
```diff
@@ -0,0 +1,262 @@
...
+  private final int contextNumber;
+  private final long words;
+  private long wordsNew;
+  private final long cMem;
+  private long cMemNew;
+  private final Bytes gWord;
+  private final Bytes gByte;
+
+  public MxpOperationV3(final MxpCall mxpCall, Wcp wcp, Euc euc) {
+    this.wcp = wcp;
```
letypequividelespoubelles: I think we don't need to keep it in memory, we just need o WCP and EUC to trigger it, we don't need it after

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpOperationV3.java
```diff
@@ -0,0 +1,262 @@
...
+    this.cMem = memoryCost(this.mxpCall.getMemorySizeInWords());
+    this.cMemNew = memoryCost(this.mxpCall.getMemorySizeInWords()); // will (may) be updated later
+
+    this.gWord = this.mxpCall.getCostBy(BillingRate.BY_WORD);
+    this.gByte = this.mxpCall.getCostBy(BillingRate.BY_BYTE);
+
+    computationsAndUpdates();
+  }
+
+  public int ctMax() {
```
letypequividelespoubelles: this should be abstract

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpOperationV3.java
```diff
@@ -0,0 +1,262 @@
...
+
+  final void traceDecoder(int stamp, Trace.Mxp trace) {
+    OpCode opCode = this.mxpCall.getOpCodeData().mnemonic();
+
+    // Remove the nRows, there is one row ?
+    trace
+        .mxpStamp(stamp)
+        .cn(this.getContextNumber())
+        .decoder(true)
+        .macro(false)
```
letypequividelespoubelles: all rows with 0s or false can be skipped: the fillAndValidateRow puts 0 on every unfilled columns of the row. Easier to read

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpOperationV3.java
```diff
@@ -0,0 +1,262 @@
...
+          .cn(this.getContextNumber())
+          .decoder(false)
+          .macro(false)
+          .scenario(false)
+          .computation(true)
+          .ct(i)
+          .ctMax(ctMax())
+          .pComputationWcpFlag(this.mxpComputation.wcpFlags[i])
+          .pComputationEucFlag(this.mxpComputation.eucFlags[i])
+          .pComputationExoInst(
```
letypequividelespoubelles: don't need the switch here if we use a generic list of MxpExoCall list for the computations row

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpUtils.java
```diff
@@ -0,0 +1,52 @@
...
+  public static boolean isBytePricingOpcode(OpCode opCode) {
+    return opCode == OpCode.MLOAD
+        || opCode == OpCode.MSTORE
+        || opCode == OpCode.MSTORE8
+        || opCode == OpCode.REVERT
+        || opCode == OpCode.RETURN
+        || opCode.isCall();
+  }
+
+  // This is a copy and past from FrontierGasCalculator.java
```
letypequividelespoubelles: don't we already have this somewhere where we compute the exceptions ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpV3.java
```diff
@@ -0,0 +1,71 @@
...
+import net.consensys.linea.zktracer.container.stacked.ModuleOperationStackedList;
+import net.consensys.linea.zktracer.module.euc.Euc;
+import net.consensys.linea.zktracer.module.hub.fragment.imc.MxpCall;
+import net.consensys.linea.zktracer.module.wcp.Wcp;
+
+/** Implementation of a {@link Module} for memory expansion. */
+@Getter
+@Accessors(fluent = true)
+@RequiredArgsConstructor
+public class MxpV3 implements OperationListModule<MxpOperationV3> {
```
letypequividelespoubelles: MXP and MXPV3 shpould extends the same abstract class, so hat you don't have to copy paste the moduleKey, etc ...

See what's done for TxnData for exemple

letypequividelespoubelles (commented):
I'd suggest to have, for each Mxp instruction, a instMxpCall tht extends an abstract MxpCall where you compute only the needed stuff (see what has been done for OOB).

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockdata/BlockdataOperation.java
amkCha: Okay I wasn't sure, it's removed :ok_hand:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpComputation.java
amkCha: Added the class `MxpExoCall` mimicking  `OobExoCall`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpComputation.java
amkCha: You're right ! removed from Operation as well

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpOperationV3.java
amkCha: Yes, not sure why I was storing that tbh - it's removed

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpOperationV3.java
amkCha: Trimmed all the false and 0s lines, tracing is a bit less verbose now

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpOperationV3.java
amkCha: Done ! :+1:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpV3.java
amkCha: We now have abstract `Mxp` class and `LondonMxp` / `CancunMxp` that derive from it

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MemoryExpansionExceptionTest.java
```diff
@@ -52,7 +52,7 @@ public void memoryExpansionExceptionTest(boolean triggerRoob, OpCode opCode) {
         MEMORY_EXPANSION_EXCEPTION,
         bytecodeRunner.getHub().previousTraceSection().commonValues.tracedException());
     assertTrue(bytecodeRunner.getHub().mxp().operations().getLast().getMxpCall().isMxpx());
-    assertEquals(triggerRoob, bytecodeRunner.getHub().mxp().operations().getLast().isRoob());
+    // assertEquals(triggerRoob, bytecodeRunner.getHub().mxp().operations().getLast().isRoob());
```
amkCha: @lorenzogentile404 not sure why we comment this part ? Should we keep it while we run London version of Mxp ?

amkCha:
> I'd suggest to have, for each Mxp instruction, a instMxpCall tht extends an abstract MxpCall where you compute only the needed stuff (see what has been done for OOB).

Okay so to keep the computation in the `Mxp` folder and not have it in the Hub, I used the same logic but instead of an abstract `MxpCall`, I created an abstract class called `MxpScenario` and I derive the computation in each scenario that extends this class. Let me know if it makes it clear to read else I can always refactor to do it it with an `MxpCall` abstract class :ok_hand:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -1067,6 +1067,10 @@ protected TxnData setTxnData() {
     throw new IllegalStateException("must be implemented");
   }
 
+  protected Mxp setMxp() {
```
letypequividelespoubelles: I've modified in a ongoing PR the generic way to do it https://github.com/Consensys/linea-tracer/pull/1993/files#diff-15886f16c740b7fa4e5c3e00b0e8a017ca6b4a772df336008cfb313077a5cd88R1065

I think it's cleaner to declare those method abstract instead of base implementation throwing an exception. But I'm not a dev :)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/MxpCall.java
```diff
@@ -23,28 +23,35 @@
...
 public class MxpCall implements TraceSubFragment {
 
   public final Hub hub;
 
-  // filled in by MXP module
+  /** The following properties will be filled in by MXP module * */
+  /** - don't necessitate computation * */
   @Getter @Setter public OpCodeData opCodeData;
+
   @Getter @Setter public boolean deploys;
```
letypequividelespoubelles: instead of setter, won't it be cleaner for all those variables to be final, and set in the constructor of each super mxp call ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/MxpCall.java
```diff
@@ -63,6 +70,78 @@ public boolean getSize2NonZeroNoMxpx() {
...
+        getOpCodeData().billing().billingRate() == billingRate
+            ? getOpCodeData().billing().perUnit().cost()
+            : 0);
+  }
+
+  /**
+   * This method is called by the Mxp module to snapshot the data of the hub into the following
+   * mxpCall properties [opCodeData, deploys, memorySizeInWords, offset1, size1, offset2, size2]
+   */
+  public void fillNoComputationMxpProperties() {
```
letypequividelespoubelles: I would expect all this filling to be in the constructor of each type of MXP call instead

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
```diff
@@ -0,0 +1,107 @@
...
+        .arg2Lo(arg2B32.slice(LLARGE, LLARGE))
+        .resultA(booleanToBytes(wcp.callLT(arg1B32, arg2B32)))
+        .build();
+  }
+
+  public static MxpExoCall callToIsZero(final Wcp wcp, Bytes arg1) {
+
+    final Bytes32 arg1B32 = Bytes32.leftPad(arg1);
+
+    // TODO: check if I need arg2 set
```
letypequividelespoubelles: no you don't need, it won't be readen by WCP. So you can put whatever you want, so 0 is best

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  public static MxpExoCall callToEUC(final Euc euc, Bytes arg1, Bytes arg2) {
+
+    final Bytes32 arg1B32 = Bytes32.leftPad(arg1);
+    final Bytes32 arg2B32 = Bytes32.leftPad(arg2);
+
+    return MxpExoCall.builder()
+        .eucFlag(true)
+        .instruction(0)
```
letypequividelespoubelles: no inst needed

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
```diff
@@ -0,0 +1,107 @@
...
+        .arg1Hi(arg1B32.slice(0, LLARGE))
```
letypequividelespoubelles: hi part will be always 0 (so nothing to do, as it's the defaukt builder). You just have to write the arg in argLo (or maybe tr leading 0s, but I think it's not needed)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Builder.Default private final boolean wcpFlag = false;
+  @Builder.Default private final boolean eucFlag = false;
+  @Builder.Default private final int instruction = 0;
+  @Builder.Default private final Bytes arg1Hi = Bytes.EMPTY;
+  @Builder.Default private final Bytes arg1Lo = Bytes.EMPTY;
+  @Builder.Default private final Bytes arg2Hi = Bytes.EMPTY;
+  @Builder.Default private final Bytes arg2Lo = Bytes.EMPTY;
+  // results for wcp computations in trace
+  @Builder.Default private final Bytes resultA = ZERO;
```
letypequividelespoubelles, line 44: why is it a Bytes ? The result of WCP is always a boolean

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/module/Mxp.java
letypequividelespoubelles, line 1: shouldn't thios class becoming abstract ? with LondonMxp and LondonMxp extending this class ?

call method should be abstract, as well as spillage, column headers, etc, as it's fork dependant.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleOperation/MxpOperation.java
```diff
@@ -0,0 +1,35 @@
...
+public abstract class MxpOperation extends ModuleOperation {
+
+  final MxpCall mxpCall;
+
+  public MxpOperation(MxpCall mxpCall) {
+    this.mxpCall = mxpCall;
+  }
+
+  public void trace(int stamp, Trace tr) {
+    throw new IllegalStateException("Should be implemented");
```
letypequividelespoubelles: abstract instead ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleScenario/MxpScenario.java
```diff
@@ -0,0 +1,102 @@
...
+import lombok.Getter;
+import net.consensys.linea.zktracer.module.euc.Euc;
+import net.consensys.linea.zktracer.module.hub.fragment.imc.MxpCall;
+import net.consensys.linea.zktracer.module.mxp.MxpExoCall;
+import net.consensys.linea.zktracer.module.wcp.Wcp;
+import net.consensys.linea.zktracer.opcode.OpCode;
+
+public abstract class MxpScenario {
+  /** Store all wcp and euc computations with params and results */
+  public final List<MxpExoCall> exoCalls;
```
letypequividelespoubelles: strange to have this in MxpScenario ... I would rather have this in CancunMxpOperation.

I would imagine MxpScenario to be just an enum (and still, I'm not sure it's still needed, as the tracing could be made in each super mxp call)

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCode.java
letypequividelespoubelles, line 1: you need to create a CancunOpcode.yml too (and check that it's created for CancunHub).

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleOperation/CancunMxpOperation.java
```diff
@@ -0,0 +1,199 @@
...
+
+  final void traceDecoder(int stamp, Trace.Mxpcan trace) {
+    OpCode opCode = this.mxpCall.getOpCodeData().mnemonic();
+
+    trace
+        .mxpStamp(stamp)
+        .cn(this.getContextNumber())
+        .decoder(true)
+        .pDecoderInst(UnsignedByte.of(opCode.byteValue()))
+        .pDecoderIsMsize(opCode == OpCode.MSIZE)
```
letypequividelespoubelles, line 68: can't this be done as you did for traceScenario ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleOperation/CancunMxpOperation.java
```diff
@@ -0,0 +1,199 @@
...
+          .ct(i)
+          .ctMax(scenario.ctMax())
+          .pComputationWcpFlag(scenario.exoCalls.get(i).wcpFlag())
+          .pComputationEucFlag(scenario.exoCalls.get(i).eucFlag())
+          .pComputationExoInst(scenario.exoCalls.get(i).instruction())
+          .pComputationArg1Hi(scenario.exoCalls.get(i).arg1Hi())
+          .pComputationArg1Lo(scenario.exoCalls.get(i).arg1Lo())
+          .pComputationArg2Hi(scenario.exoCalls.get(i).arg2Hi())
+          .pComputationArg2Lo(scenario.exoCalls.get(i).arg2Lo())
+          .pComputationResA(scenario.exoCalls.get(i).resultA().toLong())
```
letypequividelespoubelles: isn't it a boolean ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleOperation/CancunMxpOperation.java
```diff
@@ -0,0 +1,199 @@
...
+          .pComputationResB(scenario.exoCalls.get(i).resultB().toLong())
```
letypequividelespoubelles: You might have sign issues :) we have conversion Bytes to unsignedLong somewhere

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleScenario/MxpxMxpScenario.java
```diff
@@ -0,0 +1,63 @@
...
+  public void computeMxpxExpression(MxpCall mxpCall, Wcp wcp) {
+    // Row i + 3
+    exoCalls.add(
+        MxpExoCall.callToLEQ(wcp, mxpCall.getSize1(), Bytes.ofUnsignedLong(MXPX_THRESHOLD)));
+    // Row i + 4
+    exoCalls.add(
+        MxpExoCall.callToLEQ(wcp, mxpCall.getSize2(), Bytes.ofUnsignedLong(MXPX_THRESHOLD)));
+    // Row i + 5
+    exoCalls.add(
+        MxpExoCall.callToLEQ(wcp, mxpCall.getOffset1(), Bytes.ofUnsignedLong(MXPX_THRESHOLD)));
```
letypequividelespoubelles: this constant could be static

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleScenario/MxpxMxpScenario.java
```diff
@@ -0,0 +1,63 @@
...
+    // Row i + 6
+    exoCalls.add(
+        MxpExoCall.callToLEQ(wcp, mxpCall.getOffset2(), Bytes.ofUnsignedLong(MXPX_THRESHOLD)));
+
+    boolean size1IsNonZero = !this.size1IsZero;
```
letypequividelespoubelles: all of this could be final no ?

letypequividelespoubelles (commented):
just looked at the logic, not at the implementation

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
amkCha: ok thanks !

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
amkCha: Agreed, I'm switching to the abstract declaration for `setMxp` as well

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/module/Mxp.java
amkCha: Yes and done :+1:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleScenario/MxpxMxpScenario.java
amkCha: Made it for all intermediate variables in computations

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
amkCha: you're right, at one point I wanted to store the result of wcp and euc in the same variable but switched - it's fixed !

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleOperation/CancunMxpOperation.java
amkCha: now it is - and I cast it as a long as required by the `Trace.java`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/MxpCall.java
amkCha: Okay, I refactored a bit and I created a constructor in `MxpCall` that sets the values

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpUtils.java
amkCha: We can get an instance of a GasCalculator, but it takes a few lines to instantiate one (in tests, we do it once in the TestExecEnv) and then we would have to instantiate a frame as well because this specific bit of code is in private access in Besu - I checked and only method available is `memoryExpansionGasCost` with a frame as input

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleScenario/MxpScenario.java
amkCha: Yes, I created a CancunMxpCall class (aka a super mxp call) to extend `MxpCall` and do all computations

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCode.java
amkCha: - Okay, created a yaml file - value to be checked as I'm not sure what alpha, delta etc are 
```

  - mnemonic: MCOPY
    value: 0x5e
    instructionFamily: COPY
    stackSettings:
      pattern: COPY
      alpha: 0
      delta: 3
      staticGas: G_VERY_LOW
      flag1: true
    billing:
      byWord:
        type: TYPE_4
        wordPrice: G_COPY
```
- I think that it should be created for CancunHub here, as long as the fork chain name is `cancun`

```
  public ZkTracer(ChainConfig chain) {
    loadOpcodes(chain.fork);
    this.chain = chain;
    this.hub =
        switch (chain.fork) {
          case LONDON -> new LondonHub(chain);
          case SHANGHAI -> new ShanghaiHub(chain);
          case CANCUN -> new CancunHub(chain);
          case PRAGUE -> new PragueHub(chain);
        };

```

letypequividelespoubelles: Alpha / delta is NB on stack items added / deleted by the opcode

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleScenario/MxpxMxpScenario.java
amkCha: Done !

arithmetization/src/main/resources/cancunOpcodes.yml
```diff
@@ -623,6 +623,18 @@ opcodes:
...
+    value: 0x5e
+    instructionFamily: MCOPY
+    stackSettings:
+      pattern: THREE_ZERO
+      alpha: 0
+      delta: 3
+      staticGas: G_VERY_LOW
+    billing:
+      byWord:
+        type: TYPE_4
```
letypequividelespoubelles, line 636: I thought the different `type` has disappeared from Instruction Decoder / MXP ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunMxpCall.java
letypequividelespoubelles, line 1: I think you should update the tracing (for the HUB) of the MXP call in this class. For the moment, a CancunMxpCall still has a London tracing for the HUB.

letypequividelespoubelles, line 1: I find quite strange that MxpCall and CacunMxpCall are in totally different package ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/MxpCall.java
```diff
@@ -63,6 +128,40 @@ public boolean getSize2NonZeroNoMxpx() {
...
+   */
+  public CancunMxpCall getMxpScenario(Wcp wcp, Euc euc) {
+    OpCode opCode = this.opCodeData.mnemonic();
+    if (opCode == OpCode.MSIZE) {
+      return new CancunMSizeMxpCall(this.hub);
+    }
+    if (this.size1.isZero() && this.size2.isZero()) {
+      return new CancunTrivialMxpCall(this.hub, wcp);
+    }
+    // TODO: remove the computation duplicate
```
lorenzogentile404: Can this TODO be removed?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
```diff
@@ -0,0 +1,103 @@
...
+
+  public static MxpExoCall callToLT(final Wcp wcp, Bytes arg1, Bytes arg2) {
+
+    final Bytes32 arg1B32 = Bytes32.leftPad(arg1);
+    final Bytes32 arg2B32 = Bytes32.leftPad(arg2);
+
+    return MxpExoCall.builder()
+        .wcpFlag(true)
+        .instruction(EVM_INST_LT)
+        .arg1Hi(arg1B32.slice(0, LLARGE))
```
lorenzogentile404: Here we may create methods such as:

```
private Bytes hi(Bytes arg) {
   return arg.slice(0, LLARGE);
}

private Bytes lo(Bytes arg) {
   return arg.slice(LLARGE, LLARGE);
}
```
However, why not working with `EWord`, that has already `hi()` and `lo()`?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpUtils.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public static boolean isBytePricingOpcode(OpCode opCode) {
+    return opCode == OpCode.MLOAD
+        || opCode == OpCode.MSTORE
+        || opCode == OpCode.MSTORE8
+        || opCode == OpCode.REVERT
+        || opCode == OpCode.RETURN
+        || opCode.isCall();
+  }
```
lorenzogentile404, line 56: Here it seems `MSIZE` and the `LOG-type` family are missing.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    return opCode == OpCode.MCOPY || opCode.isCall();
+  }
+
+  public static boolean isWordPricingOpcode(OpCode opCode) {
+    return opCode.isLog()
+        || opCode == OpCode.SHA3
+        || opCode.isCopy()
+        || opCode.isCreate()
+        || opCode == OpCode.MCOPY;
+  }
```
lorenzogentile404, line 45: `LOG-type` family is priced by byte.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpUtils.java
```diff
@@ -0,0 +1,67 @@
...
+  public static boolean isBytePricingOpcode(OpCode opCode) {
+    return opCode == OpCode.MLOAD
+        || opCode == OpCode.MSTORE
+        || opCode == OpCode.MSTORE8
+        || opCode == OpCode.REVERT
+        || opCode == OpCode.RETURN
+        || opCode.isCall();
+  }
+
+  // This is a copy and past from FrontierGasCalculator.java
```
lorenzogentile404, line 120: Good artists copy, ~~great artists steal~~ great programmers, too.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/module/CancunMxp.java
```diff
@@ -0,0 +1,37 @@
...
+import net.consensys.linea.zktracer.module.euc.Euc;
+import net.consensys.linea.zktracer.module.hub.fragment.imc.MxpCall;
+import net.consensys.linea.zktracer.module.mxp.moduleOperation.CancunMxpOperation;
+import net.consensys.linea.zktracer.module.wcp.Wcp;
+
+@Getter
+@Accessors(fluent = true)
+@RequiredArgsConstructor
+public class CancunMxp extends LondonMxp {
+
```
lorenzogentile404, line 28: It seems that both `CancunMxp` and `LondonMxp` may extend `Mxp`, as no inheritance features are used between `Cancun` and `London` here.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/module/CancunMxp.java
```diff
@@ -0,0 +1,37 @@
...
+  private final Wcp wcp;
+  private final Euc euc;
+
+  public void call(MxpCall mxpCall) {
```
lorenzogentile404, line 30: The  `Override` annoation is missing here.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunMxpCall.java
```diff
@@ -0,0 +1,97 @@
...
+  public boolean isTrivialScenario() {
+    return false;
+  }
+
+  public boolean isMxpxScenario() {
+    return false;
+  }
+
+  public boolean isStateUpdtWPricingScenario() {
+    return false;
```
lorenzogentile404, line 113: Here I suggest to use the extended name `isStateUpdateWordPricingScenario()` and similarly for the method below.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunMxpxMxpCall.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public boolean isMxpxScenario() {
+    return true;
+  }
+
+  public void computeMxpxExpression(Wcp wcp) {
+    // Row i + 3
+    exoCalls.add(MxpExoCall.callToLEQ(wcp, this.size1, mxpxThreshold));
+    // Row i + 4
```
lorenzogentile404: Here we may consider to make the calls return the result of the comparison as a boolean, and then assign the returned value to `size1IsSmall` and so on for clarity.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunStateUpdtWPricingMxpCall.java
```diff
@@ -0,0 +1,121 @@
...
+    return true;
+  }
+
+  public void computeStateUpdt(Wcp wcp, Euc euc) {
+    final OpCode opCode = this.opCodeData.mnemonic();
+
+    // we filter the row i + 7 wcp call by double_offset to prevent unnecessary comparisons
+    if (isDoubleOffsetOpcode(opCode)) {
+      // Row i + 7
+      final var max1 = this.offset1.toUnsignedBigInteger().add(this.size1.toUnsignedBigInteger());
```
lorenzogentile404: Computation rows from 7 to 10 are common for both `STATE_UPDATE_WORD_PRICING` and `STATE_UPDATE_BYTE_PRICING`, so it makes sense to move this code in a common parent class. Also, it seems they are not computed at the moment in the `STATE_UPDATE_BYTE_PRICING` case.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleOperation/CancunMxpOperation.java
```diff
@@ -0,0 +1,167 @@
...
+        .pMacroOffset1Hi(cancunMxpCall.getOffset1().hi())
+        .pMacroOffset1Lo(cancunMxpCall.getOffset1().lo())
+        .pMacroSize1Hi(cancunMxpCall.getSize1().hi())
+        .pMacroSize1Lo(cancunMxpCall.getSize1().lo())
+        .pMacroOffset2Hi(cancunMxpCall.getOffset2().hi())
+        .pMacroOffset2Lo(cancunMxpCall.getOffset2().lo())
+        .pMacroSize2Hi(cancunMxpCall.getSize2().hi())
+        .pMacroSize2Lo(cancunMxpCall.getSize2().lo())
+        .pMacroRes(
+            cancunMxpCall.isMSizeScenario() ? cancunMxpCall.getMemorySizeInWords() : 0L) // to do
```
lorenzogentile404: Remaining TODO here.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/tables/instructionDecoder/InstructionDecoder.java
```diff
@@ -72,6 +74,7 @@ private static void traceStackSettings(OpCodeData op, Trace.Instdecoder trace) {
         .flag4(op.stackSettings().flag4());
   }
 
+  // TODO: traceMxpSettings
   private static void traceBillingSettings(OpCodeData op, Trace.Instdecoder trace) {
```
lorenzogentile404: 2 remaining TODOs in this file.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
```diff
@@ -121,6 +125,7 @@ public boolean isInvalid() {
   }
 
   public boolean isMxp() {
+    // TODO: isMxp in yml with MxpFlag
```
lorenzogentile404: Remaining TODO.

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/MemoryExpansionExceptionTest.java
lorenzogentile404: Yes, you are right @amkCha. For `London` it still makes sense.

lorenzogentile404 (commented):
Left a few comments, the most meaningful is the computation for byte and word pricing.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunMxpxMxpCall.java
amkCha: Did it in the other classes/computations as well :)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
amkCha: Good one, changing all the inputs types

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/module/CancunMxp.java
amkCha: For now we keep inheritance to be able to have all the previous features and also be able to override if needed

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/MxpCall.java
amkCha: Found a way to avoid duplicate computation :ok_hand:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunStateUpdtWPricingMxpCall.java
amkCha: Created `CancunStateUpdateMxpCall`abstract class :+1:

arithmetization/src/main/resources/cancunOpcodes.yml
amkCha: you're right, was still in todo. As agreed, will do it in a separate PR :ok_hand:

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunMxpCall.java
amkCha: Thanks for explaining how things are done on the hub part :)

amkCha: Comment addressed :ok_hand:

amkCha: CancunMxpCall and suite are many files so it would make the Hub even bigger

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxpv3/MxpUtils.java
amkCha: this part specifically is private, and it can only be accessed via a `memoryExpansionGasCost` that requires to build a frame for it

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleOperation/CancunMxpOperation.java
amkCha: Okay used bytesToLong method :)

lorenzogentile404 (commented):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunStateUpdateMxpCall.java
```diff
@@ -0,0 +1,104 @@
...
+import net.consensys.linea.zktracer.module.mxp.MxpExoCall;
+import net.consensys.linea.zktracer.module.wcp.Wcp;
+import net.consensys.linea.zktracer.opcode.OpCode;
+import org.apache.tuweni.bytes.Bytes;
+
+public abstract class CancunStateUpdateMxpCall extends CancunNotMSizeNorTrivialMxpCall {
+
+  public CancunStateUpdateMxpCall(Hub hub) {
+    super(hub);
+    computeStateUpdt(hub.wcp(), hub.euc());
```
amkCha, line 36: Here I re-use the wcp and euc from the hub directly

arithmetization/src/main/java/net/consensys/linea/zktracer/Fork.java
```diff
@@ -47,4 +47,8 @@ public static Fork fromString(String fork) {
   public static boolean isPostShanghai(Fork fork) {
     return fork.compareTo(SHANGHAI) >= 0;
   }
+
+  public static boolean isPreCancun(Fork fork) {
```
letypequividelespoubelles: sometimes it's inclusive, sometimes exculisve; Maye rename `isShanghaiAndAfter` ... or a better naming

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/MxpCall.java
```diff
@@ -15,40 +15,82 @@
...
+    this.memorySizeInWords = this.hub.messageFrame().memoryWordSize();
+    // set sizes and offsets
+    EWord[] sizesAndOffsets = getSizesAndOffsets(frame);
+    this.size1 = sizesAndOffsets[0] != null ? sizesAndOffsets[0] : EWord.ZERO;
+    this.offset1 = sizesAndOffsets[1] != null ? sizesAndOffsets[1] : EWord.ZERO;
+    this.size2 = sizesAndOffsets[2] != null ? sizesAndOffsets[2] : EWord.ZERO;
+    this.offset2 = sizesAndOffsets[3] != null ? sizesAndOffsets[3] : EWord.ZERO;
+  }
+
+  public static MxpCall getMxpCallByFork(Fork fork, Hub hub) {
```
letypequividelespoubelles, line 81: you could get the fork directly via hub.fork()

letypequividelespoubelles, line 81: an other solution (and what we do everywhere else) would be to have MxpCall abstract, with both LondonMxpCall and CancunMxpCall extending MxpCall

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/KeccakSection.java
```diff
@@ -42,7 +42,8 @@ public KeccakSection(Hub hub) {
     final ImcFragment imcFragment = ImcFragment.empty(hub);
     this.addStackAndFragments(hub, imcFragment);
 
-    final MxpCall mxpCall = new MxpCall(hub);
+    final MxpCall mxpCall = MxpCall.getMxpCallByFork(hub.fork, hub);
```
letypequividelespoubelles, line 45: if you do MxpCall abstract and use hub.fork to generate the right ForkMxpCall you don't need to modify all the section

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
letypequividelespoubelles, line 1: it would make sense to add the tracing of the exo call in this class, this is what we do elsewhere. But it's up to you :)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/module/CancunMxp.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.experimental.Accessors;
+import net.consensys.linea.zktracer.module.hub.fragment.imc.MxpCall;
+import net.consensys.linea.zktracer.module.mxp.moduleOperation.CancunMxpOperation;
+
+@Getter
+@Accessors(fluent = true)
+@RequiredArgsConstructor
+public class CancunMxp extends LondonMxp {
```
letypequividelespoubelles: why not just extends Mxp as they are completely different ? Do you need some feature from LondonMxp ?

letypequividelespoubelles (approved):
Again, just looked at the structure, not the detailed implem. Left some comments, but green light.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/module/CancunMxp.java
amkCha: yes, sry we had discussed that !

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/MxpCall.java
amkCha: true, but I wanted it to be explicit that we see we pass the fork to determine which MxpCall, vs it being hidden

arithmetization/src/main/java/net/consensys/linea/zktracer/Fork.java
amkCha: renamed, I took the opposite `isPostCancun` like the `isPostShanghai` above

arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/MxpExoCall.java
amkCha: true - I think that tracing in the operation per perspective is pretty neat, so will keep it as is for now

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/KeccakSection.java
amkCha: could create circle dependencies

Issue #1975, opened 2025-04-08
[Arithmetization Testing] Generate Execution proof requests and corresponding trace files


gauravahuja:
- After the test environement has working besu and shoemi nodes create the trace file and execution proof request.
- This will involve gather data from shomei and generating the trace file via
  -  linea_generateConflatedTracesToFileV2
  -  rollup_getZkEVMStateMerkleProofV0
 - For each unit test export proof request json file and trace file.

Issue #1973, opened 2025-04-08
[Arithmetization Testing] Update shomei such that it can run in a Thread


gauravahuja:
Having the ability to run shomei from within a thread in a unit test will allow us to integrate besu and shomei in the arithmetization tests. 

We also need to be able to export the shomei package to maven.

Once we have this setup we should be able to 
- Execute transactions on the besu node
- Update the ZkState on shomei
- Query shomei for ZkStateRootHash and Merkele proof which are needed to generate execution proof request.

Issue #1972, opened 2025-04-08
[Arithmetization Testing] Integrate Besu and Shomei nodes in unit tests


gauravahuja:
- Update the new test environment to start both besu and shomei nodes in their respective threads.
- Ensure that trie logs generated as part of transaction execution are sent to Shomei via the shomei plugin.
- Ensure that shomei responds to RPC rollup_getZkEVMStateMerkleProofV0

PR #1994, opened 2025-05-01
Create shomei node, start besu with shomei plugin, write proof request file

gauravahuja:
Implements https://github.com/Consensys/linea-tracer/issues/1973 https://github.com/Consensys/linea-tracer/issues/1972 https://github.com/Consensys/linea-tracer/issues/1975

Depends on https://github.com/Consensys/shomei/pull/115, https://github.com/hyperledger/besu/pull/8591

In order to build changes in this PR locally 

Run the test `ExampleBesuTest`. The conflated trace files, shomei files and execution proof request  will be generated in `arithmetization/build/besu/traces`

To run other tests with besu and shomei node use the gradle task `besuNodeTests`. 
`./gradlew :arithmetization:besuNodeTests`

In this PR I add the ability to run unit tests with a besu node. 

1. A besu node instance is spun up in a new Thread with the required genesis file and plugin parameters. 
2. A shomei node is also spun up for shomei traces. 
3. All the accounts from the ToyExecutionEnv are used to create the genesis file and the transactions are sent one at a time to the besu node via Eth APIs. 
4. Once all transactions are executed the test invokes the RPC linea_generateConflatedTracesToFileV2 on the besu instance to generate a conflated trace file.
5. This conflated trace file is then validated via the CorsetValidator as usual.
6. After this the test generates shomei traces for the conflation. 
7. The shomei traces and confalted files are bundled together for each test under the directory `arithmetization/build/besu/traces/<test-name>`

A lot of files have been modified however, the important changes are in:

1. testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
2. testing/src/main/java/net/consensys/linea/testing/BesuNodeBuilder.java
3. testing/src/main/java/net/consensys/linea/testing/ShomeiNode.java
4. testing/src/main/java/net/consensys/linea/testing/ToyExecutionEnvironmentV2.java

Rest of the changes are just to provide name of the test to the ToyExecutionEnvironment. The test name is provided by having each test extend a new class `TracerTestBase` which initializes a `TestInfo` object.


testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
```diff
@@ -45,31 +54,54 @@
...
+    String tmpTestName =
+        testInfo
+            .map(
+                info ->
+                    String.format(
+                        "%s-%s-%s-%s",
+                        info.getTestClass().get().getSimpleName(),
+                        info.getTestMethod().get().getName(),
+                        info.getDisplayName(),
+                        randomUUID))
```
OlivierBBB, line 97: I guess the UUID is what is responsible for the string of random numbers in the test result folder names ?

testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
```diff
@@ -45,31 +54,54 @@
...
+            .orElse(randomUUID)
+            .replace(' ', '_')
+            .replace(',', '_');
```
OlivierBBB, line 100: Not that it really maters, but is that duplication deliberate ?

testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
```diff
@@ -176,35 +167,162 @@ public void executeTest() {
...
+      final long startBlockNumber, final long endBlockNumber) throws IOException {
+    return jsonRpcRequest(
+        shomeiNode.getJsonRpcUrl(),
+        "rollup_getZkEVMStateMerkleProofV0",
+        new RollupGetZkEvmStateV0Parameter(
+            String.valueOf(startBlockNumber), String.valueOf(endBlockNumber), "test"),
+        MerkelProofResponse.class);
+  }
+
+  private <R, P> R jsonRpcRequest(
```
OlivierBBB, line 259: This is likely due to my total ignorance around how JSON-RPC are handled in general, but why would we have to implement such a method ourselves ? I have the same question around `createJsonRpcRequest` where you explicitly specify some string.

OlivierBBB (approved):
I tried my best to review your PR @gauravahuja , I doubt my review is very valuable but it looks good to me.

testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
gauravahuja: Yes, the random uuid helps deduplication with specific cases with Tests involving a TestFactory where the test class, method name and display name all are same for multiple tests.

testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
gauravahuja: Spaces and commas in directory names make them hard to read and handle in code.

testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
gauravahuja: `createJsonRpcRequest` creates the json object with the required attributes required by the JSON RPC spec.
`jsonRpcRequest` initiates the json rpc via a http request and parses the response.

I did not look for libraries that perfrom json rpc request and it was easy enough to do it myself.

PR #2006, opened 2025-05-16
feat: refactor to utilise trace interface

### LINKED ISSUE #2005: feat: support trace interfaces
In order to support multiple forks at the trace generation level, we need to have a unified interface which the different trace classes can implement.

DavePearce:
The code now generates an interface "Trace.java" which is implemented by "TraceLondon.java".  In principle, this means we can have different forks implementing this same interface, with the shared commonality being exposed.


DavePearce (commented):
LGTM

amkCha (approved):

PR #2015, opened 2025-05-22
Fix/txndata nb rows types

amkCha:
- `NB_WCP_EUC_ROWS_FRONTIER_ACCESS_LIST` is the offset used to pad the wcp and euc flag until the last row. This offset now depends on the fork as post-shanghai, we now introduce 2 extra line to limit and meter init code. This constant is now made fork dependent and fixes 3500+ unit tests

- `NB_ROWS_TYPE_0` `NB_ROWS_TYPE_1` `NB_ROWS_TYPE_2` constants are fork dependent and were left in the common TxnDataOperation class

- switch case for TraceShanghai in `ZkTracer` class

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/Fork.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/Fork.java
@@ -32,21 +32,9 @@ public enum Fork {
     };
   }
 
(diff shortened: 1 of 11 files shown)


DavePearce (approved):
LGTM

PR #2019, opened 2025-05-27
Feat/init code cost gas projection and by word create fix

amkCha:


lorenzogentile404 (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Create2.java
```diff
@@ -48,6 +48,11 @@ public long memoryExpansion() {
     return gc.memoryExpansionGasCost(frame, initCodeOffset, initCodeLength);
   }
 
+  @Override
+  public long initCode() {
+    return gc.initcodeCost((int) initCodeLength);
+  }
+
```
OlivierBBB, line 55: The name of the method is surprising. Maybe rename it to `initCodeCost` like the `gc` method.

OlivierBBB (approved):
method name change, other than that LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Create2.java
amkCha: Agreed, it's because it only takes part in a bigger method that regroups all costs 
image

To keep it consistent i'll keep it like the others if that's okay with you

PR #2020, opened 2025-05-27
Fix/shanghai mxpx tests

amkCha:
`MAX_CODE_SIZE_EXCEPTION` takes precedence on `MEMORY_EXPANSION_EXCEPTION` so we need to use the offset only to trigger an  mxpx without triggering the former


arithmetization/src/test/java/net/consensys/linea/zktracer/module/mxp/MxpTestUtils.java
```diff
@@ -123,8 +124,12 @@ public void triggerNonTrivialButMxpxOrRoobForOpCode(
       offset1 = getRandomBigIntegerByBytesSize(0, MAX_BYTE_SIZE);
       offset2 = getRandomBigIntegerByBytesSize(0, MAX_BYTE_SIZE);
```
lorenzogentile404, line 127: I suggest to change `size1 = getRandomBigIntegerByBytesSize(0, MAX_BYTE_SIZE);` to something like  `size1 =  opCode.isCreate? 1 : getRandomBigIntegerByBytesSize(0,  MAX_BYTE_SIZE);` and leave the rest as it was.

lorenzogentile404 (changes requested):

OlivierBBB (approved):

arithmetization/src/test/java/net/consensys/linea/zktracer/module/mxp/MxpTestUtils.java
amkCha: Good idea ! I'll make that change :+1:

lorenzogentile404 (approved):

PR #2027, opened 2025-06-02
Fix/shanghai add init cost to upfront cost

amkCha:
- include init cost in upfront gas cost
- fix bytes unsigned issue
- fix replays to use fork config
- import gas calculator from tracer to be dynamic

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/types/TransactionProcessingMetadata.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/types/TransactionProcessingMetadata.java
@@ -66,6 +66,7 @@ public abstract class TransactionProcessingMetadata {
   final BigInteger initialBalance;
 
   final long dataCost;
+  final long initCodeCost;
   final long accessListCost;
 
   /* g in the EYP, defined by g = TG - g0 */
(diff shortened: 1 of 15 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/types/TransactionProcessingMetadata.java
```diff
@@ -153,10 +154,12 @@ public TransactionProcessingMetadata(
     // - the contract creation cost in case of deployment
     // - the baseline gas (gas for access lists and 7702 authorizations) is set to zero, because we
     // only consider the cost of the transaction payload
+    initCodeCost = hub.gasCalculator.initcodeCost(besuTransaction.getPayload().size());
```
letypequividelespoubelles, line 157: You checked it returns 0 in pre shanghai fork ?

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/types/TransactionProcessingMetadata.java
amkCha: Yes :ok_hand: it's checked - as I use the gasCalculator from the hub, it's fork dependent, and Besu sets it to 0 prior to Shanghai

PR #2034, opened 2025-06-03
feat: ZkCounter for limitless prover

### LINKED ISSUE #1942: ZkCounter for a limitless prover
The ZkCounter will be responsible for:
- rejecting >512 bytes arg MODEXP
- rejecting RIP and Blake
- counting L2L1

All other limits will disappear. And we don't need any limits during tracing. 

Open questions: 
@AlexandreBelling : Do we want to accept exceptional RIP / Blake and trivial RIP (ie with empty input) as we don't recquire gnark circuits ?
Do we still need to count the nb of calls for MODEXP or just returning `Integer.MAX_VALUE` if huge args and 0 else is enough ?

letypequividelespoubelles:


amkCha:
After thinking about it, when we use `ConflationAwareOperationTracer` class in the `executeTest`method,, it can allow any type of tracer that extends this. Which might not exactly be what we want. We'd rather be sure that it's either `ZkTracer` or `ZkCounter`. 
A suggestion is to create a class 
```
package net.consensys.linea.zktracer;

public sealed interface ZkTracerAndCounter extends ConflationAwareOperationTracer permits ZkTracer, ZkCounter  {
}
``` 
`ZkTracer` will have to extend it 
```

public final class ZkTracer implements ConflationAwareOperationTracer, ZkTracerAndCounter {
```
and then in `executeFrom`, do a pattern matching, sth like
```
    switch (tracer) {
      case ZkTracer zkTracer -> {
        final ChainConfig chainConfig = zkTracer.getChain();
        ExecutionEnvironment.checkTracer(
                tracer,
                new CorsetValidator(chainConfig),
                Optional.of(log),
                // block number for first block
                blockHeader.getNumber(),
                // block number for last block
                blockHeader.getNumber());
      }
      case ZkCounter  zkCounter -> {}
    }
```

With this we would be sure the tracer is either ZkCounter or ZkTracer. 

Another possibility is to implement 2 methods `accept` and `checkTracer` in ZkCounter with them doing nothing (like you did for the fork classes)

letypequividelespoubelles:
> After thinking about it, when we use `ConflationAwareOperationTracer` class in the `executeTest`method,, it can allow any type of tracer that extends this. Which might not exactly be what we want. We'd rather be sure that it's either `ZkTracer` or `ZkCounter`. A suggestion is to create a class
> 
> ```
> package net.consensys.linea.zktracer;
> 
> public sealed interface ZkTracerAndCounter extends ConflationAwareOperationTracer permits ZkTracer, ZkCounter  {
> }
> ```
> 
> `ZkTracer` will have to extend it
> 
> ```
> 
> public final class ZkTracer implements ConflationAwareOperationTracer, ZkTracerAndCounter {
> ```
> 
> and then in `executeFrom`, do a pattern matching, sth like
> 
> ```
>     switch (tracer) {
>       case ZkTracer zkTracer -> {
>         final ChainConfig chainConfig = zkTracer.getChain();
>         ExecutionEnvironment.checkTracer(
>                 tracer,
>                 new CorsetValidator(chainConfig),
>                 Optional.of(log),
>                 // block number for first block
>                 blockHeader.getNumber(),
>                 // block number for last block
>                 blockHeader.getNumber());
>       }
>       case ZkCounter  zkCounter -> {}
>     }
> ```
> 
> With this we would be sure the tracer is either ZkCounter or ZkTracer.
> 
> Another possibility is to implement 2 methods `accept` and `checkTracer` in ZkCounter with them doing nothing (like you did for the fork classes)

added a check that we use either a ZkTracer or a ZkCounter

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1BlockSize.java
```diff
@@ -0,0 +1,131 @@
...
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.datatypes.Transaction;
+import org.hyperledger.besu.evm.log.Log;
+import org.hyperledger.besu.plugin.data.ProcessableBlockHeader;
+
+@Accessors(fluent = true)
+@Getter
+@RequiredArgsConstructor
+public class L1BlockSize implements Module {
```
FlorianHuc, line 36: Isn't it L2BlockSize? or just BlockSize?

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/KeccakBlocksTests.java
```diff
@@ -104,6 +104,6 @@ void keccakCountTests() {
             + Address.SIZE
             // nbBlock * (TIMESTAMP_BYTESIZE + Hash.SIZE + NB_TX_IN_BLOCK_BYTESIZE)
             + 38,
-        l2Block.lineCount());
+        l1BlockSize.lineCount());
```
FlorianHuc, line 107: are we testing with two blocks in the conflation?

FlorianHuc (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1BlockSize.java
letypequividelespoubelles: The name of the class was L2BlockSize, but the module key was L1BlockSize, and @AlexandreBelling refers it to L1BlockSize so I change the classe's name to L1BlockSize

arithmetization/src/test/java/net/consensys/linea/zktracer/module/limits/KeccakBlocksTests.java
letypequividelespoubelles: no, we don't have test for this yet

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
```diff
@@ -0,0 +1,140 @@
...
+  final EventDetectorModule rip = new EventDetectorModule(RIP) {};
+  final EventDetectorModule blake = new EventDetectorModule(BLAKE) {};
+  final L1BlockSize l1BlockSize;
+  final L2L1Logs l2l1Logs = new L2L1Logs();
+  final List<Module> moduleToCount;
+
+  public ZkCounter(LineaL1L2BridgeSharedConfiguration bridgeConfiguration) {
+    l1BlockSize =
+        new L1BlockSize(l2l1Logs, bridgeConfiguration.contract(), bridgeConfiguration.topic());
+    moduleToCount = List.of(modexp, rip, blake, l1BlockSize, l2l1Logs);
```
ahamlat, line 56: This list will be immutable, is that what was intended ?

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
```diff
@@ -0,0 +1,140 @@
...
+    if (output == null) {
+      return; // no output means exceptional precompile call
+    }
+
+    if (frame.getInputData().isEmpty()) {
+      return; // no input data implies no gnark circuit for RIP and BLAKE, and arg < 512 bytes for
+      // MODEXP, so nothing to detect
+    }
+
+    final Address precompile = frame.getContractAddress();
```
ahamlat, line 95: precompileAddress ?

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
```diff
@@ -0,0 +1,140 @@
...
+      final Bytes callData = frame.getInputData();
+      final MemoryRange memoryRange = new MemoryRange(0, 0, callData.size(), callData);
+      final ModexpMetadata modexpMetadata = new ModexpMetadata(memoryRange);
+      if (modexpMetadata.unprovableModexp()) {
+        modexp.detectEvent();
+      }
+      return;
+    }
+
+    if (precompile.equals(RIPEMD160)) {
```
ahamlat, line 107: Do you mean Address.RIPEMD160 ?

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
```diff
@@ -0,0 +1,140 @@
...
+      rip.detectEvent();
+      return;
+    }
+
+    if (precompile.equals(BLAKE2B_F_COMPRESSION)) {
```
ahamlat, line 112: The same as above, do you mean Address.BLAKE2B_F_COMPRESSION ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1BlockSize.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private final L2L1Logs l2l1Logs;
+  private final Address l2l1Address;
+  private final Bytes l2l1Topic;
+
+  private final short TIMESTAMP_BYTESIZE = 32 / 8;
```
ahamlat, line 42: Why not just put 4 and add a comment `// 32/4`

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1BlockSize.java
```diff
@@ -0,0 +1,131 @@
...
+  private final short NB_TX_IN_BLOCK_BYTESIZE = 16 / 8;
```
ahamlat, line 43: The same as above

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1BlockSize.java
```diff
@@ -0,0 +1,131 @@
...
+  public int lineCount() {
+
+    return sizesRlpEncodedTxs.lineCount()
+
+        // Calculates the data size related to the abi encoding of the list of the
+        // from addresses. The field is a simple array of bytes20.
+        + numberOfTransactions.lineCount() * Address.SIZE
+
+        // Calculates the data size related to the block
+        + nbBlock * (TIMESTAMP_BYTESIZE + Hash.SIZE + NB_TX_IN_BLOCK_BYTESIZE);
```
ahamlat, line 86: I wonder if there're cases where one of the calls to lineCount method should be cached and not always calculate it again, either for `sizesRlpEncodedTxs.lineCount()` or for `numberOfTransactions.lineCount()`

ahamlat (commented):
I added few comments. Will double check with real data during load testing when this is deployed to see if there're possible performance optimizations.

amkCha (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
letypequividelespoubelles: yes, this is it

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
letypequividelespoubelles: yes

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
letypequividelespoubelles: yes, no need to change it at any point

arithmetization/src/main/java/net/consensys/linea/zktracer/module/limits/L1BlockSize.java
letypequividelespoubelles: at most we have an addition of two int to perform only, as we already cached the lineCount for the already commited txs/bundle.

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
ahamlat: Oh I see, there is this import
```
import static org.hyperledger.besu.datatypes.Address.*;
```
This is a bit misleading as sometimes you used Address.PRECOMPILE_NAME and sometimes just PRECOMPILE_NAME. I would recommend remove the static import and always use Address.PRECOMPILE_NAME

ahamlat: Also because you have other static fields like MODEXP as a String, I guess MODEXP is referenced with Address.MODEXP to avoid the conflict even if we have the static import.

PR #2035, opened 2025-06-03
feat: run unit tests per fork

amkCha:
A flow suggestion on how to run unit tests per fork 
- set an env variable `UNIT_REPLAY_TESTS_FORK`
- add a CI Action for Shanghai Unit tests that runs in // of current Unit London and Replay
- set a property `unit.replay.tests.fork` that copies the env variable or default to `LONDON` in gradle
- leverage `TracerTestBase` class setup by Gaurav and from which all tests inherit
- use gradle property set above to add a `chainConfig` to the bare `testInfo` previously available in `TracerTestBase`
- pass `testInfoWithChainConfig` (didn't rename it for now) to the `ToyExecutionEnvironment.build`
- build tracer from the fork available in the chain config

### DIFF
--- a/.github/workflows/gradle-tests.yml
+++ b/.github/workflows/gradle-tests.yml
@@ -74,7 +74,7 @@ jobs:
           CORSET_FLAGS: disable
           GOCORSET_FLAGS: -b1024 -v --ansi-escapes=false --report --air
           UNIT_TESTS_PARALLELISM: 4
-          UNIT_TESTS_FORK: LONDON
+          UNIT_REPLAY_TESTS_FORK: LONDON
 
       - name: Upload test report
         if: ${{ always() }}
@@ -121,7 +121,7 @@ jobs:
           CORSET_FLAGS: disable
           GOCORSET_FLAGS: -b1024 -v --ansi-escapes=false --report --air
           UNIT_TESTS_PARALLELISM: 4
-          UNIT_TESTS_FORK: SHANGHAI
+          UNIT_REPLAY_TESTS_FORK: SHANGHAI
 
       - name: Upload test report
         if: ${{ always() }}
@@ -168,6 +168,7 @@ jobs:
           REPLAY_TESTS_PARALLELISM: 4
           CORSET_FLAGS: disable
           GOCORSET_FLAGS: -b1024 -v --ansi-escapes=false --report --air
+          UNIT_REPLAY_TESTS_FORK: LONDON
 
       - name: Upload test report
         if: ${{ always() }}
--- a/.github/workflows/gradle-tests.yml
+++ b/.github/workflows/gradle-tests.yml
@@ -51,9 +51,13 @@ jobs:
         run: ./gradlew --no-daemon --parallel clean spotlessCheck
 
   # ==================================================================
-  # Unit Tests for London
+  # Unit Tests
   # ==================================================================
-  unit-tests-london:
+  unit-tests:
+    strategy:
+      matrix:
+        name: [london, shanghai]
+        zkevm_fork: [LONDON, SHANGHAI]
     needs: [ build ]
     runs-on: gha-runner-scale-set-ubuntu-22.04-amd64-xxl
     steps:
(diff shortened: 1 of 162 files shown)


testing/src/main/java/net/consensys/linea/reporting/TracerTestBase.java
```diff
@@ -14,14 +14,28 @@
...
 
+import net.consensys.linea.zktracer.ChainConfig;
+import net.consensys.linea.zktracer.Fork;
+import org.hyperledger.besu.datatypes.Address;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.TestInfo;
 
 public class TracerTestBase {
-  public TestInfo testInfo;
+  public static TestInfoWithChainConfig testInfo = new TestInfoWithChainConfig();
```
amkCha, line 24: Can be renamed :ok_hand: , it was easier to keep `testInfo` to change all the tests while we agree on the implementation

testing/src/main/java/net/consensys/linea/testing/BytecodeCompiler.java
```diff
@@ -45,7 +45,7 @@ private BytecodeCompiler() {}
    * @return an instance of {@link BytecodeCompiler}
    */
   public static BytecodeCompiler newProgram() {
-    loadOpcodes(UNIT_TEST_CHAIN.fork);
+    loadOpcodes(Fork.LONDON);
```
amkCha, line 48: To do: make dynamic

arithmetization/src/test/java/net/consensys/linea/zktracer/exceptions/multiExceptions/ReturnTest.java
```diff
@@ -108,10 +109,13 @@ void maxCodeSizeAndOogExceptionForCreate() {
...
     // We run the program with a gas cost that triggers OOGX
     // We calculate all opcodes gas before the RETURN opcode
-    // 32027L = 3L PUSH + 3L PUSH + 6L MSTORE + 3L PUSH + 3L PUSH + 3L PUSH + 32000L CREATE +
-    // ((32027-3-3-6-3-3-3-32000/64))(less than 0.5 so not adding gas) + 3L PUSH + 3L PUSH
+    // 32027L = 3L PUSH + 3L PUSH + 6L MSTORE + 3L PUSH + 3L PUSH + 3L PUSH + 32000L CREATE
+    // + (post-Shanghai) 2L INIT CODE COST
+    // + ((32027-3-3-6-3-3-3-32000/64))(less than 0.5 so not adding gas) + 3L PUSH + 3L PUSH
     // 21000L for the intrinsic transaction cost
-    bytecodeRunner.run(32027L + 21000L, testInfo);
+    var initCodeCost = isPostShanghai(testInfo.chainConfig.fork) ? 2L : 0L;
```
amkCha, line 116: @DavePearce an example of what we talked about earlier where gas is hardcoded and different when running on London and Shanghai - this is a hotfix while we find a better way to account for the difference :+1:

.github/workflows/gradle-tests.yml
```diff
@@ -51,9 +51,9 @@ jobs:
         run: ./gradlew --no-daemon --parallel clean spotlessCheck
 
   # ==================================================================
-  # Unit Tests
+  # Unit Tests for London
   # ==================================================================
-  unit-tests:
+  unit-tests-london:
```
DavePearce: I wonder if its possible write a reusable action here to avoid duplicating everything, especially as some of the settings for memory usage etc, are tricky?

DavePearce (approved):
Overall, looks good to me.  The only thing I really didn't like was the name of the environment variable `UNIT_REPLAY_TESTS_FORK`.  I don't think that's a good choice of name, as it implies its only for use with the unit/replay tests.  I would just use `ZKEVM_FORK` and then we can reuse the same variable across e.g. replay tests, blockchain tests, whatever.

The other comment I have, which is just to give you an idea of what I might have done differently.  You have a lot of things like this:

```
final ToyExecutionEnvironmentV2 executionEnvironmentV2 =
        ToyExecutionEnvironmentV2.builder(testInfo)
```

and e.g.

```
final Bytes bytecode =
        BytecodeCompiler.newProgram(testInfo)
```

Now, given that all the tests extend `TraceTestBase`, I would probably have taken the opportunity to hide some of this configuration.  It doesn't save time for this PR though, but it might in future PRs (then again, it might not).  Roughly, I would have added methods like this to `TracerTestBase`:

```
protected BytecodeCompiler newBytecodeProgram() {
   return BytecodeCompiler.newProgram(testInfo);
}
```

And, something similar for the `ToyExecutionEnvironmentV2`, etc.  Then, we would call these methods in our tests instead of calling e.g. `BytecodeCompiler.newProgram(testInfo)` directly, etc.  It means if we want to make future changes regarding configurations like this, we might only have to change these methods rather than all the places they are called from.  

Anyways, I wouldn't actually make the change at this stage, since you've got the PR ready to go.

DavePearce:
Despite the tests passing, there is a problem related to running tests individually.  For example, running:

```
./gradlew :arithmetization:test --tests "*.RevertingLogsTests" 
```

On my machine, this crashes with an exception.  The reason is to do with accessing `testInfo` as a static variable before it has been initialised.  I can probably suggest a fix, but maybe we should do it after this is merged?

.github/workflows/gradle-tests.yml
amkCha: Good idea, let me try to do that !

amkCha:
> Overall, looks good to me. The only thing I really didn't like was the name of the environment variable `UNIT_REPLAY_TESTS_FORK`. I don't think that's a good choice of name, as it implies its only for use with the unit/replay tests. I would just use `ZKEVM_FORK` and then we can reuse the same variable across e.g. replay tests, blockchain tests, whatever.
> 
> The other comment I have, which is just to give you an idea of what I might have done differently. You have a lot of things like this:
> 
> ```
> final ToyExecutionEnvironmentV2 executionEnvironmentV2 =
>         ToyExecutionEnvironmentV2.builder(testInfo)
> ```
> 
> and e.g.
> 
> ```
> final Bytes bytecode =
>         BytecodeCompiler.newProgram(testInfo)
> ```
> 
> Now, given that all the tests extend `TraceTestBase`, I would probably have taken the opportunity to hide some of this configuration. It doesn't save time for this PR though, but it might in future PRs (then again, it might not). Roughly, I would have added methods like this to `TracerTestBase`:
> 
> ```
> protected BytecodeCompiler newBytecodeProgram() {
>    return BytecodeCompiler.newProgram(testInfo);
> }
> ```
> 
> And, something similar for the `ToyExecutionEnvironmentV2`, etc. Then, we would call these methods in our tests instead of calling e.g. `BytecodeCompiler.newProgram(testInfo)` directly, etc. It means if we want to make future changes regarding configurations like this, we might only have to change these methods rather than all the places they are called from.
> 
> Anyways, I wouldn't actually make the change at this stage, since you've got the PR ready to go.

I renamed the env variable to `ZKEVM_FORK` :ok_hand: 
And I think you are right, we should leverage `TracerTestBase` a bit more but might need more time to understand what should go where. So I'll merge this and we can refactor it afterwards. For instance, the issue you brought up with the test `RevertingLogsTests` not launching locally because `testInfo` is initialized only when tests starts (and it was called outside of a test so it was failing) makes me think that we could actually have a task to store the `chainConfig` in a class and run this task prior to all gradle test tasks - tbd

PR #2040, opened 2025-06-04
Introduce LineCountingTracer

fab-10:
Introduce an interface common to the `ZkTracer` and `ZkCounter` so it simpler to switch from one to another in the Sequencer, and maybe other places.


letypequividelespoubelles (approved):
LGTM

PR #2044, opened 2025-06-05
feat: run weeklys and nightlys on shanghai and london

amkCha:


arithmetization/src/test/java/net/consensys/linea/zktracer/ZkTracerTest.java
```diff
@@ -28,7 +27,7 @@ public class ZkTracerTest extends TracerTestBase {
 
   @Test
   public void createNewTracer() {
-    final ZkTracer zkTracer = new ZkTracer(MAINNET_LONDON_TESTCONFIG);
+    final ZkTracer zkTracer = new ZkTracer(testInfo.chainConfig);
```
amkCha, line 30: I changed tests with tracers so that they run on Shanghai and London with the chainConfig in param

gradle/tests.gradle
```diff
@@ -149,8 +149,8 @@ tasks.test.configure {
   systemProperty("besu.plugins.dir", System.getProperty("besu.plugins.dir"))
 
   def fork = System.getenv().getOrDefault("ZKEVM_FORK", "LONDON").toString()
-  System.setProperty("unit.replay.tests.fork", fork)
-  systemProperty("unit.replay.tests.fork", fork)
+  System.setProperty("zkevm.fork", fork)
```
DavePearce: FYI: I'm not sure that you need to use a property here.  I manage to do it just using the environment variable `ZKEVM_FORK`.

DavePearce (approved):

gradle/tests.gradle
amkCha: Great, then we just need to put it somewhere that you need to change that env variable to run it locally

PR #2048, opened 2025-06-09
fix: weekly tests

### LINKED ISSUE #2047: fix: weekly tests
These are still failing because a custom property `unit.replay.tests.fork` has been hardcoded.  This is not needed, and should be removed.

DavePearce:
These are currently failing because the property `unit.replay.tests.fork` is not set.  In fact, this property is not required at all since we already have an environment variable `ZKEVM_FORK` for this purpose.  Hence, have refactored to use the environment variable directly, rather than the property.  A default is given of `LONDON` ... which may or may not be a sensible choice.

### DIFF
--- a/.github/actions/setup-go-corset/action.yml
+++ b/.github/actions/setup-go-corset/action.yml
@@ -9,4 +9,4 @@ runs:
 
     - name: Install Go Corset
       shell: bash
-      run: go install github.com/consensys/go-corset/cmd/go-corset@13225e50
+      run: go install github.com/consensys/go-corset/cmd/go-corset@v1.1.9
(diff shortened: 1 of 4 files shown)


DavePearce:
There is still a problem in the `HubShomeiTests` which are reporting an error:

```
net.consensys.linea.zktracer.statemanager.HubShomeiTests > uselessPrewarming(OpCode) [1] SSTORE FAILED

  java.lang.NullPointerException: Cannot read field "fork" because "testInfo.chainConfig" is null
```

letypequividelespoubelles (approved):
LGTM

PR #2052, opened 2025-06-10
fix: rejecting more precompile scenarii in ZkCounter

### LINKED ISSUE #2051: fix ZkCounter rejects some inputs

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
```diff
@@ -104,12 +95,16 @@ public void tracePrecompileCall(MessageFrame frame, long gasRequirement, Bytes o
       return;
     }
 
-    if (precompile.equals(RIPEMD160)) {
+    if (precompileAddress.equals(RIPEMD160)) {
+      // We accept empty input data, as it implies no gnark circuit, so nothing to detect.
+      if (frame.getInputData().isEmpty()) {
+        return;
+      }
```
OlivierBBB: That's acceptable, but not what we agreed upon yesterday, whereby the simplest solution would just be to reject transactions containing any RIPEMD call (which don't fail at the CALL opcode).

arithmetization/src/test/java/net/consensys/linea/ZkCounterTest.java
```diff
@@ -269,13 +271,9 @@ void ripBlakeCall(Address prc, boolean exceptional, boolean emptyCds) {
 
     // no precompile call, but a PRC:
     assertEquals(0, lineCountMap.get(MODEXP));
-    final int expectedRIP =
-        (prc.equals(Address.RIPEMD160) && !exceptional && !emptyCds) ? Integer.MAX_VALUE : 0;
+    final int expectedRIP = (prc.equals(RIPEMD160) && !emptyCds) ? Integer.MAX_VALUE : 0;
```
OlivierBBB: idem

OlivierBBB (approved):
We can keep things as they are, but it's not what we discussed yesterday wrt rejecting RIPEMD wholesale.

PR #2054, opened 2025-06-11
test: add delay

jonesho:
Just for testing purpose


linea-constraints
letypequividelespoubelles, line 1: please don't update this

zkevm-constraints
letypequividelespoubelles, line 1: same

PR #2063, opened 2025-06-13
Improve ux of test with besu and shomei nodes

gauravahuja:
In order to run a Replay test with besu and shomei nodes just add a junit tag `@Tag("replay_with_besu") ` on top of the test and run it with the gradle task `besuNodeReplayTests`


fluentcrafter:
@gauravahuja what is the status of this? Shall it be merged or closed?

rajmani098:
Nice

PR #2071, opened 2025-06-16
test: trivial push0

### LINKED ISSUE #2070: PUSH0 testing

letypequividelespoubelles:


amkCha:
Maybe you have it in mind, but as is, the test is going to run for London test suite as well. A solution is to add an @Tag("shanghai") like we do for weeklys and account for that in the tasks

letypequividelespoubelles:
yes, I want it to run for every fork. Because it's still an acceptable bytecode to run in London, but instead of PUSH0, it should be interpreted as 'INVALID' opcode. So we want this test to run for every fork.

We want to be sure that it fulfills the constraints and doesn't throw a java excption in London.

amkCha (approved):

amkCha:
> yes, I want it to run for every fork. Because it's still an acceptable bytecode to run in London, but instead of PUSH0, it should be interpreted as 'INVALID' opcode. So we want this test to run for every fork.
> 
> We want to be sure that it fulfills the constraints and doesn't throw a java excption in London.

ok got it :ok_hand: do you still want it in a shanghai folder then ?

letypequividelespoubelles:
> ok got it 👌 do you still want it in a shanghai folder then ?

we can discuss this, but I did it because it's test coming fro Shanghai EIPs, I did the same for Paris, Cancun ... but of course we could do differently. Open to suggestions

PR #2075, opened 2025-06-16
feat: restict slack notifications to arith-dev

### LINKED ISSUE #2074: feat: trigger slack notification only on `arith-dev`
Currently, failing tests on branches other than `arith-dev` are annoying since they generate warnings even though we might be expecting them to fail.

DavePearce:
This restricts slack failure or cancellation notices to those workflows running on arith-dev.  This means that workflows being run on other people's branches will not result in annoying failure notifications.


letypequividelespoubelles (approved):
LGTM

PR #2083, opened 2025-06-18
Feat: make line counter tracer selection configurable

fab-10:
Allow to choose the `ZkCounter` implementation via the CLI option `--plugin-linea-limitless-enabled=true`, by default is `false`, thus using `ZkTracer`, to allow for a smooth transition.


letypequividelespoubelles (approved):
LGTM

PR #2098, opened 2025-06-23
feat: add recommendations to improve ci matric strategy

amkCha:


.github/workflows/reference-blockchain-tests.yml
```diff
@@ -83,7 +84,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: failedBlockchainReferenceTests-${{ matrix.zkevm_fork }}.json
+          name: failedBlockchainReferenceTests.json
```
DavePearce: Did you mean to set this back to how it was?

DavePearce (approved):
LGTM

.github/workflows/reference-blockchain-tests.yml
amkCha: Nope ! good catch, let me put it back :)

PR #2105, opened 2025-06-24
fix: load opcodes in BytecodeCompiler before each test i/o with every…

amkCha:
… new program

### DIFF
--- a/testing/src/main/java/net/consensys/linea/reporting/TracerTestBase.java
+++ b/testing/src/main/java/net/consensys/linea/reporting/TracerTestBase.java
@@ -14,6 +14,7 @@
  */
 package net.consensys.linea.reporting;
 
+import net.consensys.linea.testing.BytecodeCompiler;
 import net.consensys.linea.zktracer.ChainConfig;
 import net.consensys.linea.zktracer.Fork;
 import org.hyperledger.besu.datatypes.Address;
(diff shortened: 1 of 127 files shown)


testing/src/main/java/net/consensys/linea/testing/BytecodeCompiler.java
```diff
@@ -40,34 +40,34 @@ public class BytecodeCompiler {
...
+    }
+    ;
+  }
+
+  /**
+   * Create a new program instance that will contain a new bytecode sequence.
+   *
+   * @return an instance of {@link BytecodeCompiler}
+   */
+  public static BytecodeCompiler newProgram() {
```
amkCha, line 69: decouple `newProgram` method from loading the opcodes

testing/src/main/java/net/consensys/linea/reporting/TracerTestBase.java
```diff
@@ -39,6 +40,8 @@ public void init(TestInfo testInfo) {
                   "Unknown fork: " + System.getProperty("unit.replay.tests.fork"));
             };
     TracerTestBase.testInfo.testInfo = testInfo;
+    // Load opcodes for the specified fork
+    BytecodeCompiler.initOpcodes(TracerTestBase.testInfo);
```
amkCha, line 44: Load opcodes once per test so it's not loaded again when we launch parameterized tests

testing/src/main/java/net/consensys/linea/testing/BytecodeCompiler.java
DavePearce: Unfortunately, this is still not thread safe --- there's no locking (or similar) going on.

amkCha:
Replace by https://github.com/Consensys/linea-tracer/pull/2107

PR #2107, opened 2025-06-24
fix: ensure loadOpcodes is threadsafe

### LINKED ISSUE #2106: fix: make `loadOpcodes` threadsafe

DavePearce:
This puts in place some structure within the OpCodes class for ensuring that loadOpcodes is threadsafe (when previously it wasn't).  This solution is constrained by the desire to retain the existing design as much as possible.


letypequividelespoubelles (approved):
LGTM, but my multi thread knowledge is quite poor

PR #2108, opened 2025-06-25
feat: get BlobBaseFee from BlockHeader and not the frame

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/plugins/rpc/batchlinecount/ConflatedCountTracesV2.java
```diff
@@ -116,8 +116,6 @@ private LineCountingTracer createLineCountingTracer() {
         : new ZkTracer(
             Fork.LONDON,
             l1L2BridgeSharedConfiguration,
-            BesuServiceProvider.getBesuService(besuContext, BlockchainService.class)
-                .getChainId()
-                .orElseThrow());
+            BesuServiceProvider.getBesuService(besuContext, BlockchainService.class));
```
fab-10, line 119: little opt: the `BlockchainService` can be retrieved and put in an instance field when the class is created, since it is a singleton

arithmetization/src/main/java/net/consensys/linea/zktracer/ZkTracer.java
```diff
@@ -83,16 +84,17 @@ public ZkTracer(
...
     this.hub =
         switch (chain.fork) {
-          case LONDON -> new LondonHub(chain);
-          case PARIS -> new ParisHub(chain);
-          case SHANGHAI -> new ShanghaiHub(chain);
-          case CANCUN -> new CancunHub(chain);
-          case PRAGUE -> new PragueHub(chain);
+          case LONDON -> new LondonHub(chain, null);
+          case PARIS -> new ParisHub(chain, null);
+          case SHANGHAI -> new ShanghaiHub(chain, null);
```
fab-10, line 95: nit: why not always passing it even if not really used? This way you can assume it is never null, and in tests you can use a mock

fab-10 (approved):
LGTM, just a couple a suggestions

PR #2121, opened 2025-07-02
test: not trigger Ci when draft

amkCha:


letypequividelespoubelles (approved):
lokks like it works as your check has been skipped :)

PR #2140, opened 2025-07-16
fix: add tstore check in load store method from stack

amkCha:


amkCha:
Done here https://github.com/Consensys/linea-tracer/pull/2141/files

PR #2145, opened 2025-07-22
Fix/debug cancun unit

amkCha:
1. Fix and re-enable disabled tests that were hanging for Cancun because of the use of types for Mxp Opcodes
2. `setMayTriggerNontrivialMmuOperation` for Cancun even if not traced
3. add `BLOBHASH` to gasProjector
4. Fix self-destruct in Hub when not in same tx 
- SelfdestructSection fork dependent
- AccountFragment `resolveAtEndTransaction` fork dependent
5. Record fork in `TracerTestBase` to avoid calling lengthy `testInfo.chainConfig.fork` in many places
6. Set WordsNew and CMemNew to default values in a no-state-update scenario case in mxp
7. Don't need to cast Trace  to TraceShanghai anymore
8. Fix `needsUpdate` in `TimeAndExistence` for had code initially logic


letypequividelespoubelles (changes requested):
as discusssed for selfDestructSection and AccountFragment, just make a boolean method that returns true for london and hadCodeInitially for cancun

letypequividelespoubelles (approved):

lorenzogentile404 (approved):

PR #2160, opened 2025-07-29
feat: integrate assembly add module

### LINKED ISSUE #2156: feat: support assembly implementation of `add` module
An assembly version of the `add` module is being added for `SHANGHAI` and `CANCUN`, and the goal here is to support that.  The challenge is that we must retain the old implementation for the other two forks (`LONDON` and `PARIS`).  This is required because we are currently maintaining backwards compatibility with older versions of `go-corset`.  Eventually, when this restriction is lifted, we should not need separate implementations.

DavePearce:
This splits the Add module into LondonAdd and ShanghaiAdd, with the latter being responsible for the assembly module.


.github/actions/setup-go-corset/action.yml
```diff
@@ -12,4 +12,4 @@ runs:
 
     - name: Install Go Corset
       shell: bash
-      run: go install github.com/consensys/go-corset/cmd/go-corset@v1.1.14
+      run: go install github.com/consensys/go-corset/cmd/go-corset@1b82df46e4690fc20293d1b1bdbe19c68af754a9
```
letypequividelespoubelles: the 1.1.15 is ok ?

linea-constraints
letypequividelespoubelles, line 1: please merge first the constraint into master and then update to latest master

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -203,7 +203,7 @@ public int spillage(Trace trace) {
   // stateless modules
   @Getter private final Wcp wcp = new Wcp();
 
-  private final Add add = new Add();
+  private final Add add = setAdd();
```
letypequividelespoubelles, line 206: can't wait to kill the LondonAdd :)

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
DavePearce: Yes, indeed.

PR #2177, opened 2025-08-11
test: storage consistency for system transactions

### LINKED ISSUE #2157: test system transaction account/storage consistency
have multi block conflation where we call system addresses

letypequividelespoubelles:


testing/src/main/java/net/consensys/linea/testing/ToyExecutionTools.java
```diff
@@ -104,6 +112,12 @@ public static void executeTest(
 
     tracer.traceStartConflation(1);
     tracer.traceStartBlock(worldStateUpdater, blockHeader, blockBody, blockHeader.getCoinbase());
+    if (isPostCancun(fork)) {
+      final BlockProcessingContext context =
+          new BlockProcessingContext(
+              blockHeader, initialWorldState, protocolSpec, (BlockHashLookup) null, tracer);
+      protocolSpec.getPreExecutionProcessor().process(context);
```
amkCha, line 119: QQ for me : What's the pre-exec processor used for ? We had a bug not using it for Cancun yet then ?

amkCha (approved):

testing/src/main/java/net/consensys/linea/testing/ToyExecutionTools.java
letypequividelespoubelles: yep, if you don't have this, yoh don't execute system transaction that occurs at the begining of the block (and thus besu doesn't update the storage of the system contracts, but the tracer was, so creating constraints issues)

Issue #2159, opened 2025-07-29
Update `SELFDESTRUCT` for Cancun


OlivierBBB:
For reference and potential readers: `accountFragmentWiping` returns `true` in London. Cancun will likely require more thought.

_Originally posted by @OlivierBBB in https://github.com/Consensys/linea-tracer/pull/2158#discussion_r2241074382_
            

## Note

In Cancun `SELFDESTRUCT` may either lead to account wiping (if the account had empty code prior to the transaction) or not (if it had nonempty code prior to the transaction.) The first case should be correctly handled in the current implementation. The second case, which applies starting with Cancun, is not.

One must look at the spec for the correct behaviour. From memory, one only has to add the final context row in the `!accountFragmentWiping()` case, and likely set the scenario
```java
if (accountFragmentWiping()) {
      // ...
} else {
      // Cancun only
      selfdestructScenarioFragment.setScenario( /* TBD */ );
      this.addFragment(finalUnexceptionalContextFragment);
}
```

OlivierBBB:
# What to implement

In the above, for Cancun, we have
```
accountFragmentWiping() <=> account_had_no_code_initially
```
The main point is that the spec can be re-ordered as follows (assuming `CN_WILL_REVERT = false`, so that the scenario is either `scn/SD/WONT_REVERT_NOT_YET_MARKED` or `scn/SD/WONT_REVERT_ALREADY_MARKED`)
- if `account_had_no_code_initially`:
  - future account wiping (`scn/SD/WONT_REVERT_NOT_YET_MARKED`) takes place iff `trigger_future_account_deletion`
- if `!account_had_no_code_initially` (which corresponds to the `else` case):
  - no account wiping is possible, so we impose `scn/SD/WONT_REVERT_ALREADY_MARKED`

# Spec exerpt

<img width="1145" height="727" alt="Image" src="https://github.com/user-attachments/assets/410435de-a8f4-450f-8fcf-aed0ab0aad0f" />

OlivierBBB:
Done with https://github.com/Consensys/linea-tracer/pull/2184

PR #2184, opened 2025-08-18
fix: set scenario for self dest section for non wiping

amkCha:


OlivierBBB (approved):
The scenario `SELFDESTRUCT_WONT_REVERT_ALREADY_MARKED` is the right one, see https://github.com/Consensys/linea-tracer/issues/2159 comment. LGTM

PR #2189, opened 2025-08-20
fix: InstructionByteStringPrefix compareTo method error when comparing equality

amkCha:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpUtils/InstructionByteStringPrefix.java
```diff
@@ -214,6 +214,9 @@ protected short compareTo(RlpUtilsCall other) {
     if (firstByteComparison != 0) {
       return (short) firstByteComparison;
     }
+    if (isList == ((InstructionByteStringPrefix) other).isList) {
+      return 0; // both are lists or non-lists, so they are equal
```
letypequividelespoubelles, line 218: but here this case should never happen ... 

Can you point to me which test was failing ? If all arguments are the same, then it's a dupmicate operation, and we should trace only one.

letypequividelespoubelles (changes requested):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpUtils/InstructionByteStringPrefix.java
letypequividelespoubelles: the issue is that I forget to add the `@EqualsAndHashCode` annotation

amkCha: thank you for the review ! Here are the tests failing 
- https://github.com/Consensys/linea-tracer/blob/43ed3e0af998e6732c849604b842eeaff992ae23/arithmetization/src/test/java/net/consensys/linea/zktracer/module/blockhash/BlockhashTest.java#L166
- https://github.com/Consensys/linea-tracer/blob/43ed3e0af998e6732c849604b842eeaff992ae23/arithmetization/src/test/java/net/consensys/linea/zktracer/module/stp/StpTest.java#L61
- https://github.com/Consensys/linea-tracer/blob/43ed3e0af998e6732c849604b842eeaff992ae23/arithmetization/src/test/java/net/consensys/linea/zktracer/module/stp/StpTest.java#L90

letypequividelespoubelles:
done in #2206

PR #2196, opened 2025-08-21
fix: ensure loadOpcodes() called by BlockCapturer

### LINKED ISSUE #2195: fix: initialise opcodes for `BlockCapturer`
There is a bug whereby `OpCodes.loadOpcodes()` is not being called before the `BlockCapturer` is run.  This still results in a replay test, but one which is severely corrupted.  An exception is logged, but this is hard to see.

DavePearce:
This ensures that the BlockCapturer calls `OpCodes.loadOpcodes()` with the fork before it starts capturing.  Previously, this did not occur and it resulted in exceptions arising during tracing.

At this stage, there remains an outstanding issue that the fork is currently fixed on LONDON, and is not determined from BESU.  That is a separate issue to be fix.

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
+++ b/arithmetization/src/main/java/net/consensys/linea/blockcapture/BlockCapturer.java
@@ -21,7 +21,9 @@ import java.util.Set;
 import com.google.gson.Gson;
 import net.consensys.linea.blockcapture.reapers.Reaper;
 import net.consensys.linea.zktracer.ConflationAwareOperationTracer;
+import net.consensys.linea.zktracer.Fork;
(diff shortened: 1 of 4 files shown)


DavePearce:
NOTES: there remains an outstanding (though pre-existing) issue with the `Fork` calculation.  Specifically, in `AbstractLineaSharedOptionsPlugin` we have the following:

```java
  public Fork fork() {
    return LONDON; // TODO: IMPORTANT
  }
```

This already existed as is being used for the `GenerateConflatedTracesV2` plugin as well.

FlorianHuc (approved):
This will indeed fix the non initialised opcode list to capture json for replay, but we may still want to improve the solution to cover the case of fork transition.

PR #2204, opened 2025-08-27
feat: refactor `OpCodes` API

### LINKED ISSUE #2201: Dynamic switch between forks
In **`OpCodes.java`** class, we've identified that we can use a map to load the opcodes list for all forks 

As a base for discussion, this branch https://github.com/Consensys/linea-tracer/compare/arith-dev...2201-dynamic-fork-switching-for-opcodes

The 5 methods in this class would then have to have the `fork` as input to get the correct `opCodeDataList` 

- `isValid`
- `iterator`
are easily refactored, as shown in the branch

However for the `of` methods, it requires more thoughts.

They are mainly used in **`OpCode.java`** class in the following methods
- `getData`
- `of`

These methods are called in many modules, at different levels - after checking, it looks feasible to bring the fork information in all places with a big refactoring

At least 2 possible solutions would be
1 - make `OpCode` fork dependent
2 - feed the 2 methods above with fork input

For solution 1, here are the pros and cons that I see
Cons
 - `OpCode` is an enum, not as easily extendable as a class is
Pros
- it can be cleaner to not have an enum that encompasses all opcodes from all forks

For solution 2, 
Cons
- we keep an enum that mixes opcodes form all forks
Pros
- as the methods are called in many modules, it is lighter to feed the methods with the fork than to make a switch on the fork to call the correct fork instance of ̀`OpCode`

In terms of impact/refactoring, I think both are equivalent. Solution 1 seems cleaner but more verbose, solution 2 keeps status-quo on Opcode class

DavePearce:
Previously, the OpCodes API was implemented via a number of static methods and, once a given fork was loaded, it was not possible to load another fork (doing so would cause a runtime exception).

Therefore, this refactors the OpCodes API to remove entirely the static API and, instead, replace it with an API based around OpCodes *instances*.  That is, we have an instance of OpCodes for each fork under consideration.  This means that, in situations where previously we could look up OpCodeData information statically, we must now be able to access a relevant instance of OpCodes.  As such, quite a lot of refactoring was required to make this possible.

The interface now promoters OpCodeData over OpCode as the primary source of useful information.  The OpCode class is now largely relegated to being simply a "mnemonic" with essentially no actual functionality.  In most cases, this works out well because OpCodeData already had a mnemonic accessor, meaning whereever we have an instance of OpCodeData we also have an instance of OpCode.

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
@@ -112,6 +112,7 @@ import net.consensys.linea.zktracer.module.txndata.module.TxnData;
 import net.consensys.linea.zktracer.module.wcp.Wcp;
 import net.consensys.linea.zktracer.opcode.OpCode;
 import net.consensys.linea.zktracer.opcode.OpCodeData;
+import net.consensys.linea.zktracer.opcode.OpCodes;
 import net.consensys.linea.zktracer.opcode.gas.projector.GasProjector;
 import net.consensys.linea.zktracer.runtime.callstack.CallFrame;
 import net.consensys.linea.zktracer.runtime.callstack.CallFrameType;
@@ -139,8 +140,10 @@ import org.hyperledger.besu.plugin.data.ProcessableBlockHeader;
 @Slf4j
 @Accessors(fluent = true)
 public abstract class Hub implements Module {
-
+  /** Active fork for this hub. */
   public final Fork fork;
+
+  private final OpCodes opcodes;
 
   /** The {@link GasCalculator} used in this version of the arithmetization */
   public final GasCalculator gasCalculator = setGasCalculator();
@@ -371,6 +374,7 @@ public abstract class Hub implements Module {
 
   public Hub(final ChainConfig chain) {
     fork = chain.fork;
+    opcodes = OpCodes.load(fork);
     checkState(chain.id.signum() >= 0);
     Address l2l1ContractAddress = chain.bridgeConfiguration.contract();
     final Bytes l2l1Topic = chain.bridgeConfiguration.topic();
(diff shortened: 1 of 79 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
```diff
@@ -1101,4 +1120,9 @@ protected abstract void traceSystemInitialTransaction(
   protected abstract void traceSystemFinalTransaction();
 
   protected abstract void setSelfdestructSection(Hub hub, final MessageFrame frame);
+
+  /** Provides access to fork-specific opcode information. */
+  public OpCodes getOpcodes() {
```
letypequividelespoubelles: better uing a @Getter annotation no ?

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodes.java
```diff
@@ -40,74 +40,46 @@
...
+   */
+  public static OpCodes load(Fork fork) {
+    OpCodes instance = opCodesMap.get(fork);
+    //
+    if (instance == null) {
+      instance = new OpCodes(fork);
+      opCodesMap.put(fork, instance);
     }
+    // Done
+    return instance;
```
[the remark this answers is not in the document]
DavePearce: That behaviour is expected and is preferable to requiring an explicit lock.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
DavePearce: Well, if you like Lombok I suppose so ;)

DavePearce: I'll make this change, no probs

letypequividelespoubelles (approved):

PR #2206, opened 2025-08-27
fix & perf: add equal and hash annotation for rlpUtils calls

### LINKED ISSUE #2205: missing equalAndHash for RlpUtils

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpUtils/InstructionByteStringPrefix.java
```diff
@@ -30,19 +31,20 @@
...
-  @Getter private final boolean isList;
+  @EqualsAndHashCode.Include @Getter private final Bytes32 byteStringLength;
+  @EqualsAndHashCode.Include @Getter private final byte firstByte;
+  @EqualsAndHashCode.Include @Getter private final boolean isList;
 
   // outputs
   @Getter private boolean rlpPrefixRequired;
   private boolean byteStringIsNonEmpty;
-  @Getter private Bytes16 rlpPrefix;
+  @Getter private Bytes16 rlpPrefix = Bytes16.ZERO; // default value if no RLP prefix
```
amkCha: Why do we need a default value now ?

amkCha:
Very clear and clean ! Was a bit more complex of a fix :) Thanks for re-enabling the tests

amkCha (approved):

arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlpUtils/InstructionByteStringPrefix.java
letypequividelespoubelles: because we don't, it was a tmp fix that is no longer required, thanks :)

PR #2217, opened 2025-08-29
feat: update sysi transaction

### LINKED ISSUE #2216: update sysi transaction

### LINKED ISSUE #2222: `SYSI` transaction tests
We should test the following in multi-block tests (2 or 3 typically). Some variables
- first block (`n`) is genesis **(yes/no)**
- the EIP-smc's exists before processing prior to block `n` processing **(yes/no)**

In the case of nonexistence
- some transaction of block `n` interacts with these smc's, regardless of their existence
- deployer smc's are (a) pre-funded or (b) get funded in txn of block `n` **(a / b)**
- block `n + 1` sees the execution of one or both synthetic deployment transactions **(4788 / 2935 / both)**
- some transactions of block `n + 1` interact with these smc's, before / after the EIP-smc deployment transaction, **(before / after / both)**
- block `n + 2` has transactions interacting with the EIP-smc's

letypequividelespoubelles:


testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
OlivierBBB, line 1: So this is adding a new field to the `ReplayExecutionEnvironment` which we can tweak in its builder, right ?

arithmetization/src/test/java/net/consensys/linea/zktracer/forkSpecific/systemTransaction/SystemTransactionTests.java
```diff
@@ -0,0 +1,141 @@
...
+            .build();
+
+    final List<Transaction> genesisBlockTransactions = new ArrayList<>();
+    final List<Transaction> blockNb1Transactions = new ArrayList<>();
+    final List<Transaction> blockNb2Transactions = new ArrayList<>();
+
+    final MultiBlockExecutionEnvironment.MultiBlockExecutionEnvironmentBuilder builder =
+        MultiBlockExecutionEnvironment.builder(testInfo)
+            .startingBlockNumber(0)
+            .systemContractDeployedPriorToConflation(systemContractDeployedBeforeBlockNumber == 0);
```
OlivierBBB: I'm not understanding: the loop allows values 0, 1, 2 but the above treats cases 1 and 2 the same. What is the purpose of the 3 cases ?

testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
letypequividelespoubelles: exactly.

arithmetization/src/test/java/net/consensys/linea/zktracer/forkSpecific/systemTransaction/SystemTransactionTests.java
```diff
@@ -0,0 +1,141 @@
...
+        ToyAccount.builder()
+            .address(Address.wrap(leftPadTo(Bytes.fromHexString("0x4788"), Address.SIZE)))
+            .code(
+                    byteCodeCallingBeaconRootSystemAccountFromCallData(
+                    testInfo, BEACONROOT_ADDRESS))
+            .build();
+
+    final List<Transaction> genesisBlockTransactions = new ArrayList<>();
+    final List<Transaction> blockNb1Transactions = new ArrayList<>();
+    final List<Transaction> blockNb2Transactions = new ArrayList<>();
```
OlivierBBB, line 251: The above suggests that the blocks will only contain system transactions (which may or may not be trivial depending on whether `systemContractDeployedBeforeBlockNumber == 0`). AFAIU this test will test 3 empty blocks in succession, with in the first test the system contracts already being live.

And so far we don't have tests that actually involve the deployer accounts for either synthetic deployment transactions.

OlivierBBB (approved):
LGTM. AFAICT we need multi-block tests (e.g. 3 blocks) à la
- block n: sys contracts not deployed, some tx interacts with one or both (yes/no)
- block n + 1: one or both gets deployed using the synthetic transactions, some tx interacts with one or both (yes/no) before / after deployment
- block n + 2: one or both sys contracts exist, , some tx interacts with one or both (yes/no)
- block n + 4: same

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/account/CancunAccountFragment.java
```diff
@@ -71,6 +71,6 @@ void traceHadCodeInitially(Trace.Hub trace) {
     trace.pAccountHadCodeInitially(
         isUserTransaction(txType)
             ? tx.hadCodeInitiallyMap().get(oldState().address()).hadCode()
-            : true);
+            : oldState().tracedHasCode());
```
letypequividelespoubelles, line 74: real fix, now we can't assume systema transaction are already deployed

testing/src/main/java/net/consensys/linea/testing/ReplayExecutionEnvironment.java
```diff
@@ -270,7 +283,8 @@ private static void executeFrom(
           useCoinbaseAddressFromBlockHeader
               ? header.getCoinbase()
               : CliqueHelpers.getProposerOfBlock(header);
-      tracer.traceStartBlock(world, header, body, miningBeneficiary);
+      tracer.traceStartBlock(world.updater(), header, body, miningBeneficiary);
```
letypequividelespoubelles, line 286: breaks all acc / sto consistancies if you pass the world and not the worldUpdater, took me so long to figure out what was happening

DavePearce (approved):
LGTM

There were some bugbot messages which might be worth looking at I think.

letypequividelespoubelles:
> There were some bugbot messages which might be worth looking at I think.

which one ? I thought I've looked at all of them, am I missing any ?

PR #2224, opened 2025-09-01
feat: allow and test empty blocks post cancun

### LINKED ISSUE #2149: Tests for empty blocks
- multi block tests with some empty blocks

letypequividelespoubelles:


arithmetization/src/test/java/net/consensys/linea/zktracer/EmptyBlockTests.java
```diff
@@ -0,0 +1,189 @@
...
+    if (isPostCancun(testInfo.chainConfig.fork)) {
+
+      final MultiBlockExecutionEnvironment.MultiBlockExecutionEnvironmentBuilder builder =
+          MultiBlockExecutionEnvironment.builder(testInfo);
+
+      // One empty block
+      builder.addBlock(List.of());
+
+      builder.build().run();
+    }
```
lorenzogentile404, line 197: is there some `checkArgument` call we can make it here as well to ensure the behaviour is as expected?

arithmetization/src/test/java/net/consensys/linea/zktracer/EmptyBlockTests.java
```diff
@@ -0,0 +1,189 @@
...
+      short nbOfLog = 0;
+      for (State.HubTransactionState state : hub.getState().getAll()) {
+        for (TraceSection section : state.traceSections().trace()) {
+          if (section instanceof LogSection) {
+            nbOfLog += 1;
+          }
+        }
+      }
+      checkArgument(nbOfLog == 1, "There should be exactly one log section");
+    }
```
lorenzogentile404, line 178: could you add a bit more documentation about the logic of the test? It is clear that there is a mix of empty and non-empty blocks, but could you add something about the purpose of those specific contracts?

lorenzogentile404 (changes requested):

lorenzogentile404 (approved):

PR #2234, opened 2025-09-05
feat: refactor GAS / ADD tracers for asm implementations

### LINKED ISSUE #2232: feat: refactor `GAS` tracer for assembly module
The purpose of this PR is to refactor the tracer for the `GAS` module so it correctly supports the assembly implementation of this module.  In addition, this will simplify the existing `ADD` module implementation by removing the london-specific tracer.

DavePearce:
This refactors both of these modules to account for their corresponding zkasm implementations.  For GAS, we have an entirely new implemetnation replacing the original Corset constraints.  For ADD, we are dropping legacy support in the LONDON fork for the non-assembly implementation.


arithmetization/src/main/java/net/consensys/linea/zktracer/module/add/AddOperation.java
```diff
@@ -36,7 +40,27 @@ public AddOperation(OpCode opCode, Bytes32 arg1, Bytes32 arg2) {
     this.arg2 = arg2;
   }
 
-  public abstract void trace(int stamp, Trace.Add trace);
+  public void trace(int stamp, Trace.Add trace) {
+    UInt256 res;
+    // Compute result of this operation.
+    if (opCode == ADD) {
```
letypequividelespoubelles: can we have a switch ADD, SUB, default -> exception instead ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/Hub.java
letypequividelespoubelles, line 1: The Gas module doesn't require the WCP as argument, you can rm it

letypequividelespoubelles: same for gasOperation

letypequividelespoubelles (approved):
LGTM

PR #2251, opened 2025-09-08
fix: mxpx for cancun and mcopy debugging

### LINKED ISSUE #2238: cancun reference tests debuging - mcopy
failing constraint: failing constraint 

failing constraint hub.mcopy-instruction---setting-the-gas-cost#0

failing constraint hub.execution-environment-consistency---linking#2:

one of the many failing tests for similar reason:

1) blockchainreferencetest_7:  MCOPY_memory_expansion_cost ...

2) NPE for refTest 558, ex: Cancun-full_memory_clean

letypequividelespoubelles:


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
```diff
@@ -187,22 +187,29 @@ public void setTracedException(TracedException tracedException) {
...
+      boolean stackException,
+      boolean isExec) {
     if (!isExec || stackException) {
       return 0;
     }
 
+    // General Case:
     if (!opCode.isNonTrivialPush() && !opCode.isJump()) return pc + 1;
 
+    // PUSHX (with X !=0 ) case
```
OlivierBBB, line 204: This works for all `PUSH`es.

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
```diff
@@ -64,7 +64,7 @@
 @Slf4j
 public class BlockchainReferenceTestTools {
   // Keep the forkName and the zkevm_fork in github worklow in PascalCase
-  private static final String forkName = getForkOrDefault("London");
+  private static final String forkName = getForkOrDefault("Cancun");
```
amkCha: Might want to put it back to London :)

amkCha:
Double catch : MCopy tests + gas projector, nice one !

reference-tests/src/test/java/net/consensys/linea/BlockchainReferenceTestTools.java
letypequividelespoubelles: ... :(

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
```diff
@@ -126,13 +126,12 @@ private static boolean isStackOverflow(final MessageFrame frame, OpCodeData opCo
         > 1024;
   }
 
-  private static boolean isMemoryExpansionFault(
-      MessageFrame frame, OpCodeData opCode, GasProjector gp) {
-    return gp.of(frame, opCode).largestOffset() > 0xffffffffL;
+  private static boolean isMemoryExpansionFault(GasProjection op) {
+    return op.largestOffset() > 0xffffffffL;
```
OlivierBBB: We should be able to replace `0xffffffffL` with `MXPX_THRESHOLD` from the constraints (and `>` with `≥`)

OlivierBBB: Image

OlivierBBB: Also I'm not sure the above implements the constraints faithfully, _at least judging by the name_. We don't look at the largest offset but, cf screenshot, at
- is (one of) the size(s) large ?
- ~if not,~ is (one of) the size(s) nonzero and the corresponding offset large ?

Then and only then should we trigger `MXPX`.

OlivierBBB: Judging by the name (`op.largestOffset()`) it seems the implementation looks at something à la
```java
(size != 0)
    ? offset + (size - 1)
    : 0;
```

OlivierBBB: This also suggests new tests for the `MXP` module
```rust
size_k   = 0, 1, MXPX_THRESHOLD, MXPX_THRESHOLD + 1, randomHuge, MAX_UINT256 // or so
offset_k = <same list>
```
to make sure we precisely trigger the `MXPX` case of the `MXP` module.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/GasProjection.java
```diff
@@ -125,8 +125,4 @@ public final long gasCostExcludingDeploymentCost() {
         + sStoreValue()
         + initCode();
   }
-
-  public final long childGasAllowance() {
-    return gasPaidOutOfPocket() + stipend();
-  }
```
OlivierBBB, line 131: What replaces this one ? Was it never used ?

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/DataCopy.java
```diff
@@ -57,6 +57,10 @@ public long linearPerWord() {
 
   @Override
   public long largestOffset() {
-    return size == 0 ? 0 : Words.clampedAdd(targetOffset, size);
+    return size == 0 ? 0 : Words.clampedAdd(offset(), size);
```
OlivierBBB: This is no longer the right quantity to observe for `MXPX` and to use in `isMemoryExpansionFault`, see previous comment here:

https://github.com/Consensys/linea-tracer/pull/2251/files#diff-ae26c945b2ccfec372622ac5c44b603cefac11c4ccd70ffc38a037ed200a76beR130

OlivierBBB (changes requested):
I believe there are some MXP spec changes that aren't correctly reflected in the current implementation of this module. Also it suggests writing more tests.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/DataCopy.java
OlivierBBB: Also, I would need to confirm, but I believe the London `largestOffset()` should return
```java
return size == 0 ? 0 : Words.clampedAdd(offset(), size - 1);
```

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/common/CommonFragmentValues.java
letypequividelespoubelles: yes, but the case PUSH0 is already dealt with in the first statement

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/GasProjection.java
letypequividelespoubelles: never used yes

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/DataCopy.java
OlivierBBB: If confirmed (which would honestly shock me) this would suggest another test for MXP (London) (maybe @lorenzogentile404 ?) which is to test MXP operations, e.g. for CODECOPY and others, where we set
```rust
DESTINATION_OFFSET = 13 + (<old mxpx threshold> - 32) // more or less, basically 32 away from the threshold + 13
SIZE               = 16, 17, 18, 19, 20, 21 // however many to make sure we cover the <, =, > cases of the threshold
```
The point being: the current implementation of `largestOffset` seems off, though maybe since we compare it via `> 0xff...` the extraneous +1 is compensated for.

OlivierBBB: image

OlivierBBB: The above extract from the old `MXP` spec confirms that in the max offsets are
```
offset_k + (size_k - 1)
```

OlivierBBB: Also the London `MXP` spec specifies essentially with

```rust
LONDON_MXPX_THRESHOLD := 256**4

MXPX = <true>  <=>  max_offset ≥ LONDON_MXPX_THRESHOLD
               <=>   [ max_offset_1 ≥ LONDON_MXPX_THRESHOLD ]
                   ∨ [ max_offset_2 ≥ LONDON_MXPX_THRESHOLD ]
```

image

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
amkCha: Don't know if it helps, but the spec bit copied above where mxpx expression is computed is written in the mxp module here https://github.com/Consensys/linea-tracer/blob/arith-dev/arithmetization/src/main/java/net/consensys/linea/zktracer/module/mxp/moduleCall/CancunMxpCall.java#L166 - the `isMemoryExpansionFault` seems to be the method that triggers a `MEMORY_EXPANSION_EXCEPTION`, not the method that compute the MXPX boolean. But maybe I don't understand and they should be the same

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
```diff
@@ -60,14 +60,28 @@ public long memoryExpansion() {
...
     }
 
-    return Math.max(
-        inputData.isEmpty() ? 0 : Words.clampedAdd(inputData.offset(), inputData.size()),
-        returnData.isEmpty() ? 0 : Words.clampedAdd(returnData.offset(), returnData.size()));
+    switch (fork) {
+      case LONDON, PARIS, SHANGHAI -> {
+        return Math.max(
+            inputData.isEmpty() ? 0 : Words.clampedAdd(inputData.offset(), inputData.size() - 1),
+            returnData.isEmpty()
```
OlivierBBB: General comment: `returnData` is a misnomer in this class (and slightly less so: `inputData`). It's what the we call the `returnAtRange` (and `callDataRange`) elsewhere:

https://github.com/Consensys/linea-tracer/blob/0d8e5e6581a9a17ab53d5e4c8e213ca8a769a727/arithmetization/src/main/java/net/consensys/linea/zktracer/runtime/callstack/CallFrame.java#L84-L86

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
```diff
@@ -126,8 +127,12 @@ private static boolean isStackOverflow(final MessageFrame frame, OpCodeData opCo
         > 1024;
   }
 
-  private static boolean isMemoryExpansionFault(GasProjection op) {
-    return op.largestOffset() > 0xffffffffL;
+  private static boolean isMemoryExpansionFault(Fork fork, GasProjection op) {
+    return switch (fork) {
+      case LONDON, PARIS, SHANGHAI -> op.mxpxOffset(fork) >= MXPX_THRESHOLD;
+      case CANCUN, PRAGUE, OSAKA -> op.mxpxOffset(fork) > MXPX_THRESHOLD;
```
OlivierBBB: Is MXP_THRESHOLD fork dependent in the above ? It seems fishy, as I believe both use the same criterion (which is `≥ 256 ** 4`) except they detect it differently:

```rust
// London etc:
mxpx = <true>  <=> mxpxOffset ≥ LONDON_MXPX_THRESHOLD = 256 ** 4

// Cancun and later:
mxpx = <false> <=> mxpxOffset ≤ CANCUN_MXPX_THRESHOLD = 256 ** 4 - 1
```

which ultimately boils down to the same criterion:
```rust
mxpx = <true>  <=> mxpxOffset ≥ 256 ** 4
```

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
```diff
@@ -60,14 +60,28 @@ public long memoryExpansion() {
...
+                ? 0
+                : Words.clampedAdd(returnData.offset(), returnData.size() - 1));
+      }
+      case CANCUN, PRAGUE, OSAKA -> {
+        return (inputData.isEmpty() && returnData.isEmpty())
+            ? 0
+            : Math.max(
+                Math.max(inputData.offset(), inputData.size()),
+                Math.max(returnData.offset(), returnData.size()));
```
OlivierBBB: The above is technically correct given that the `Range` constructor returns `Range.EMPTY` if the size is 0. A comment to that effect would dispel future quiproquos / misunderstandings.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/DataCopy.java
```diff
@@ -56,11 +57,11 @@ public long linearPerWord() {
...
-  }
-
-  protected long offset() {
-    return targetOffset;
+  public long mxpxOffset(Fork fork) {
+    return switch (fork) {
+      case LONDON, PARIS, SHANGHAI -> size == 0 ? 0 : clampedAdd(targetOffset, size - 1);
+      case CANCUN, PRAGUE, OSAKA -> size == 0 ? 0 : Math.max(targetOffset, size);
+      default -> throw new IllegalArgumentException("Unknown fork: " + fork);
+    };
```
OlivierBBB, line 65: This computation is replicated several times in the PR, maybe it should be extracted ?

OlivierBBB (commented):
One possible error in `isMemoryExpansionFault` + some other minor comments

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
letypequividelespoubelles: I can modify it, I just kept the same name as before.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/DataCopy.java
letypequividelespoubelles: yes and no: the names are sometimes different (size vs length for exemple), and it would need an intermediate abstracyt class just for this, not sure it worths it ...

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
letypequividelespoubelles: well ...  it wasn't the case, it was failing

letypequividelespoubelles: we're creating it with a `Range.fromOffsetAndSize(offset, length);` so we keep the ofsfet value even if size == 0

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
letypequividelespoubelles: I've added a LONDON_MXPX_THRESHOLD for disambiguite

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
OlivierBBB: Yes, it just jumped out at me when reading the PR.

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
OlivierBBB: Hm, I looked over that class yesterday while reviewing the above and I remember reading something to the effect
```java
return size == 0 ? Range.EMPTY : Range(offset, size);
```

OlivierBBB: https://github.com/Consensys/linea-tracer/blob/b13deae81023d249c252403f84f1735b3a3b6aa2/arithmetization/src/main/java/net/consensys/linea/zktracer/types/Range.java#L51-L53

OlivierBBB: Ah, the main constructor is the one with long arguments ... Maybe we should introduce the same logic for the `(long,long)` constructor. I have a vague feeling that I didn't do this for a particular reason back when I wrote those but I'm not sure ...
https://github.com/Consensys/linea-tracer/blob/b13deae81023d249c252403f84f1735b3a3b6aa2/arithmetization/src/main/java/net/consensys/linea/zktracer/types/Range.java#L38-L40

.github/workflows/reusable-unit-tests.yml
```diff
@@ -31,7 +31,6 @@ jobs:
         run: GOMEMLIMIT=26GiB ./gradlew :arithmetization:test
         env:
           JAVA_OPTS: -Dorg.gradle.daemon=false
-          CORSET_FLAGS: disable
```
OlivierBBB: What do these do @letypequividelespoubelles ?

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
letypequividelespoubelles: yes exactly
Maybe we could look at this later (open a reminder issue), but I didn't want to touch this, as I had the same (very) vague reminiscence ...

testing/src/main/java/net/consensys/linea/reporting/TracerTestBase.java
```diff
@@ -31,24 +28,26 @@ public class TracerTestBase {
...
-              case "SHANGHAI" -> ChainConfig.MAINNET_TESTCONFIG(SHANGHAI);
-              case "CANCUN" -> ChainConfig.MAINNET_TESTCONFIG(CANCUN);
-              case "PRAGUE" -> ChainConfig.MAINNET_TESTCONFIG(PRAGUE);
-              default -> throw new IllegalArgumentException(
-                  "Unknown fork: " + System.getProperty("unit.replay.tests.fork"));
-            };
+    // Configure chain information and fork before any tests are run, including any methods used as
+    // MethodSource.
+    TracerTestBase.chainConfig =
+        switch (getForkOrDefault("CANCUN")) {
```
OlivierBBB: So we are switching to `CANCUN` after all ?

arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/gas/projector/Call.java
OlivierBBB: Issue created, https://github.com/Consensys/linea-tracer/issues/2260

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
```diff
@@ -129,7 +130,7 @@ private static boolean isStackOverflow(final MessageFrame frame, OpCodeData opCo
 
   private static boolean isMemoryExpansionFault(Fork fork, GasProjection op) {
     return switch (fork) {
-      case LONDON, PARIS, SHANGHAI -> op.mxpxOffset(fork) >= MXPX_THRESHOLD;
+      case LONDON, PARIS, SHANGHAI -> op.mxpxOffset(fork) >= LONDON_MXPX_THRESHOLD;
       case CANCUN, PRAGUE, OSAKA -> op.mxpxOffset(fork) > MXPX_THRESHOLD;
```
OlivierBBB: I added a `CANCUN_MXPX_THRESHOLD` constant in the constraints PR.

OlivierBBB (approved):
LGTM, minor comments / questions

.github/workflows/reusable-unit-tests.yml
letypequividelespoubelles: hum ... it has been deleted by @amkCha in a previous PR, already merged in main, let me check if I have any rebase issue.

testing/src/main/java/net/consensys/linea/reporting/TracerTestBase.java
letypequividelespoubelles: no, I need to put it back to London ...

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/signals/Exceptions.java
letypequividelespoubelles: oh nice, I'll update constraints then

PR #2257, opened 2025-09-09
feat: switch nightly batch to weekly

### LINKED ISSUE #2247: fix: nightly tests
The purpose here is to move some number of nightly tests over to weekly tests.

DavePearce:
This switches one batch of nightly tests to weekly tests.  This also re-enables a set of nightlys which were previously disabled, though putting them as weeklys.

### DIFF
--- a/arithmetization/src/test/java/net/consensys/linea/replaytests/WeeklyTests.java
+++ b/arithmetization/src/test/java/net/consensys/linea/replaytests/WeeklyTests.java
@@ -26,9 +26,9 @@ import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @Disabled
-@Tag("nightly")
+@Tag("weekly")
 @ExtendWith(UnitTestWatcher.class)
-public class NightlyTests extends TracerTestBase {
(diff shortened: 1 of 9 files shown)


arithmetization/src/test/java/net/consensys/linea/replaytests/WeeklyTests.java
```diff
@@ -26,9 +26,9 @@
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @Disabled
```
letypequividelespoubelles, line 28: does that mean they're all disabled ?

letypequividelespoubelles (approved):
some questions

arithmetization/src/test/java/net/consensys/linea/replaytests/WeeklyTests.java
DavePearce: Yes, and they have been for some time.  We can try to get them going again, however, as weekly tests.  It just depends how long they take.

DavePearce: If I can get the weeklys to pass I will try to enable these ones.

PR #2275, opened 2025-09-12
update-tracer-release-name

letypequividelespoubelles:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-releaseVersion=beta-v4.0-rc0-counting-only---Prague-EXCLUDING-new-precompiles-and-7702
+releaseVersion=beta-v4.0-rc1-up-to-CANCUN-EXCLUDING-new-precompiles-and-7702
```
OlivierBBB: The `EXCLUDING-new-precompiles-and-7702` part doesn't refer to Cancun but to Prague so the name is surprising. I guess this release does both counting and tracing for Cancun and it contains the stuff necessary to count Prague excluding `...` ?

letypequividelespoubelles: yes, it's everything up to Cancun + Prague excluding bla

letypequividelespoubelles: but if you put all infos, it becomes unreadable :(

OlivierBBB: Maybe
```
beta-v4.0-rc1---Cancun-counting-and-tracing---Prague-counting-only-EXCLUDING-new-precompiles-and-7702
```
or so

lorenzogentile404 (approved):

gradle.properties
OlivierBBB: `beta-v4.0-rc1---Cancun-counting-and-tracing---Prague1-counting-only`

letypequividelespoubelles: no because we can count Prague only with limitless prover, not with ZkTracer, as we might get NPE

letypequividelespoubelles: I propose beta-v4.0-rc1-CANCUN only, too long is too long. POlus it's missing some lookups so we'll have to de a newer release soon

PR #2276, opened 2025-09-12
Improve OOB module performance

ahamlat:
We noticed during that performance complexity of this method  `net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.addAndGet` is not good as it has two loops to check if an element is inside two sets (`2 * O(n)`). The current implementation is justified by the fact that we need to get the existing element from those sets and not the verified (input) element, and set doesn't expose a `get(e)` method.

This is an attempt to reduce the complexity of that method by introducing LinkedHashMap data structures, the improvement is from `2 * O(n)` to `2 * O(1) `for average cases (no collisions) and `2 * O(log(n))` for worst cases (collisions).

The main issue with this implementation is the duplication of data. I need to check the impact in terms of allocation rate and GC activity. Depending on the result, we can decide to ship this PR or not.

This PR improves the performances of `addAndGet` method from 76 samples to 4 samples.

### Before this PR

```
for (E op : operationsInTransactionBundle()) {
      if (op.equals(e)) {
        return existingOperation(op);
      }
    }
    for (E op : operationsCommitedToTheConflation()) {
      if (op.equals(e)) {
        return existingOperation(op);
      }
    }
```

image

### Wit this PR

```
  E existing = operationsInTransactionBundle().get(e);
    if (existing != null) return existingOperation(existing);

    E existingInCommitted = operationsCommitedToTheConflation().get(e);
    if (existingInCommitted != null) return existingOperation(existingInCommitted);
```

image


letypequividelespoubelles:
LGTM, I'd rather keep the name StackedSet as we use the map as a set, as we always have k == v ...

lu-pinto:
> LGTM, I'd rather keep the name StackedSet as we use the map as a set, as we always have k == v ...

Same, even though impl is a Map, StackedMap<T> doesn't look right as there's always 2 types.
Also looks good to me, +1 on checking memory and GC.

ahamlat:
Ok, I did rollback the change in the names.

letypequividelespoubelles:
To be sure we really use it a set, I would even like an assert (at commit time ? ) that k==v. It'll be horrible to debug if end up somehow with k != v at a point

linea-constraints
letypequividelespoubelles, line 1: plz don't update the constraint with an older commit :)

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -67,44 +67,39 @@ public int lineCount() {
...
-      if (op.equals(e)) {
-        return existingOperation(op);
-      }
-    }
-    for (E op : operationsCommitedToTheConflation()) {
-      if (op.equals(e)) {
-        return existingOperation(op);
-      }
-    }
+    E existing = operationsInTransactionBundle().get(e);
```
letypequividelespoubelles: final ?

arithmetization/src/main/java/net/consensys/linea/zktracer/container/stacked/ModuleOperationStackedSet.java
```diff
@@ -67,44 +67,39 @@ public int lineCount() {
...
+    if (existing != null) return existingOperation(existing);
+
+    E existingInCommitted = operationsCommitedToTheConflation().get(e);
```
letypequividelespoubelles: final ?

letypequividelespoubelles (approved):
LGTM, I would like an assert (at commit time, ie when we `getAll`)  that we have k == v

linea-constraints
ahamlat: Yeah I didn't pay attention to that. I will remove it.

PR #2212, opened 2025-08-28
perf: rm oob duplicate in the trace

letypequividelespoubelles:

PR #2296, opened 2025-09-16
fix: missing equals and hashcode for `ExpCall` and `GasParameters`

### LINKED ISSUE #2295: fix: missing `equals()` and `hashCode()` for `ExpCall` implementations
The generated trace can contain duplicate frames for the `exp` module due to the lack of `equals()` and `hashCode()` methods on the implementations of `ExpCall`.  We can see this by looking at a trace for this module:

```
exp:
      |     #0 |     #1 |     #2 |     #3 |     #4 |     #5 |     #6 |     #7 |     #8 |     #9 |
 INST | 0xee0a | 0xee0a | 0xee0a | 0xee0a | 0xee0a | 0xee0a | 0xee0a | 0xee0a | 0xee0a | 0xee0a |
  ARG |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |
  CDS |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |
  EBS |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |
  RES |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |    0x0 |
```

DavePearce:
Both of these classes / interfaces are used within instances of `ModuleOperation` and require proper `equals()` and `hashCode()` methods.  Some notes on this:

* For the two implementations of `ExpCall`, the outputs are included in the `equals()` and `hashCode()` methods.  However, this is safe because these methods are filled when `ExpOperation` is created.
* For `GasParameters` there are no output values per se.

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/exp/ExplogExpCall.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/exp/ExplogExpCall.java
@@ -17,6 +17,7 @@ package net.consensys.linea.zktracer.module.hub.fragment.imc.exp;
 
 import static net.consensys.linea.zktracer.Trace.EXP_INST_EXPLOG;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.experimental.Accessors;
@@ -27,9 +28,10 @@ import org.apache.tuweni.bytes.Bytes;
 @Accessors(fluent = true)
 @Getter
 @Setter
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
 public class ExplogExpCall implements ExpCall {
-  EWord exponent;
-  long dynCost;
+  @EqualsAndHashCode.Include EWord exponent;
+  @EqualsAndHashCode.Include long dynCost;
 
   @Override
   public int expInstruction() {
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/exp/ExplogExpCall.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/exp/ExplogExpCall.java
@@ -56,4 +56,8 @@ public class ExplogExpCall implements ExpCall {
         .pMiscExpData2(exponent.lo())
         .pMiscExpData5(Bytes.ofUnsignedLong(dynCost));
   }
+
+  public String toString() {
+    return "EXPLOG(" + exponent.toString() + ")=" + dynCost;
+  }
 }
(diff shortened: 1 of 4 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/oob/OobCall.java
```diff
@@ -43,4 +43,8 @@ protected OobCall() {
   public abstract void callExoModulesAndSetOutputs(final Add add, final Mod mod, final Wcp wcp);
 
   public abstract int ctMax();
+
+  public abstract boolean equals(Object o);
```
letypequividelespoubelles, line 47: why do you add this ?

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/exp/ExplogExpCall.java
letypequividelespoubelles, line 1: for the moment, as it is, I think it won't work, let's see after your asm PR is merged. 

see why: https://github.com/Consensys/linea-tracer/pull/2212#discussion_r2307086164

letypequividelespoubelles (changes requested):
let's wait the asm exp to be merged

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/oob/OobCall.java
DavePearce: To force every implementation to have an `equals()` and `hashCode()` method.  Its just a safety precaution ... not actually needed.

arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/fragment/imc/exp/ExplogExpCall.java
DavePearce: Why not?  Are you talking about `GasParameters` or `ExpCall`?  The `ExplogExpCall` has the advantage that its never put into a `StackedSet` directly, but only when constructed via `new ExpOperation()` which fills the output values.

DavePearce: Also, your comment doesn't seem to apply in this case because the output values are always computed.  So it is different from `OobCall` in this regard.

letypequividelespoubelles: yes, that's exactly my point, we shouldn't compute the output when not needed. Anyway, we can already merge this as it is, but keep an issue open to fix this.

letypequividelespoubelles (approved):
ok

PR #2297, opened 2025-09-16
test: no modulekey duplicate

### LINKED ISSUE #2292: check that no moduleKey duplicate in modules / modulesToCount

letypequividelespoubelles:


DavePearce (approved):
LGTM

arithmetization/src/test/java/net/consensys/linea/LineCountingTracerTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class LineCountingTracerTest extends TracerTestBase {
+
+  @Test
+  void noDuplicateNamesInModules() {
+    final ZkTracer tracer = new ZkTracer(chainConfig);
+    final List<String> tracerToCount =
+        tracer.getModulesToCount().stream().map(Module::moduleKey).toList();
+    final List<String> tracedModules =
+        tracer.getHub().getModulesToTrace().stream().map(Module::moduleKey).toList();
```
OlivierBBB, line 36: You could define these as sets rather than lists directly (thus no stream distinct etc ...)

OlivierBBB (approved):
LGTM. Minor comment about using sets from the start

arithmetization/src/test/java/net/consensys/linea/LineCountingTracerTest.java
letypequividelespoubelles: I want to stick as much as possible as what we do in ZkTracer / ZkCounter where we use List

PR #2301, opened 2025-09-16
Reduce memory allocations on non opcode data

### LINKED ISSUE #2300: Reduce memory allocation rate on non opcodes
Memory allocation profiling showed that line counting during block builing on the sequencer generated around 70 GiB in 60 seconds, which corresponds to an allocation rate of more than 1 GiB/s.

Image

This happens when OpCode.of(int) is not resolved to an existing OpCode in the EVM and execute the method forNonOpCodes below

```
  public static OpCodeData forNonOpCodes(int value) {
    return new OpCodeData(
        OpCode.INVALID,
        value,
        INVALID,
        new StackSettings(
            Pattern.ZERO_ZERO,
            0,
            0,
            GasConstants.G_ZERO,
            false,
            false,
            false,
            false,
            false,
            false,
            false,
            false),
        new Billing(GasConstants.G_ZERO, BillingRate.NONE, MxpType.NONE));
  }
```
Most of the allocation rate seems to be related to 
```
 new Billing(GasConstants.G_ZERO, BillingRate.NONE, MxpType.NONE));
```
Is there a reason why StackSettings and Billing are not static ? something like this
```
 return new OpCodeData(
        OpCode.INVALID,
        value,
        INVALID,
        StackSettings.DEFAULT,
        Billing.DEFAULT);
  }
```
Interestingly, `Billing.DEFAULT `already exists.

ahamlat:
Make OpCodeData data instances use static instances of Billing and StackSettings when created using `forNonOpCodes(int value)` method.
More details in the issue below.

Fixes https://github.com/Consensys/linea-tracer/issues/2300

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/opcode/OpCodeData.java
@@ -24,10 +24,7 @@ import java.util.Objects;
 
 import net.consensys.linea.zktracer.Fork;
 import net.consensys.linea.zktracer.opcode.gas.Billing;
-import net.consensys.linea.zktracer.opcode.gas.BillingRate;
-import net.consensys.linea.zktracer.opcode.gas.GasConstants;
 import net.consensys.linea.zktracer.opcode.gas.MxpType;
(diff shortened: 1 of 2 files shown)


lorenzogentile404 (approved):
LGTM

PR #2314, opened 2025-09-18
feat: run Prague unit, change genesis for protocol spec per fork, fix…

amkCha:
… canonical account snapshot


arithmetization/src/main/java/net/consensys/linea/zktracer/module/hub/section/systemTransaction/EIP2935HistoricalHash.java
```diff
@@ -51,7 +51,7 @@ public EIP2935HistoricalHash(final Hub hub, WorldView world, ProcessableBlockHea
     final short previousBlockNumberModulo =
         currentBlockIsGenesis ? 0 : (short) ((blockHeader.getNumber() - 1) % HISTORY_SERVE_WINDOW);
     final AccountSnapshot blockhashHistoryAccount =
-        AccountSnapshot.canonical(hub, EIP2935_HISTORY_STORAGE_ADDRESS);
+        AccountSnapshot.canonical(hub, world, EIP2935_HISTORY_STORAGE_ADDRESS, false);
```
amkCha, line 54: Would the fix look like this ? Should this address be cold as well ? @letypequividelespoubelles

letypequividelespoubelles: yes, and we don't care about it's warmth, so let's put cold as in 4788.

letypequividelespoubelles (approved):
LGTM, do you know why the Prague genesis file is specific ?

amkCha:
> LGTM, do you know why the Prague genesis file is specific ?

Yes, we now need these addresses 
image

I checked in Besu codebase, for Cancun, they were retrieved directly in the code.

Actually even in the rlp from the transaction, is that possible ? Maybe before we needed to add the address of the beacon contract when doing a withdrawal 
image

PR #2322, opened 2025-09-22
2146 fork switching test

amkCha:
- Adapt `BesuExecutionTools` to handle fork switching and check the traces against the correct constraints
- Test `LondonToParis` fork switch : this is a separate test, as Clique sets live timestamping, it's difficult to predict future timestamps for fork switching in Units with a hardcoded Genesis file
- Test `ParisToCancun` fork switch
- Add debug log trace to identify which fork we are tracing on, coming from Besu API
- Add fork tracing test on Prague


testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
```diff
@@ -362,4 +353,169 @@ private static int findFreePort() {
...
+    var shanghaiTime = genesisConfigBuilder.getShanghaiTime();
+    var cancunTime = genesisConfigBuilder.getCancunTime();
+    var pragueTime = genesisConfigBuilder.getPragueTime();
+
+    // No fork switch specified in the genesis file, stay in London
+    if (TTD == null && shanghaiTime == null && cancunTime == null && pragueTime == null) {
+      return Fork.LONDON;
+    }
+
+    var terminalTotalDifficulty = new BigInteger(TTD);
```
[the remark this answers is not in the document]
amkCha: If TTD is null, then there is no transition to Paris, so there can't be one for Shanghai

arithmetization/src/test/java/net/consensys/linea/zktracer/ExampleBesuTest.java
```diff
@@ -47,7 +45,7 @@ void test(TestInfo testInfo) {
 
     switch (fork) {
       case LONDON -> {}
-      case PARIS -> compiler.op(OpCode.DIFFICULTY);
+      case PARIS -> compiler.op(OpCode.PREVRANDAO);
       case SHANGHAI -> compiler.op(OpCode.PUSH0);
       case CANCUN -> compiler.op(OpCode.MCOPY);
```
letypequividelespoubelles: is it intended to get a stack underflow exception for some fork ?

Before the fork specific opcode you have only one stack element wich is not enough for a MCOPY

letypequividelespoubelles (approved):
LGTM with one comment

arithmetization/src/test/java/net/consensys/linea/zktracer/ExampleBesuTest.java
amkCha: You're right ! Thanks for the catch, it's updated :)

PR #2334, opened 2025-09-23
fix: don't pop trace start block operations

### LINKED ISSUE #2333: poping first tx will pop the line from start block also in ZkTracer

letypequividelespoubelles:


amkCha (approved):

linea-constraints
amkCha, line 1: Updating constraints right ?

letypequividelespoubelles: to latest master yes

arithmetization/src/test/java/net/consensys/linea/LineCountingTracerTest.java
```diff
@@ -117,4 +122,27 @@ void sameModuleAcrossAllForkWithZkCounterAndTracer() {
...
+            .baseFee(DEFAULT_BASE_FEE)
+            .buildBlockHeader();
+
+    final ZkTracer tracer = new ZkTracer(chainConfig);
+    tracer.traceStartConflation(1);
+    tracer.traceStartBlock(world, blockHeader, DEFAULT_COINBASE_ADDRESS);
+    final int sizeBefore = tracer.getHub().lineCount();
+    tracer.popTransactionBundle();
+    final int sizeAfter = tracer.getHub().lineCount();
+    checkArgument(sizeBefore == sizeAfter, "Some start block stuff have been popped");
```
OlivierBBB: So you _want_ the `tracer.popTransactionBundle()` to not modify the line count ?

OlivierBBB: I'm not understanding the intent I guess.

To recap, you've added `commitTransactionBundle()` to `traceStartBlock(...)`, so that calling `tracer.popTransactionBundle()` as you do in in this test should have no effect, and this is what you expect to find here (`sizeBefore == sizeAfter`). I guess my question is:

> What was getting popped previously that shouldn't have been popped ?

OlivierBBB (approved):
LGTM, though I'm not really sure what it was that was getting popped when it shouldn't have been.

arithmetization/src/test/java/net/consensys/linea/LineCountingTracerTest.java
letypequividelespoubelles: When popBundle is called it's removing everything up to the previoius (if any) commitTransactionBundle().

Before:
 all op that happens in traceStartBlock are not commited (ie system transactions, blocData + related wcp calls). So if the first tx/bundle of the block overflows the limits, this tx is popped, but all the block related stuff too.
 
 Now: block related stuff are commited at the end of traceStartBlock, so if teh first tx / bundle is popped, we keep the block related stuff in the line count.

PR #2335, opened 2025-09-24
feat: upgrade go-corset to 1.1.22

amkCha:
This will enable a more explicit error message in case we are trying to trace an unexisting column

Closes #2332


letypequividelespoubelles (approved):
Can you update to latest master of constraint by the occasion ?

PR #2376, opened 2025-10-01
fix: access list can be empty for Cancun

amkCha:
> [!NOTE]
> Use an empty list when a transaction's access list is absent, avoiding reliance on a present Optional.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ea6269ee265dfec3f19a6d7eacb909922d87bf8f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxn/cancun/CancunRlpTxnOperation.java
+++ b/arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxn/cancun/CancunRlpTxnOperation.java
@@ -76,8 +76,7 @@ public class CancunRlpTxnOperation extends RlpTxnOperation {
     phaseSectionList.add(new DataPhaseSection(rlpUtils, tx));
 
     // Phase Access List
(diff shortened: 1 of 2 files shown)


arithmetization/src/main/java/net/consensys/linea/zktracer/module/rlptxn/cancun/CancunRlpTxnOperation.java
```diff
@@ -76,7 +76,8 @@ public CancunRlpTxnOperation(RlpUtils rlpUtils, Trm trm, TransactionProcessingMe
     phaseSectionList.add(new DataPhaseSection(rlpUtils, tx));
 
     // Phase Access List
-    if (tx.getBesuTransaction().getType() != FRONTIER) {
```
letypequividelespoubelles, line 79: no, it'll break the constraint. We have to say that if it's not present, then we do as if it's an empty one.

letypequividelespoubelles (approved):
LGTM

PR #2417, opened 2025-10-10
fix: rebase / merge issue with prover columns

OlivierBBB:
> [!NOTE]
> Always populate `proverRelativeUserTxnNumberMax` from the block snapshot for all transactions and bump `linea-constraints` submodule.
> 
> - **Txndata (`CancunTxnDataOperation`)**:
>   - Always set `proverRelativeUserTxnNumberMax` to `blockSnapshot.getNbOfTxsInBlock()` for all transactions.
> - **Dependencies**:
>   - Update `linea-constraints` submodule to `2c9e095`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 816f3e1a97c077add0a5a43364b68ebf5cbd2994. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


letypequividelespoubelles (changes requested):
why not tell the prover to get the nb of tx per block in the BlockData module ? This info is already present, would be easier imho

OlivierBBB:
> why not tell the prover to get the nb of tx per block in the BlockData module ? This info is already present, would be easier imho

Because they asked for it this way. I guess it means they can get the data from one place rather than from 2 and re-use their original interface. If we can avoid changes on our side disrupting their side we should do it.

letypequividelespoubelles (approved):
LGTM

PR #2458, opened 2025-10-29
Feat/replays for prague with besu and issue1335

amkCha:
When running a replay test for an lt trace + shomei trace

use `besuNodeFastReplayTests` here
image

Comment line 99-106 in `net/consensys/linea/testing/ExecutionEnvironment.java` to keep the traces

and you can find traces in 
image

---

> [!NOTE]
> Increase timeouts and block-build delay in Besu replay tools and update `ConflationSnapshot.from` usage to include fork name.
> 
> - **Testing**:
>   - **BesuExecutionTools**:
>     - Set `OkHttpClient` read timeout to 30s.
>     - Increase block build delay: `blockBuildingTimeMs` from `period*1000` to `period*10000`.
>     - Extend trace file wait timeout from `10` to `60` seconds.
>   - **MultiBlockExecutionEnvironment**:
>     - Update `ConflationSnapshot.from(...)` call to include `testsChain.fork.name()` (new API signature).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6f562a257684df6868839705c0fe8b0d76cc8d23. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/BlockHeaderSnapshot.java
```diff
@@ -87,14 +87,10 @@ public BlockHeader toBlockHeader() {
             .mixHash(Hash.fromHexString(this.mixHashOrPrevRandao))
             .prevRandao(Bytes32.fromHexString(this.mixHashOrPrevRandao))
             .nonce(this.nonce)
-            .blockHeaderFunctions(new CliqueBlockHeaderFunctions());
+            .blockHeaderFunctions(new MainnetBlockHeaderFunctions())
```
amkCha: I'm not sure if it's the right class to use from now on - to be checked with @fab-10 :pray:

testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java
```diff
@@ -97,7 +98,7 @@ public BesuExecutionTools(
     this.testName = tmpTestName.substring(0, Math.min(tmpTestName.length(), 200));
     int besuPort = findFreePort();
     int shomeiPort = findFreePort();
```
amkCha, line 100: In this class, we change all the timeouts to allow bigger blocks and tracers to be traced than our regular units are doing

arithmetization/src/main/java/net/consensys/linea/blockcapture/snapshots/BlockHeaderSnapshot.java
amkCha: but it is working though :)

DavePearce: The problem you could have here is that old replay tests may fail, as they require Clique.  However, that perhaps depends on whether `MainnetblockHeaderFunctions` accounts for this or not.

DavePearce:
FYI @amkCha you can set the environment variable `PRESERVE_TRACE_FILES` to e.g. `1` to ensure the trace files are not deleted, rather than commenting out lines of code.

DavePearce (approved):
LGTM!

PR #2491, opened 2025-11-18
feat: update WCP with assembly implementation

### LINKED ISSUE #2490: feat: update WCP with assembly implementation
The goal of this feature is to replace the existing `wcp` module with an assembly implementation.  This is not entirely straightforward, as it requires removing the dependency of `rlputils` on the `wcp.CT_MAX` column (as this column will no longer exist).  To do this, we will make use of a new `go-corset` feature `defcall`.  Specifically, this will be used to call into an assembly function to perform the same calculation _but without the need to modify the tracer_.

DavePearce:
The purpose of this PR is to replace the existing `wcp` module with an assembly implementation.  This requires first breaking the dependency of `rlputils` on the column `wcp.CT_MAX` (as this column will no longer exist).  To do this, we make use of a secondary function `maxlog()` which computes the necessary information instead (i.e. it calculates the value previously determined by `wcp.CT_MAX`).

---

> [!NOTE]
> Simplifies WCP by deleting `WcpCall` and reducing `WcpOperation` to a single-row trace while updating CI to Go 1.24.6 and go-corset v1.1.28.
> 
> - **Arithmetization / WCP**:
>   - Simplify `WcpOperation` tracing to a single row via `trace.inst(...).arg1(...).arg2(...).res(...)` and `computeLineCount()` now returns `1`.
>   - Remove per-byte/bit logic, counters, and `ctMax`-driven paths; delete `WcpCall` class.
>   - Keep opcode result computation; comparator unchanged aside from a clarifying comment.
> - **CI**:
>   - Update `.github/actions/setup-go-corset/action.yml` to Go `1.24.6` and `go-corset` `v1.1.28`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ca39d093cab7ea1829246f6b6e87f8b19558e66f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


letypequividelespoubelles:
`update WCP with assembly implementation` is not done in this PR, it's only chnging the call from RLP_UTILS here :)

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/WcpOperation.java
```diff
@@ -157,97 +83,15 @@ private Bytes16 calculateAdj(boolean cmp, BigInteger arg1, BigInteger arg2) {
   }
 
   void trace(Trace.Wcp trace, int stamp) {
-    this.compute();
-
-    final boolean resLo = this.resLo;
-    final boolean oli = isOli();
-    final boolean vli = isVli();
+    // Calculate result
+    final boolean resLo = calculateResult(wcpInst, arg1, arg2);
```
letypequividelespoubelles: better name it resBool, or just res, there are no high/lo part of the (bin) result.

letypequividelespoubelles (approved):
LGTM

arithmetization/src/main/java/net/consensys/linea/zktracer/module/wcp/WcpOperation.java
DavePearce: Done

PR #2513, opened 2025-11-24
feat: read blobBaseFee from besu blockchainService

### LINKED ISSUE #2511: get blobBaseBee from frame and not hardcode its value

letypequividelespoubelles:
> [!NOTE]
> Add blob base fee as a first-class public input fetched from Besu and plumbed through capture, tracing, hubs, and tests, including Cancun BLOBBASEFEE handling and snapshot support.
> 
> - **Public Inputs & Data Flow**:
>   - Introduce `PublicInputs` aggregating `historicalBlockhashes` and `blobBaseFees`, with helpers (`retrieveHistoricalBlockHashes`, `getBlobBaseFees`, defaults).
>   - Refactor `ZkTracer`, `Hub`, and all fork hubs to accept/use `PublicInputs` (replacing raw hash maps).
> - **Blob Base Fee Support**:
>   - Capture and trace `BLOBBASEFEE`:
>     - `BlockCapturer`/`Reaper`: record per-block blob base fees; carry into `ConflationSnapshot` with sensible defaults.
>     - Blockdata modules/operations (`BlockData`, London/Paris/Shanghai/Cancun variants): plumb blob base fees; implement Cancun `BLOBBASEFEE` tracing.
> - **RPC & Tooling**:
>   - `GenerateConflatedTracesV2`: fetch blob base fees from Besu, construct `PublicInputs`, and pass to `ZkTracer`.
>   - `CaptureToFile`: update imports to use `PublicInputs` retrieval.
> - **Blockhash Module**:
>   - Minor fixes/renames (`checkBlockHashConsistencies`, arg handling); remove internal historical-hash retrieval.
> - **Testing**:
>   - Update replay and toy/multi-block environments to provide blob base fees (use defaults when absent).
> - **Misc**:
>   - Log tweaks; bump `releaseVersion` to `beta-v4.4-rc6`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1e8dd92c1ecf295b5ccadae5bea5808c8289577d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


arithmetization/src/main/java/net/consensys/linea/zktracer/ZkCounter.java
```diff
@@ -387,7 +387,7 @@ public ZkCounter(LineaL1L2BridgeSharedConfiguration bridgeConfiguration, Fork fo
             LogTopic.of(bridgeConfiguration.topic()));
     moduleToCount = Stream.concat(checkedModules().stream(), uncheckedModules().stream()).toList();
 
-    log.info("[ZkCounter] Created ZkCounter");
+    log.info("[ZkCounter] Created ZkCounter for fork {}", fork);
```
OlivierBBB, line 390: 👍

arithmetization/src/main/java/net/consensys/linea/zktracer/types/PublicInputs.java
```diff
@@ -0,0 +1,125 @@
...
+      BlockchainService blockchain,
+      long firstBlockNumberOfConflation,
+      long lastBlockNumberOfConflation) {
+
+    final Map<Long, Hash> historicalBlockHashes =
+        new HashMap<>(
+            (int)
+                (BLOCKHASH_MAX_HISTORY
+                    + lastBlockNumberOfConflation
+                    - firstBlockNumberOfConflation));
```
OlivierBBB, line 76: Yes. Initially thought this might require
```java
BLOCKHASH_MAX_HISTORY + (lastBlockNumberOfConflation - firstBlockNumberOfConflation + 1)
```
but it doesn't.

arithmetization/src/main/java/net/consensys/linea/zktracer/types/PublicInputs.java
```diff
@@ -0,0 +1,125 @@
...
+
+    final long firstBlockToRetrieve =
+        Math.max(firstBlockNumberOfConflation - BLOCKHASH_MAX_HISTORY, 0);
+    final long lastBlockToRetrieve =
+        lastBlockNumberOfConflation == 0 ? 0 : lastBlockNumberOfConflation - 1;
```
OlivierBBB, line 81: Sanity check: block number 0 corresponds to the chain's genesis block. I'm confused as to why this shouldn't be illegal / throw an exception. Do we have any tests where we "prove" the genesis block i.e. where the conflation starts at `startBlockNumber == 0`?

Hm, now that I think of it we do have certain binary flags à la `IS_GENESIS` for instance for system transactions IIRC ...

OlivierBBB (approved):
LGTM. Cursor makes a few comments, in particular one about not trimming the leading zeros off of the block number and a risk for a NPE.

arithmetization/src/main/java/net/consensys/linea/zktracer/types/PublicInputs.java
letypequividelespoubelles: what should fail ? 

And yes, the system transaction tests have tests where we start at the genesis block

arithmetization/src/main/java/net/consensys/linea/zktracer/module/blockhash/Blockhash.java
```diff
@@ -139,7 +136,7 @@ private void checkBlockHashConsistancies(BlockhashOperation op) {
     if (!op.blockhashRes().equals(Bytes32.ZERO)) {
       checkArgument(
           op.blockhashArg().trimLeadingZeros().size() <= 8, "Block number must fit in a long");
-      final long blockNumber = op.blockhashArg().trimLeadingZeros().toLong();
+      final long blockNumber = op.blockhashArg().toLong();
```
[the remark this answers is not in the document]
letypequividelespoubelles: the toLong() has been updated in Tuweni to not require the trimming

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/tracegeneration/GenerateConflatedTracesV2.java
```diff
@@ -127,16 +131,17 @@ private TraceFile generateTraceFile(PluginRpcRequest request) {
           BesuServiceProvider.getBesuService(besuContext, BlockchainService.class);
       // Retrieve fork from Besu plugin API with block number
       final Fork fork = getForkFromBesuBlockchainService(blockchainService, fromBlock, toBlock);
+      final BigInteger chainId =
+          blockchainService
+              .getChainId()
+              .orElseThrow(() -> new IllegalStateException("ChainId must be provided"));
       final Map<Long, Hash> historicalBlockHashes =
           retrieveHistoricalBlockHashes(blockchainService, fromBlock, toBlock);
+      final Map<Long, Bytes> blobBaseFees = getBlobBaseFees(blockchainService, fromBlock, toBlock);
```
amkCha, line 140: so in the end, we do 2 calls to the blockchain service, one to get the blocks, then the headers then the hash, and after the blobGasPrice. Maybe we can keep track of the headers we retrieved of merged the 2 methods ?

arithmetization/src/main/java/net/consensys/linea/zktracer/types/PublicInputs.java
```diff
@@ -0,0 +1,128 @@
...
+      long firstBlockNumber, long lastBlockNumber) {
+    return new PublicInputs(
+        new HashMap<>(), getDefaultBlobBaseFees(firstBlockNumber, lastBlockNumber));
+  }
+
+  public boolean allPublicInputsKnown() {
+    if (historicalBlockhashes == null || blobBaseFees == null) {
+      return false;
+    }
+    return historicalBlockhashes.size() >= BLOCKHASH_MAX_HISTORY && !blobBaseFees.isEmpty();
```
amkCha, line 58: we might not know 1st and last block number in conflation but if I have a conflation of 5, and I have 256 historical block hashes, do we consider we have all public inputs known ?

amkCha:
Nice PR :+1: and thanks for changing the logging on ZkCounter, will be helpful !

arithmetization/src/main/java/net/consensys/linea/zktracer/types/PublicInputs.java
letypequividelespoubelles: we should have 256 + nb_of_blocks_in_the_conflation, but when we create the ZkTracer we don't know the number of blocks of the conflation, so just put 256.
And in some cases, we could have all needed info while having less than 256 blocks (eg if we trace the conflation 4-8, we should have only 7 blockhashes).

arithmetization/src/main/java/net/consensys/linea/plugins/rpc/tracegeneration/GenerateConflatedTracesV2.java
letypequividelespoubelles: so, the blockhash makes `256 + (nb_of_blocks_in_the_conflation -1)`  call to get the header (and then the blockhash), while blobBaseFee does `2 * nb_of_blocks_in_the_conflation`.
So yes, we could save `nb_of_blocks_in_the_conflation -1` calls to get the header ...
We could optimize it indeen, but not a priority imho.

letypequividelespoubelles: plus when we'll want RTP, we'll do conflation of one block, so the perf will disappear ...

